PR #1, opened 2016-05-17
fixes

mrRay:
howdy-

this repos required a couple minor changes to get it up and running on my end- please note that i may have done something funky with the cmake stuff, as i am largely unfamiliar with it.

thanks very much for curating this!

specifics:
- updated build instructions for os x
- added FindSDL2 cmake module
- fixed a bug causing a crash with PerlinNoise/ShaderEngine.cpp
- calling projectM::projectM_resetGL() applies the new dimensions to the window width/height in the settings
- modified the parser to return an error rather than abort and terminate the process to prevent a crash with the provided sample presets

cheers
: : ray

### DIFF
--- a/src/libprojectM/Renderer/ShaderEngine.cpp
+++ b/src/libprojectM/Renderer/ShaderEngine.cpp
@@ -64,11 +64,11 @@ void ShaderEngine::setParams(const int texsize, const unsigned int texId, const 
 
 	//std::cout << "Generating Noise Textures" << std::endl;
 
-	PerlinNoise noise;
+	PerlinNoise		*noise = new PerlinNoise;
 
 	glGenTextures(1, &noise_texture_lq_lite);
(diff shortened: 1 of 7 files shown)


revmischa:
Looks great! 👍 
I could have sworn it was finding SDL2 just fine without that module, I forget how that was working through.

src/libprojectM/Renderer/ShaderEngine.cpp
```diff
@@ -64,11 +64,11 @@ void ShaderEngine::setParams(const int texsize, const unsigned int texId, const
 
 	//std::cout << "Generating Noise Textures" << std::endl;
 
-	PerlinNoise noise;
+	PerlinNoise		*noise = new PerlinNoise;
```
revmischa: does this need a corresponding `delete` somewhere?

mrRay:
oh FFS- yes, nice catch

PR #39, opened 2018-03-11
Rewrite ifdefs to work on bsd

splbio:


revmischa:
What about putting this little doodad into a header file and including it everywhere?

splbio:
It's generally bad form to have an "always included header" that pulls things in.  aka "all.h"

However I believe your earlier fix to switch __linux__ to __unix__ makes more sense than this patch.  However you can use this pattern in the future when it makes sense to do so.

revmischa:
i don't mean an "all.h", though i thought it was accepted practice to use a PCH file for projects.
i just mean a "projectM-openGL.h" that could be included in lieu of copying and pasting this block in every file that includes GL headers

splbio:
yes that is good.

PR #43, opened 2018-03-12
Fix the config.inp generation with 2 steps in AC and Automake

abramhindle:
Try to fix the config.inp file problem with both autoconf (configure and sed) and then a second build time stage with automake. config.inp.in becomes config.inp.in.in which configure builds as config.inp.in which automake uses to generate the actual config file.


revmischa:
thx! could use a test! :p

Issue #29, opened 2018-02-13
SDL conf file


revmischa:
Should be a config file for the SDL version

revmischa:
Works now ish

Issue #11, opened 2017-07-08
Support OpenGL ES


einsteinx2:
On the projectM website, it says that it has been ported to OpenGL ES 2.0 for Android and iOS use. Has that code ever been released anywhere? All I see in the repo seems to be ES 1.0 support and so far I've only been able to get that to half work in a test iOS app (it renders but not correctly).

Is the ES 2.0 version of the library closed source?

revmischa:
The android and iOS ports are unfortunately closed-source and I don't have the code. I tried to port to ES myself but I don't really have the chops or time to complete it. If you want to continue the effort I'd be happy to point you in the right direction. Sorry.

revmischa:
There is in fact a branch where some of the work has been done: https://github.com/projectM-visualizer/projectm/tree/gles

Notes on the old openGL immediate-mode calls that need to be ported are here: https://github.com/projectM-visualizer/projectm/blob/gles/glES-porting.txt

If anyone wants to work on this, please be my guest!

einsteinx2:
Damn I was hoping that wouldn't be the case. 

I actually have a "working" version using ES 1.0 and an iOS test app here: https://github.com/einsteinx2/libprojectm-ios. I even backfilled some missing rendering functions and got it compiling properly as an iOS static library. However, I'm totally stuck on some rendering issues. 

I posted to the /r/opengl subreddit, but didn't get any answers, however I have a detailed description there: https://www.reddit.com/r/opengl/comments/6kyvk6/looking_for_help_porting_projectm_to_ios_about_90/

Also, here is a video of the rendering differences between the Mac SDL version and my iOS version: https://www.youtube.com/watch?v=6otFBzkbd3Y

If you have a little OpenGL experience, maybe you can figure out what's wrong. I feel like I'm 90% there and am just missing something simple with the rendering.

It doesn't seem to be blending properly and also isn't rendering at the correct resolution. If I could at least get a working iOS test app using the ES 1.0 code, I could try and expand it from there.

einsteinx2:
Also possibly of interest, the latest Milkdrop 2 source code has been open sourced: https://sourceforge.net/projects/milkdrop2/

However it's written in DX9, so it would require porting to OpenGL first to be used anywhere but Windows. Though I did find Valve's ToGL DX9 > OpenGL wrapper here: https://github.com/ValveSoftware/ToGL so maybe there's hope without rewriting the whole thing.

[...]

PR #44, opened 2018-03-13
Native presets

revmischa:
Build and install native presets (aka .so/.dylib shared object files for presets instead of milkdrop equations)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,6 +31,7 @@ AC_CONFIG_FILES([
   src/libprojectM/Renderer/Makefile
   src/libprojectM/NativePresetFactory/Makefile
   src/libprojectM/MilkdropPresetFactory/Makefile
+  src/NativePresets/Makefile
 ])
 dnl FIXME
 dnl AC_CONFIG_FILES([src/libprojectM/config.inp],
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,1 +1,1 @@
-SUBDIRS=libprojectM projectM-sdl
+SUBDIRS=libprojectM projectM-sdl NativePresets
--- a/src/NativePresets/Makefile.am
+++ b/src/NativePresets/Makefile.am
@@ -9,8 +9,11 @@ AM_CFLAGS = ${my_CFLAGS} \
 -ffunction-sections \
 -fdata-sections
 
+presetsdir = $(pkgdatadir)/presets
+
 # native presets are shared object files.
-lib_LTLIBRARIES = \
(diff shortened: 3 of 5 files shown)


schnitzeltony:
I am preparing build of 2.2.1 with Openembedded for embedded linux system. The build system is not happy, installing libraries in /usr/share/...
Would it be possible to put them at /usr/lib/projectm/presets or similar?

revmischa:
Sure, patches welcome! 
Just curious - does it work on embedded? It doesn't support GLES (yet)

schnitzeltony:
Haven't seen it running yet but did not spend too much time on it. Am building for Raspi VC4 (open source/mesa) and that supports OpenGL 2.1 - don't know which limitations there are.
One question: regarding my original question: We have now native presets (=libs) and the milk files (which are no more installed by default with 2.2.1). Are both still supported and if yes: How to get different locations handled?

revmischa:
The milk files should be installed - to /usr/local/share/projectM/presets. Are they not?
projectM probably won't run on any embedded systems until GLES support is added - https://lwn.docx.pro/q6IHamrUd

revmischa:
If I do:
`./configure --prefix=$PWD/t --enable-shared=no --enable-sdl --enable-cg`
and then `make install`
```
$ ls -l t/share/projectM/presets/ | wc -l
    1996
```

revmischa:
Alright, I do warn you again that GLES is needed for it to work on an embedded system - https://github.com/projectM-visualizer/projectm/issues/11

The relevant preset install target is in the root Makefile.am:
```
# aka /usr/local/share/projectM
pm_data_dir = $(pkgdatadir)
pm_font_dir = $(pm_data_dir)/fonts
pm_presets_dir = $(pm_data_dir)/presets
pm_shaders_dir = $(pm_data_dir)/shaders

pm_data__DATA = src/libprojectM/config.inp
pm_font__DATA = fonts/Vera.ttf fonts/VeraMono.ttf
pm_shaders__DATA = src/libprojectM/Renderer/blur.cg \
                   src/libprojectM/Renderer/projectM.cg

install-data-local:
	test -z $(pkgdatadir) || $(MKDIR_P) $(pm_presets_dir)
	find "$(PRESETSDIR)" -type f -exec $(INSTALL_DATA) {} $(pm_presets_dir) \;
```

However, projectMSDL currently has settings hard-coded. There's an issue to make it use the normal config file (#29). It looks for the fonts etc from the current directory of the SDL executable right now. It needs to be changed to use the paths in a config file.

PR #42, opened 2018-03-12
Qt5

revmischa:
This is almost ready to merge!

PR #46, opened 2018-04-02
Qt fixes

deltaoscarmike:
These commits fixes:
- Qt automake files
- Pulseaudio automake files
- Both needed to be fixed because projectM-pulseaudio application combine them
- Compile flag conditionnals (SDL/Qt)

Tested and working on debian Sid with Qt5.9


revmischa:
Awesome, thank you!

revmischa:
Hm, even though `--enable-qt` is not enabled in the test (though it should be... or we should test different combinations...), the test builds ok but it seems to fail on install: https://travis-ci.org/projectM-visualizer/projectm/jobs/361505562

revmischa:
Please review https://github.com/projectM-visualizer/projectm/pull/42 !

PR #48, opened 2018-04-05
Add FTGL configure flag

deltaoscarmike:
Not enabled by default.

### DIFF
--- a/src/libprojectM/Common.hpp
+++ b/src/libprojectM/Common.hpp
@@ -30,6 +30,7 @@
 #include <typeinfo>
 #include <cstdarg>
 #include <cassert>
+#include "config.h"
 #ifdef _MSC_sVER
 #define strcasecmp(s, t) _strcmpi(s, t)
 #endif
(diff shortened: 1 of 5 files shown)


revmischa:
libftgl needed for OSX travis build https://travis-ci.org/projectM-visualizer/projectm/jobs/362786185

revmischa:
shipit!

src/libprojectM/Common.hpp
```diff
@@ -30,6 +30,7 @@
 #include <typeinfo>
 #include <cstdarg>
 #include <cassert>
+#include "config.h"
```
revmischa: in re https://github.com/projectM-visualizer/projectm/pull/53#discussion_r180357094 -
is this necessary in Common.hpp? this header is (probably) needed to be "public" and shouldn't include config.h which is "private" to the project

deltaoscarmike: It could be deleted, I should have defined "USE_FTGL" with a cflag instead.

PR #53, opened 2018-04-09
install additional headers needed to compile against libprojectm

walshbp:
install additional headers needed to compile against libprojectm

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,9 +17,6 @@ pm_font__DATA = fonts/Vera.ttf fonts/VeraMono.ttf
 pm_shaders__DATA = src/libprojectM/Renderer/blur.cg \
                    src/libprojectM/Renderer/projectM.cg
 
-# system headers/libraries/data to install
-include_HEADERS = config.h
-
 # find and install all preset files
 install-data-local:
 	test -z $(pkgdatadir) || $(MKDIR_P) $(pm_presets_dir)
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -17,6 +17,9 @@ pm_font__DATA = fonts/Vera.ttf fonts/VeraMono.ttf
 pm_shaders__DATA = src/libprojectM/Renderer/blur.cg \
                    src/libprojectM/Renderer/projectM.cg
 
+# system headers/libraries/data to install
+include_HEADERS = config.h
```
revmischa: It's preferable to not need to install config.h:
https://sourceware.org/autobook/autobook/autobook_66.html
https://stackoverflow.com/questions/19586211/correct-installation-of-config-h-for-shared-library-using-autotools?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

src/libprojectM/Makefile.am
```diff
@@ -12,7 +12,7 @@ AM_CPPFLAGS = \
 	$(FTGL_CFLAGS) $(CG_CFLAGS)
 
 # system headers/libraries/data to install
-include_HEADERS = projectM.hpp
+include_HEADERS = projectM.hpp Common.hpp dlldefs.h event.h fatal.h PCM.hpp
```
revmischa: 👍

walshbp: So Common.hpp includes config.h as of 4/5.  Commit log says "Add FTGL configure flag".  I can remove config.h from Common.hpp but I would break whatever was being tried with the FTGL configure flag.   For places that need this flag, could we include config.h at the lower level files?

revmischa: I think that sounds like a fine idea, yes 👍

revmischa: https://github.com/projectM-visualizer/projectm/pull/48/files#r180543415

PR #59, opened 2018-05-10
Expression evaluation performance

mbellew:
I saw lots of comments about expression evaluation being performance critical, so I took a look...
I changed eval() to a virtual function, and added an optimization step to get rid of extra nodes (E.g. TreeExpr wrapping a GenExpr).  I think most of the performance gain is from avoiding a lot of if/switch statements by creating more specific subclasses.

Profiles show a pretty significant improvement, FPS shows a more modest improvement.  I'll let you decide...


mbellew:
Working on doing more careful perf comparision...

src/libprojectM/MilkdropPresetFactory/Expr.cpp
```diff
@@ -92,102 +68,243 @@ float PrefunExpr::eval_prefun_expr ( int mesh_i, int mesh_j )
...
-		if (infix_op == Eval::infix_mult)
-			right_arg = 1;
+	// this is gratuitious, but a*b+c is super common, so as proof-of-concept, let's make a special Expr
+	if (infix_op->type == INFIX_ADD && 
+		((left->clazz == TREE && ((TreeExpr *)left)->infix_op->type == INFIX_MULT) ||
+		(right->clazz == TREE && ((TreeExpr *)right)->infix_op->type == INFIX_MULT)))
+	{
+		Expr *a, *b, *c;
+		if (left->clazz == TREE && ((TreeExpr *)left)->infix_op->type == INFIX_MULT)
+		{
```
revmischa: neato 💯

mbellew:
OK. I might make a change to be conservative on that check.

BTW, check out this branch I made to teach myself SSE.  It makes PresetOutputs::PerPixelMath() much  much faster.  Not sure it's worth adding to the main code base, however.

revmischa:
Cool, could be worth adding as long as it's conditionally compiled. Is it doing SIMD ops? Kind of hard to tell what's going on without being familiar with SSE and with no comments.

PR #62, opened 2018-05-20
SSE version of PresetOutputs::PerPixelMath

mbellew:
tested using 320x240, yin - 315 - Ocean of Light (yo im peakin yo Eo.S.-Phat).milk just about doubled FPS from 60ish to 110ish.

From google-pprof, doesn't entirely match up with my FPS observation, but does show CPU usage going down.

before
     881  48.6%  48.6%      881  48.6% __iscanonicall
     453  25.0%  73.7%     1225  67.6% PresetOutputs::PerPixelMath
after
     837  47.0%  47.0%      837  47.0% __iscanonicall
     131   7.4%  54.4%      965  54.2% PresetOutputs::PerPixelMath_sse

### DIFF
--- a/src/libprojectM/wipemalloc.cpp
+++ b/src/libprojectM/wipemalloc.cpp
@@ -27,7 +27,8 @@
 #include "wipemalloc.h"
 
  void *wipemalloc( size_t count ) {
-    void *mem = malloc( count );
+    count = (count + 15) & ~(size_t)15;
+    void *mem = aligned_alloc( 16, count );
     if ( mem != NULL ) {
         memset( mem, 0, count );
       } else {
--- a/src/libprojectM/wipemalloc.cpp
+++ b/src/libprojectM/wipemalloc.cpp
@@ -47,7 +47,7 @@
 
 void *wipe_aligned_alloc( size_t align, size_t size )
 {
-#if TARGET_OS_MAC
+#if __APPLE__
     // only support powers of 2 for align
     assert( (align & (align-1)) == 0 );
     void *allocated = malloc(size + align - 1 + sizeof(void*));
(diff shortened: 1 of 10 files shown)


revmischa:
Build is failing on OSX https://travis-ci.org/projectM-visualizer/projectm/jobs/381463843

mbellew:
I have an old iMac mini around somewhere I can blow the dust off and fire up...

revmischa:
If it's easier you can make changes and push them and then look at travis. It takes 5-10 minutes to run through usually.

src/libprojectM/wipemalloc.cpp
```diff
@@ -25,21 +25,61 @@
...
+    if (allocated == NULL)
+    {
+        printf( "wipe_aligned_malloc() failed to allocate %d bytes\n", (int)size );
+        return NULL;
+    }
+    void *ret = (void*) (((size_t)allocated + sizeof(void*) + align -1) & ~(align-1));
+    *((void**)((size_t)ret - sizeof(void*))) = allocated;
+    return ret;
+#else
+    void *mem = aligned_alloc( align, size );
```
revmischa: is aligned_alloc available on EVERY other platform that isn't apple? raspi? BSD? windows?
I know mac is special and needs ^2 alignment, that's fine. i'm concerned about the aligned_alloc being portable.

mbellew: I honestly don't know.  It is C11 standard, and the only google references I found about lack of support seem related to OSX.  I could remove the #ifdef and always use the hand-written version.

revmischa: Well do whatever you think is best

exp: MacOS seems to contain posix_memalign, would that not be sufficient? It seems to work (almost) exactly the same except takes an address to write the pointer to.

mbellew: There's this from stackoverflow

revmischa:
I have one error compiling w/ xcode:

screen shot 2018-05-22 at 22 03 23

PR #26, opened 2018-02-02
Shaders using GLSL not Cg/HLSL

revmischa:
Trying to get rid of super ancient unsupported nVidia Cg framework and use standard GL shaders.
The shaders in preset files are written in HLSL for DirectX. projectM used a toolkit from nVidia called Cg to handle them, but Cg is long deprecated and obsolete.
I'm attempting to use https://github.com/Thekla/hlslparser to transpile from HLSL to GLSL on the fly, and it's close to working but has some syntax issues.

The main conversion is done in:
https://github.com/projectM-visualizer/projectm/pull/26/files#diff-6df95f02d59c83baf42ed81c8668722a
https://github.com/projectM-visualizer/projectm/pull/26/files#diff-f4b5c1904e3d8fa6c1248fe1ac843d79

This may be close to working but it has issues right now. Needs work.

PR #66, opened 2018-06-02
OpenGL modernization

revmischa:
This was an attempt to get GLES3 compatibility but #70 did it a lot better. There are some improvements from this branch I need to port over though.


revmischa:
Changes are in #26

PR #70, opened 2018-06-08
GLES3

revmischa:
Modern openGL, compatible with GLES3. Huge ups to @deltaoscarmike 👍 👍 

Needs testing. 

GLSL (#26) coming soon


src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -81,36 +83,86 @@ Renderer::Renderer(int width, int height, int gx, int gy, int texsize, BeatDetec
...
+    glGenBuffers(1, &m_vbo_Interpolation);
+    glGenVertexArrays(1, &m_vao_Interpolation);
+
+    glBindVertexArray(m_vao_Interpolation);
+    glBindBuffer(GL_ARRAY_BUFFER, m_vbo_Interpolation);
+
+    glEnableVertexAttribArray(0);
+    glVertexAttribPointer(0, 2, GL_FLOAT, GL_FALSE, sizeof(float)*4, (void*)0); // Positions
+
+    glDisableVertexAttribArray(1);
```
revmischa: i'm still learning this stuff so forgive my ignorance -
curious why this is index 1? after we just enabled index 0? and then the textures are index 2? what's index 1?

psperl:
Is this code a work in progress or is it fully functional with the rendered output matching the current master?

I'm doing a quick code comparison against the GLES2 Android/iOS Renderer and this is quite different.  That could be because I changed a ton of stuff in the years it's been closed source, or it could be that @deltaoscarmike is just a better graphics developer.

src/libprojectM/Renderer/Renderer.cpp
deltaoscarmike: The vertex shader (v2f_c4f_t2f_vert) used for interpolation takes 3 parameters:
```
layout(location = 0) in vec2 vertex_position
layout(location = 1) in vec4 vertex_color
layout(location = 2) in vec2 vertex_texture
```

`glEnableVertexAttribArray(0)` map an array of positions for the first parameter of the shader and `glVertexAttribPointer(0...` describe the array data format.

`glDisableVertexAttribArray(1)` allows to map a single color value for the second parameter of the shader, this color value is set by `glVertexAttrib4f(1, 1.0, 1.0, 1.0, pipeline.screenDecay)` in Renderer::Interpolation

The third parameter (textures coordinates), it is similar to the position array.

deltaoscarmike:
@psperl
This code is finished and tested and should be fully functional, however **Cg and FTGL** support have been dropped due to openGL incompatibilities. I've tested this code only on Debian with an Intel GPU.

How did I proceed:
- Disable openGL compatibility profile (this remove deprecated openGL API like glMatrix*)
- Activate openGL debug messages
- Adapt every bit of code until there is no openGL messages complaining about using deprecated features
- Finally, testing every preset parameters to visually compare with the original code 

Performance wise, I'm a little disappointed because with vblank unlocked, I've benchmarked a very small fps improvement due to this port (tested with Idle preset; 5 second test run)
Original (without GLES) ~5300 frames drawn
Port  (without GLES): ~5500 frames drawn
Port  (with GLES): ~5450 frames drawn

I would really appreciate to compare our GLES implementations if you disclose yours and I think I will not be the only one to be obliged if you do it :yum:

revmischa:
My long-term dream is to figure out if there's a way to evaluate the preset equations inside of shaders so all the work can be moved from CPU to GPU.

revmischa:
I want to test this code out but it's crashing when calling drawArrays in Renderer::Interpolation(). On OSX/xcode SDL.

revmischa:
Looks great! Need to do a new release soon

einsteinx2:
Holy crap, definitely did not expect to see this today when I opened my GitHub notifications! Thanks a million for your work @deltaoscarmike!!

growingant:
Dear sirs：
I want to cross compile projectM as a static library on ubuntu(x86-64bit) so that android can use it. I use 
standalone-toolchain from android NDK in order to use projectM "configure,make,make install" tool sets. But error "configure: error: Could not find a valid OpenGL implementation" occurred during configure script running(enable SDL and gles) .
Is projectM now support gles? How could i port projectM to android platform?

revmischa:
Try `--enable-gles` with configure

growingant:
> Try `--enable-gles` with configure

I had already tried enable gles yesterday(./configure enable_gles=yes enable_sdl=yes --host=${CROSS_COMPILE} --disable-shared --prefix="${ANDROID_NDK_HOME}/Pre-built/projectM");
Today i tried again as you recommended(./configure --enable-gles --host=${CROSS_COMPILE} --disable-shared --prefix="${ANDROID_NDK_HOME}/Pre-built/projectM").
Both with the same error(configure: error: Could not find a valid OpenGL implementation)
There is no OpenGL implementation in android and I wonder how to port projectM to android? Now i am    working on a demo using ProjectM on android. Thanks very much

deltaoscarmike:
Autotools doesn't detect openglES properly when there is no opengl compatibility.
Comment the [macro](https://github.com/projectM-visualizer/projectm/blob/6240b9d14a2efa9a5a5a4bdbf33406e51ad53c9e/configure.ac#L37) with a `#`.
If problem persist please use pastebin to report your config.log file.
Also, sdl should not be useful for an android build and, morever, for a static library.

growingant:
[config.log](https://github.com/projectM-visualizer/projectm/files/2378418/config.log)

I commented AX_CHECK_GL, but got error related with OpenGL 

> configure: error: libglm is required.

By the way, I had compiled sdl source to share library using android studio, and i want to use sdl and projectM static library for a test. I am not gonna to compile sdl example by ubuntu for an android build.

deltaoscarmike:
Fetch [libglm](https://github.com/g-truc/glm/releases/tag/0.9.9.1) and install it into your crosscompile environnement.

growingant:
> Fetch [libglm](https://github.com/g-truc/glm/releases/tag/0.9.9.1) and install it into your crosscompile environnement.

It means i should compile and install ` autoconf libtool libsdl2-dev libglew-dev libftgl-dev libsdl2-dev libdevil-dev libglm-dev` into my cross compile environment, just the same like doing it in ubuntu??That is a lot of work. Can you recommend me a better solution?

deltaoscarmike:
> It means i should compile and install autoconf libtool libsdl2-dev libglew-dev libftgl-dev libsdl2-dev libdevil-dev libglm-dev into my cross compile environment, just the same like doing it in ubuntu??That is a lot of work. Can you recommend me a better solution?

I don't have a better solution, libglew and libdevil are not used anymore and libftgl is used only with the --enable-ftgl configure flag. You should really get rid of sdl for an android app, the first thing you need to do is initializing an opengl context (that's what sdl do in a sdl sort of way).

> Question: can configurefile generated by autogen.sh on ubuntu be directly used for a cross-compile for an android target?? I think i have to invoke autogen.sh on android platform so that generated configure file can be valid for an android phone. Can autogen.sh be used on android platform?

Autogen is done by and for the host system, not for the target. It's configure which handles the specifics of your target.

See this [example](http://www.learnopengles.com/calling-opengl-from-android-using-the-ndk/) for a starting point. Applying this example for a very basic projectm integration should look like:
* `on_surface_created()` call projectM constructor
* `on_surface_changed()` call projectM::projectM_resetGL
* `on_draw_frame()` call projectM::renderFrame

Good luck.

deltaoscarmike:
Yes, the readme should be updated, can't do it right now.
liglm integration is prepared in 3b29e22.

growingant:
@deltaoscarmike Thanks very much for your kindly reply!
I found that there is glm source files existing in android ndk, and i added that dir to be included by CFLAGS. It seems that there is no error about glm, but another error occurred  "GL/gl.h: No such file or directory". Is this error related to opengl? I can only find gl.h in GLES directory of android ndk. As far as i know, android is now using gles2 or gles3. 
How to solve this?

growingant:
configure trace: https://pastebin.com/QZs02vJu
compile trace: https://pastebin.com/W5DQCVUH

Thanks a lot!

deltaoscarmike:
You're right this lines are missing in makefiles, please make a PR for these.

Presets files should be embedded within your app assets. See [this example](https://github.com/Suvitruf/Android-ndk/tree/master/NDK-Writing-files-from-native-code) on how to access files. The part that matters in this example is the java code providing the data path for native code.

However, presets loading is not a important. Displaying projectm idle preset (which is embedded in c++) should be the first step (and it's a big one) before taking care of other presets.

growingant:
I had already made PRs.

For presets handling, Can i store presets files to an android external storage area and set that directory URL to projectM `_settings.presetURL`, just the same like in utuntu, so that my projectM demo can access this presets?

For the [example](https://github.com/growingant/Android-ndk/tree/master/NDK-Writing-files-from-native-code) you recommended, I searched by google and found that files under assets are compressed in .apk file, so i can not get presets absolute file path. Any suggestion?

Thanks a lot!

revmischa:
@growingant check out what the osx itunes plugin does: https://github.com/projectM-visualizer/projectm/blob/master/src/projectM-iTunes/iprojectM_mac.mm#L170 maybe you can do something similar in android?

growingant:
I had made a projectM demo on android base on android ndk samples from google -` native-activity`, but have problems for testing. 

Dose current projectM only support hardware with opengl es3? I cannot compile my demo when linking against  GLESv1_CM, while i can not run demo compiled linking against GLESv3 on android phone with GPU that only support opengl es1(qualcomm gpu can tell something like "api call invalid").

revmischa:
The main issue I'm aware of with GLES versions is the supported shader versions. We need to target a particular GLSL version, I'm not sure what the compatibility is with different ES versions. If you can work that out maybe it can work with GLES1? I have no idea

deltaoscarmike:
Projectm is compatible with opengl ES starting at 3.0 and up, but it is not compatible with opengl ES 1.x.
OpenGL ES 3.0 is backwards compatible with OpenGL ES 2.0.

growingant:
I think i can use opengl es3.0 in android now(GPU is supporting opengl es 1.0,2.0,3.0, and i choose version 3.0  to use), but the whole android phone screen is just black. I modified sdl programs running on x86 ubuntu(disable audio capture, does not set preset URL), and there did appear idle preset when running(a lot of rotating projectM logs)
My demo is almost the same with sdl example. Sdl example can run on x86 ubuntu with opengl, but my demo can not run on android with opengles3.0. I think maybe problem is related with opengl part.
My [demo ](https://pastebin.com/xbyHm9d2)is quite simple, hopefully your sirs can check for me (i cross compiled projectM on x86 ubuntu to static library and use it in android studio)

growingant:
Finally my demo can run with idle presets on android! I have to call `eglSwapBuffers` after `projectM::renderFrame()`. I thought `eglSwapBuffers` had already embedded into projectM library.

But there is one question: I can not run my demo with full screen on android with [this](https://pastebin.com/wkpEAx4K) configuration. My phone resolution is 2048*1536

growingant:
Thanks a lot! @deltaoscarmike @revmischa 
Finally i got projectM running on android with presets and audio capture on!

I have done a very basic projectm integration on android following @deltaoscarmike  [advice](https://github.com/projectM-visualizer/projectm/pull/70#issuecomment-421458971), but pictures drawing by projectM is not amazing and sometimes totally black for a while. Would you pls give me some advice or tutorials for advanced projectM usage so that i can make amazing visualization just like official projectM app on android app store.

revmischa:
Sure, the visuals are based on whatever presets you're using. I'm trying to put together a pack some cool ones that I like.
try em out https://mischa.lol.s3.amazonaws.com/projectM/presets-revmischa-v1.zip

growingant:
@revmischa Access to https://mischa.lol.s3.amazonaws.com/projectM/presets-revmischa-v1.zip is denied and it seems i do not have the permission. Would you pls send it to me wangfn@visionox.com

Thanks a lot!

revmischa:
@growingant  ffs sorry try http://mischa.lol.s3.amazonaws.com/projectM/presets-revmischa-v1.zip

growingant:
@revmischa There are two folders there: `__MACOSX` & `presets mish`. Which one should i use on android?

revmischa:
presets

PeterMalkin:
@growingant  Is the android port you have produced open source? Is there a chance you could publish the code? 

If not, did you happen to make notes on how you proceeded, and may be you would be willing to share those?

revmischa:
i think only minor changes are needed to use with GLES2 - the shaders need to be slightly tweaked

PeterMalkin:
@revmischa I was asking more for building libprojectM inside ndk. It seems like @growingant  has got it to work, and I was wondering if he was willing to share the details.

StefanBruens:
There are still 2 small parts which require Desktop GL:
- SOIL2, can be done by defining SOIL_GLES2: https://github.com/projectM-visualizer/projectm/blob/6cb87ee6990c4c136f305134b823cf729ae339dd/src/libprojectM/Renderer/SOIL2/SOIL2.c#L40
- libprojectM-qt, e.g. glShadeModel: https://github.com/projectM-visualizer/projectm/blob/6cb87ee6990c4c136f305134b823cf729ae339dd/src/projectM-qt/qprojectmwidget.hpp#L272-L299

revmischa:
PRs very welcome!

PeterMalkin:
I seemed to have managed to build with NDK and uploaded for PR. But I still hear nothing from @psperl and @growingant. So if those guys are silent, someone will have to reimplement an Android JNI wrapper to hook up libprojectM into and android app? I'd hope not to have to do the same thing twice (thrice?) :/

psperl:
My branch of projectM is pretty different at this point because the
conversion to GLES and GLSL was done independently by each effort. In any
case, the JNI layer needed is pretty simple. There's not much to it. Once
you understand how projectM integrates with any opengl app, and how jni
works, it's an hour or two.

revmischa:
@StefanBruens if you know what to do here and could send a PR for that, that would be super helpful and amazing

PeterMalkin:
Ok, I spent the last three days staring at it, I think I got to a point where all the glue code is written, the libprojectM is built and linked properly. I had to also include libc++_shared.so into the gradle.build - otherwise it would not link.

However, it seems I am still not properly linking with egl. I imagine this should be done by the dynamic linker on the target, and egl should come as a shared library on the system. But all I get is 

E/art: dlopen("/data/app/com.github.projectm_android-2/lib/arm/libprojectM.so", RTLD_LAZY) failed: dlopen failed: cannot locate symbol "eglGetProcAddress" referenced by "libprojectM.so"...

You guys have any pointers?

PeterMalkin:
@psperl  If I am not mistaken, this project is LGPL, which means you need to release your GLES port, since you have made modifications to libprojectM

@revmischa I tried the current implementation of libprojectM for GLESv2 android devices. There's a number of gl calls that are made that are not supported, so I think for older devices with no GPU implementation its a no-go.

PeterMalkin:
@mancoast 

I think there's another github issue on that: https://github.com/projectM-visualizer/projectm/issues/168

I did get GLES3 to render on android. I am still not 100% done with the port. 
None of the v2 presets with shaders work. Some other presets get stuck as well. 

On GLES2 hardware the library fails with "E/libEGL  ( 2235): called unimplemented OpenGL ES API"
I traced the call to the GLES library, and the "glSampler*" functions are the first culprit, but who knows what else is used from GLESv3 that's not supported.

It seems like @psperl did a GLES2.0 port a few years before @deltaoscarmike's GLES3.0 port. But for some reason he keeps it closed sourced, even though it is violation of the license, since his apps are for sale on the app stores.

mancoast:
@psperl can you share the branch with GLES2? It would be helpful for community. Thanks!

psperl:
@PeterMalkin Carm and I reverse engineered Milkdrop and created projectM in our dorm rooms over the course of a year and a half starting in 2003.  We released a fully functional LGPL-licensed libprojectM with multiple ports (xmms, jack, etc.) in 2004.  We are the original "projectM Team," and are the copyright holders of libprojectM, at least up until community contributions started recently.  As such, libprojectM has 2 main forks: The public LGPL branch described above, and a private branch which meaningfully diverged a decade ago and can be licensed at our discretion.

I added support for GLES to our private fork back in 2010 or so after no community formed after 6 years of libprojectM being available with an open-source license.  I can't use the code in github now as others have added to it and complicated the copyright. For my purposes, the public branch is useless and that's why I haven't contributed much over the last decade.  

@mancoast The non-public branch of projectM diverged from the code you see in github a very very long time ago.  The code you're asking for would not be useful if applied to this branch.  The GLES code was written differently, especially the shader portion.  It would be more efficient to start from scratch.

All this excitement around open-source projectM might be the catalyst that finally motivates me to resurrect the VR branch.  If only the VR APIs weren't so obtuse and I wasn't so lazy.

mancoast:
@psperl Any plans to release GLES2 branch?
doesn’t have to be your latest and greatest with chromcast, etc. 

maybe just release an early functioning revision that we can run with for merge back upstream. This would fix raspberry pi and other embedded platforms using LGPL projectM codebase.

PeterMalkin:
@mancoast 

I made a PR with my current branch of the android port. It could definitely use some attention, but I felt I'd rather share earlier than later. https://github.com/projectM-visualizer/projectm/pull/191

Or if the maintainers would have objections and wont want to accept - you can just follow my fork.

teotigraphix:
Regardless of contribution I will probably attempt the VBO/GPU stuff down the road, if the OS vibe dies on this thread. Reinventing this wheel sucks but I will do what I need to do and contribute back to project M source for Android and iOS.

mancoast:
Greetings, I will be releasing a new iOS app shortly that reuses most of the Xbox app code. Everything here on this repo when it’s working if you want to focus energy on Android. Also, while in that realm, i wonder how chrome cast fits into the equation, that would be a huge contribution to this repo.

teotigraphix:
Hey @mancoast I deleted my comment since I felt it was now irrelevant. Ha, now that you posted your comment it is. ;-)

Yeah, if you get the iOS stuff up I can be the Android guy. :)

I will check on Chrome Cast as well, I have it setup on my tablet and phone.

PeterMalkin:
@teotigraphix 
Excellent! Great to hear. 
Everything should be more of less working for Android. I tested on a number of systems. So far the biggest pain is that the device has to support openGLES3.1. And some presets are not rendered correctly. So may be at some point we could test all presets to verify which are working and which arent.

From my testing it seems like @psperl port works better - since its GLES2.0 based, and covers more devices.

teotigraphix:
@PeterMalkin Yeah I am excited, I am digging into the framework internals now and when I come up for air we will talk more. :)

> From my testing it seems like @psperl port works better - since its GLES2.0 based, and covers more devices.

Could yous expand a bit? I have only been looking at this code for 2 days so his "port", what code is that (just android parts or the whole opengel `Renderer` code)?

teotigraphix:
@PeterMalkin This issue is confusing, thus my questions above. I see the GLES2 comments but as far as if it exists yet? That is where I am lost.

mancoast:
in essence, the original author of this code, @psperl, rebuilt projectm. one branch was release to public under LGPL. the other branch with GLES2 support was not.   Now this LGPL branch has GLES3 support.

teotigraphix:
Alright, got it thanks @mancoast , completely clear. So it's the wagon in front of the horse right now.

So essentially, the `Renderer` needs to implement GLES2, we don't have it, someone has it and we have to write it.

So is this and instance with define blacks and such, has it been talked about if we have to do it without that port?

psperl:
Does the 3.1 port actually use 3.1 features? You could probably downgrade
to 2.0 somewhat easily.

teotigraphix:
@psperl Cool I will check it out.

teotigraphix:
>  The OpenGL ES 3.0 specification was publicly released in August 2012. OpenGL ES 3.0 is backwards compatible with OpenGL ES 2.0, enabling applications to incrementally add new visual features to applications.

The problem is I am not an OpenGL guru yet. :) So I guess I will have to defer help on other areas, since I couldn't trust myself to know the actual correct answer atm.

Although, if I can be pointed in the right direction by others that may not have the time but knowledge, it could happen as well.

teotigraphix:
I ran a really heavy preset on a Pixel, butter smooth. 

Nexus 7, eh as it stands it has problems but still works.

teotigraphix:
For reference; https://www.khronos.org/registry/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.pdf

psperl:
Just change all the headers and defines to GLES2 and see what's stops
compiling. Then look at the docs and replace the new gles3 calls with the
older ones. Most should be the same, so i imagine there isn't much to do.

teotigraphix:
Ok, thanks @psperl I appreciate you advice.

Sounds like a plan to me!

mancoast:
I have one quick question about the OpenGL programming. Why does the system require text strings for shaders, etc? Is there any way to redline the shaders?

psperl:
I think the GPU/GL developers couldn't agree on an assembly or bytecode
format that would work for all GPU families now and in the future, so they
standardized on a high level language and left the compiler and machine
code as an implementation detail for the GPU driver.  At the end of the
day, all code is just string input to a compiler, it's just more in your
face here.

PeterMalkin:
@psperl 

I tried to do what you suggest with GLES2.0, but got stuck pretty quickly. It builds fine, since GLES is supposed to be backwards compatible. But it fails during dynamic linking. The functions are defined in opengles.so, but some (the 3.1 specifics) are not implemented by the hardware manufacturer (I was using a device with a mali gpu).

The place I got stuck - on TextureSample*() functions.

revmischa:
There's an open issue to support different GLSL versions - https://github.com/projectM-visualizer/projectm/issues/168 suggest moving discussion there.
I think part of the problem is that we have to not only support the built-in shaders (easy enough to modify) but also the shaders in presets, which are transpiled from HLSL to GLSL, and I'm not sure if that bit needs to be modified too to support older GLSL versions

PeterMalkin:
@revmischa 
I dont think its just GLSL. At least when I tried, there were regular GLES API functions used that were not present in GLES2. Doesn't mean there's nothing to do for GLSL of course, but its not the only thing missing

revmischa:
I'm sure that is true

PeterMalkin:
@psperl 
Peter, is there a chance you could email me? I tried reaching you earlier through the email published on the Android Play Store account, but I think you may have missed the email. Here's my address: https://tinyurl.com/y44vf9aw

stuaxo:
Reading this thread made me wonder what the differences between GLES 2 and 3 are, theres a good summary here:
https://stackoverflow.com/questions/17543364/what-are-the-differences-between-opengl-es-2-0-and-opengl-es-3-0

PR #74, opened 2018-06-18
Add double-click for fullscreen, right click for menu to projectm-qt

aots-aots:
This is a feature I always thought was missing from projectm-qt/pulseaudio/jack. Pretty straightforward:

Double left click on the visualisation to enter/leave fullscreen mode (same as pressing F key). Right click to show/hide them menu (same as M key).

Important so that the main features and functionality of the GUI are available using the mouse only, saving me reaching for the keyboard on the coffee table. Fullscreen and menu are much more discoverable with this as well -- which was a problem for new users I believe.


revmischa:
Why does this diff show all these other patches?

revmischa:
@aots-aots happy to merge this if it's cleaned up... the diff is rather odd

axe312ger:
I guess a simple rebase could help here!

PR #75, opened 2018-06-21
Minor fixes

luigino:


src/libprojectM/Makefile.am
```diff
@@ -14,7 +14,7 @@ AM_CPPFLAGS = \
 lib_LTLIBRARIES = libprojectM.la  # public, possibly-shared library
 
 # link flags
-libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 0:2:0
+libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 2:0:0
```
revmischa: shouldn't version info be coming from pkg_version? or is this separate from package version

luigino: That's the version of the library, which was 2.0.0 until the last sourceforge release.

Only you know if the API has been changed, but "-version-info 0:2:0" produces a libprojectM.so.0.0.2: was that the original intent?

revmischa: probably not

PR #92, opened 2018-08-02
Glsl emscripten

axe312ger:
* add glm as dependency so users don't have to grab it on their own
* add some finding i had on another machien to the readme to help others
* fix makefile with our latest findings


src/projectM-emscripten/README
```diff
@@ -1,27 +1,39 @@
...
 
-Does not work? Try:
+## Troubleshooting
 
-* `mkdir vendor && cp /usr/local/Cellar/glm/xx/include/glm ./vendor/`
-* `emmake make -j4 -lvendor`
+### General
+
+Want to restart the process after pulling or changing config?
+* `rm src/projectM-emscripten/projectW*`
```
revmischa: should be a clean target

src/projectM-emscripten/README
```diff
@@ -1,27 +1,39 @@
 Running on the web.
 
-# Easy way:
+# TLDR:
+* `./autogen.sh`
+* `emconfigure ./configure --enable-emscripten --enable-gles --enable-sdl`
```
revmischa: i tried making it so all you need is `--enable-emscripten` now and it infers the others

src/projectM-emscripten/README
```diff
@@ -1,27 +1,39 @@
...
+* `emmake make -j4 -lvendor`
```
revmischa: is -lvendor needed? that should be in CPPFLAGS or whatever.. `-I`

src/projectM-emscripten/README
axe312ger: Is there any way to ensure this like a cleanup script?

revmischa:
@deltaoscarmike chill?

deltaoscarmike:
As is, it doesn't work: 
- FF displays the page but PM doesn't start PM(output log is empty) and finally it freeze
- Chrome displays the page, but PM doesn't start (output log is empty)

You've got it working ?

revmischa:
No, I never got it to fully run. But it's not a regression either; emscripten doesn't work in glsl now. It just gets a little closer to maybe working now
@axe312ger do you know what the current problem is? is it still GLSL versions?

deltaoscarmike:
Oops, i've made a mistake while trying to push a commit on this branch, please delete branch PR92.
Oth, i've got it working (sort of), try this [patch](https://gist.github.com/deltaoscarmike/228535d73718f8abdf851681c68c4cd1)

axe312ger:
Thank you, I gonna have a look!

axe312ger:
Thats my Chrome console output. (It freezes, make sure to open the console first, then navigate to the server running emscripten

screen shot 2018-08-28 at 13 39 26


I guess the shaders are still broken / using the wrong types. I tried to fix this 2 weeks ago but no, no luck. (Well fixing as of randomly changing values that seem to be related and 🤞 )

```
Failed to compile shader 'Fragment: blur1'. Error: ERROR: 0:39: 'assign' : cannot convert from 'const int' to 'mediump float'
Failed to compile shader 'Fragment: blur2'. Error: ERROR: 0:30: '-' : wrong operand types - no operation '-' exists that takes a left-hand operand of type 'const int' and a right operand of type 'mediump float' (or there is no acceptable conversion)
```

I wonder how to add presets to the wasm as well:
```
[PresetLoader] ENOENT error. The path "/presets" probably does not exist. "man open" for more info.
```

Thanks for your help 👋

deltaoscarmike:
WebGL is rather strict concerning implicit casts, this [patch](https://gist.github.com/deltaoscarmike/c840cfbe41219a0513f8f9a2af8c3fcb) will fix blur shaders, re-enable WASM and print root content of the virtual filesystem.
For the preset directory issue, i think this [link](http://kripken.github.io/emscripten-site/docs/api_reference/Filesystem-API.html) could be useful on how to embed presets.
My priority is to get a rendering closer to Milkdrop/winamp and that's lot of work :wrench:
Good luck

axe312ger:
It works now!!! Thanks so much @deltaoscarmike 

screen shot 2018-08-29 at 11 25 12

axe312ger:
I fixed the merge conflicts :)

axe312ger:
Still need to read into the file system api for emscripten to get the presets running

Issue #95, opened 2018-08-17
Cleanup ?


kyrios123:
I have just built the version 3.0.0 and when I look at what has been installed I see things like this:
```
/usr/share/projectM/presets/!!!---flexi + amandio c - organic12-3d-2.milk
/usr/share/projectM/presets/.DS_Store
/usr/share/projectM/presets/._.DS_Store
/usr/share/projectM/presets/._Aderrasi + Geiss - Airhandler (Kali Mix) - Canvas Mix.milk
/usr/share/projectM/presets/._Aderrasi - Accelerator (Hot Lead Transfusion).milk
/usr/share/projectM/presets/._Aderrasi - Arcworld (Arcwheel - Train Wheel).milk
/usr/share/projectM/presets/._Aderrasi - Calabi-Jau Space Bar.milk
/usr/share/projectM/presets/._Aderrasi - Contortion (Flair Mix).milk
/usr/share/projectM/presets/._Aderrasi - Potion of Spirits.milk
/usr/share/projectM/presets/._Eo.S.+Phat Fractical_dancer - light in the distance.milk
/usr/share/projectM/presets/._Eo.S.+Phat Quadrent_fractal.milk
/usr/share/projectM/presets/._Eo.S.+Phat Speak with the orb_more_colour_mix.milk
/usr/share/projectM/presets/._Eo.S.+Phat last of it's kind_sinking.milk
/usr/share/projectM/presets/._Flexi - reality tunnel.milk
/usr/share/projectM/presets/._Flexi - smashing fractals [acid etching mix].milk
/usr/share/projectM/presets/._Fvese - Round and Round (geiss gamma mix).milk
/usr/share/projectM/presets/._Fvese - The Tunnel (Final Stage Mix).milk
/usr/share/projectM/presets/._Geiss - Aieeeeee!!!.milk
/usr/share/projectM/presets/._Geiss - Asymptote.milk
/usr/share/projectM/presets/._Geiss - Bass Kaleidosphere.milk
/usr/share/projectM/presets/._Geiss - Bass Zoom.milk
/usr/share/projectM/presets/._Geiss - Bipolar 1.milk
/usr/share/projectM/presets/._Geiss - Bonfire.milk
/usr/share/projectM/presets/._Geiss - Calligraphy.milk
/usr/share/projectM/presets/._Geiss - Cartographie.milk
/usr/share/projectM/presets/._Geiss - Cepiasound.milk
[...]

PR #96, opened 2018-08-18
Install native presets libraries in libdir

kyrios123:
As requested [here](https://github.com/projectM-visualizer/projectm/issues/95#issuecomment-413964324)
It seems to work fine on my computer but please check if this is okay because there was a comment saying 
```
# native presets are shared object files.
# they should get installed to $presetsdir/
```
Signed-off-by: Pierre-Yves <pyu@riseup.net>

### DIFF
--- a/src/NativePresets/Makefile.am
+++ b/src/NativePresets/Makefile.am
@@ -9,10 +9,8 @@ AM_CFLAGS = ${my_CFLAGS} \
 -ffunction-sections \
 -fdata-sections
 
-presetsdir = $(pkgdatadir)/presets
+presetsdir = $(libdir)/projectM
 
-# native presets are shared object files.
-# they should get installed to $presetsdir/
 presets_LTLIBRARIES = \
   libMstressJuppyDancer.la \
   libRLGFractalDrop7c.la \


revmischa:
Please check the code in NativePresetFactory and see where it loads from... I'm assuming it loads from the preset dir

kyrios123:
Eeuuhh well I'm not at all a c++ developer, so I fear I won't be of a big help on this.
I might be wrong, but if a library is not found on a specified location, it should fallback on the "standard library location" where the libraries were moved, isn't it?

I saw [this](https://github.com/projectM-visualizer/projectm/blob/master/src/libprojectM/projectM.cpp#L190) and [this](https://github.com/projectM-visualizer/projectm/blob/master/src/projectM-sdl/projectM_SDL_main.cpp#L131).
For the first one, the prefix I specified during configure is `/usr`, so I don't know if this path starting with `/usr/local` could be an issue.
For the 2nd one, I can't find the `presets_tryptonaut` anywhere ?

revmischa:
The second path is fixed in `glsl` https://github.com/projectM-visualizer/projectm/blob/glsl/src/projectM-sdl/projectM_SDL_main.cpp#L131

deltaoscarmike:
Actually, native presets (.so) are loaded like other presets (.milk) from the presets directory (/usr/share/projectM/presets). Thus moving native presets installation directory is not enough.
Native presets are not standard libraries, they're loaded explicitly and exclusively by projectM. 

Concerning the necessity of this pull request, i have remarks:
1. @revmischa [mentionned](https://github.com/projectM-visualizer/projectm/issues/95#issuecomment-413964324) the creation of a configuration option to disable native presets which make sense as these are not essential. For example, they're not even distributed in debian packages.
2. Native presets are very basic vs milkdrop ones, they're not worth the effort to move them into another directory.

As stated by linux Filesystem Hierarchy Standard in [4.11.1:](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html)
> The /usr/share hierarchy is for all read-only architecture independent data files

Native presets are not architecture independent, to resolve this i think it is easier to add a configure option (example: --enable-nativepresets) as raised in (1).

kyrios123:
I will abandon this PR is you don't mind because as mentioned, moving the presets will cause issues.

PR #112, opened 2018-10-04
added: allow specifying datadir in settings on runtime

notspiff:
i maintain the add-on (plugin) for kodi. i have patched this for years, but since this upstream is now alive i figured i'd try sending it back upstream.

this allows for specifying the datadir from settings, instead of the path hardcoded on compile time. this is necessary in our case, since the add-on bundles the lib and all the data, and it is not necessarily installed in a unix manner, but rather in user home and/or the kodi add-on directory.

i tried to do a no-impact solution for others, if you want to see it done otherwise, just yell, i did not spend much time/brain on this.


revmischa:
I totally support specifying datadir in settings instead of hardcoding it. The reason for it to be specified at compile time is that it needs to know where to read the settings file from in the first place... how do you suggest that be handled?

Let us know if there is anything else that would help out with Kodi! It's a great project.

notspiff:
i think the current handling is fine for the default, if none is specified, have a sane built-in default.

i did not add this to config files at all, only allowed it to be programatically specified. but if you want it from config files, that's of course a small addition.

src/libprojectM/Renderer/TextureManager.cpp
```diff
@@ -40,9 +40,11 @@ TextureManager::TextureManager(const std::string _presetsURL, const int texsizeX
     extensions.push_back(".dib");
 
     Preload();
+    if (datadir.empty())
```
revmischa: Add a helpful comment here

notspiff:
okay, i intended to squash the fixup, but thanks for merging!

Issue #115, opened 2018-10-15
Trying to build ProjectM on Raspberry Pi 3


swinokur:
Hello,
I'm trying to build ProjectM on a Rpi3, thinking that it would be cool to be able to bolt it onto "volumio" (https://volumio.org/)

The code builds on the rpi (ableit with a bunch of warnings). I've built SDL 2.0.8 (more on that later) and verified that SDL works using some test code. I've verified that there's GLES working using some of the test code in the rpi.

Running projectMSDL results in:
```
root@volumio:~# projectMSDL 
INFO: GL_VERSION: (null)
INFO: GL_SHADING_LANGUAGE_VERSION: (null)
INFO: GL_VENDOR: (null)
[projectM] config file: /usr/local/share/projectM/config.inp
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted
```

This clearly isn't optimal. :-)

Here's as much detail as I can think to provide, please let me know if more is needed, if there are other tests / code I can run, etc.

---
system environment:
```
root@volumio:~# uname -a
Linux volumio 4.14.69-v7+ #1141 SMP Mon Sep 10 15:26:29 BST 2018 armv7l GNU/Linux

root@volumio:~# more /etc/apt/sources.list
deb http://archive.volumio.org/raspbian/ jessie main contrib non-free rpi
deb-src http://archive.volumio.org/raspbian/ jessie main contrib non-free rpi

```

compiler version
```
root@volumio:~# g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/4.9/lto-wrapper
Target: arm-linux-gnueabihf
[...]

PR #123, opened 2018-11-17
DX to openGL/openGLES transpilation fixes

deltaoscarmike:
Improve GLES support: all presets with shaders are now successfully transpiled to openGL ES

It should fix one issue seen in #115

### DIFF
--- a/src/projectM-sdl/projectM_SDL_main.cpp
+++ b/src/projectM-sdl/projectM_SDL_main.cpp
@@ -8,8 +8,8 @@
 
 #include "pmSDL.hpp"
 
-#define FAKE_AUDIO	0
-#define TEST_ALL_PRESETS	0
+#define FAKE_AUDIO          0
+#define TEST_ALL_PRESETS    0
 
 #if OGL_DEBUG
 void DebugLog(GLenum source,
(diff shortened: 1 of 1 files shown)


src/libprojectM/Renderer/ShaderEngine.cpp
```diff
@@ -19,8 +19,10 @@
 
 #ifdef USE_GLES
     #define GLSL_VERSION "300 es"
+    #define SHADER_VERSION  M4::GLSLGenerator::Version_300_ES
```
revmischa: i think this still won't fly on a raspi will it?

src/libprojectM/Renderer/hlslparser/src/GLSLGenerator.cpp
```diff
@@ -1826,47 +1875,59 @@ void GLSLGenerator::OutputDeclaration(HLSLDeclaration* declaration)
...
+                globalVarsAssignements.push_back(declaration);
+            }
+        }
 
 		lastDecl = declaration;
 		declaration = declaration->nextDeclaration;
 	}
 }
 
+void GLSLGenerator::OutputDeclarationAssignement(HLSLDeclaration* declaration)
```
revmischa: Assignment, not Assignement

src/libprojectM/Renderer/ShaderEngine.cpp
deltaoscarmike: no, hopefully gles3 support will be implemented on broadcom vc4 driver someday.

src/libprojectM/Renderer/hlslparser/src/GLSLGenerator.cpp
deltaoscarmike: done

deltaoscarmike:
GL mode (desktop GL or GLES) should be detected on app launch and not rely on a compilation flag.

PR #125, opened 2018-12-02
Increase soversion to 2.2.0

luigino:
Right now 2.0.0 is even older than the original libprojectM before the refactoring.

### DIFF
--- a/src/libprojectM/Makefile.am
+++ b/src/libprojectM/Makefile.am
@@ -14,7 +14,7 @@ AM_CPPFLAGS = \
 lib_LTLIBRARIES = libprojectM.la  # public, possibly-shared library
 
 # link flags
-libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 4:0:2
+libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 4:1:1
 
 # link libRenderer, MilkdropPresetFactory, NativePresetFactory, and libprojectM sources
 libprojectM_la_LIBADD = \
--- a/src/libprojectM/Makefile.am
+++ b/src/libprojectM/Makefile.am
@@ -14,7 +14,7 @@ AM_CPPFLAGS = \
 lib_LTLIBRARIES = libprojectM.la  # public, possibly-shared library
 
 # link flags
-libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 2:0:0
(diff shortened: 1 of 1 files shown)


src/libprojectM/Makefile.am
```diff
@@ -14,7 +14,7 @@ AM_CPPFLAGS = \
 lib_LTLIBRARIES = libprojectM.la  # public, possibly-shared library
 
 # link flags
-libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 2:0:0
+libprojectM_la_LDFLAGS = $(CG_LDFLAGS) -no-undefined -version-info 4:0:2
```
revmischa: How about 3:0:0?
Also update configure.ac

luigino: In what ways has the API changed since then? As I understand, one should increase the major version only when compatibility with the past has been broken.

revmischa:
We should increase to 2.2.0. I don't understand the `4:0:2`

luigino:
To be honest, I don't understand [libtool's syntax](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html) either, but 4:0:2 produces a 2.2.0 soversion.

If you think that's the appropriate one instead of 3.0.0 (has there been backwards API breakage?) it shouldn't require any other change. What part of configure.ac requires updating?

revmischa:
just the top line in configure.ac which has been `3.1.0` for a long time... not sure if it affects anything
i still am really unclear how 4:0:2 results in 2.2.0. That makes no sense to me.

luigino:
Does the projectM version have to match the soversion?

revmischa:
no, but any reason it wouldn't?

luigino:
The soversion usually concerns the library API, not necessarily the main program. If the former has been left backwards-compatible, there is no reason to increase its major.
Also I might need to add another small patch to restore libprojectM-qt.so as it used to be in earlier versions (a certain package relies on it) and that was 1.2.0 in the last cmake release.

You tell me if either or both have changed in a significant way.

Anyway, with regard to libtool, after a few tests I believe the rule is:

given major.minor.patch and f1:f2:f3

f3 must never be higher than f1
major is the difference between f1 and f3    
minor is f3
patch is f2

e.g.
```
awk -F "." '{print $1+$2 ":" $3 ":" $2}'
2.2.0
4:0:2
2.2.1
4:1:2
2.3.0
5:0:3
```

revmischa:
> The soversion usually concerns the library API, not necessarily the main program.
There is no main program. It's all about the library. 

There are some changes to the -qt build in a branch https://github.com/projectM-visualizer/projectm/pull/163 which will be merged soon

PR #137, opened 2018-12-29
rgb and rgba support

mbellew:
First off, my approach to this lead to one architectural change.  I changed Expr::eval() from float to double so that it could encode a color (a double can encode a 48 bit integer thus a 12 bit rgba color, actually I think 13 bit would have also worked).

see packRGB() and unpackRGB()

So if we don't like that, then we need a different strategy.

All these should work
ib_rgb = #ff00ff
ib_rgba = #ff00ff70
ib_rgb = rgb(1.0,0.0,1.0)
ib_rgb = rgba(1.0,0.0,1.0,0.1)
equations also work, note unlike css, the values are 0.0-1.0 not 0-255
per_frame_1=ib_rgb= rgb(0.5 + 0.5*sin(time), ... )

That the code for handling rgb() and rgba() for initcond and equations is different.  InitCond is handled in Parser.cpp, equations are handled as two new built-in functions.


src/libprojectM/MilkdropPresetFactory/BuiltinParams.cpp
```diff
@@ -248,6 +248,39 @@ std::transform(lowerName.begin(), lowerName.end(), lowerName.begin(), tolower);
...
+  Param *a = nullptr==name_a ? nullptr : builtin_param_tree[name_a];
+
+  if (nullptr == r || nullptr == b || nullptr == g || (nullptr == a && nullptr != name_a))
+    return PROJECTM_ERROR;
+
+  std::string lowerName(name);
+  std::transform(lowerName.begin(), lowerName.end(), lowerName.begin(), tolower);
+
+  ParamRGBA * param_rgba = new ParamRGBA(name, r, g, b, a);
+  CValue iv, ub, lb;
```
revmischa: these aren't used

src/libprojectM/MilkdropPresetFactory/InitCondUtils.hpp
```diff
@@ -48,14 +59,37 @@ inline void LoadUnspecInitCond::operator() (Param * param) {
...
+	    param_name = param_name.substr(0,param_name.length()-1);
+	    suffix = "2";
+    }
+    if (ends_with(param_name, "_r") || ends_with(param_name,"_g") || ends_with(param_name,"_b"))
+    {
+        std::string param_rgb = param_name.substr(0,param_name.length()-2) + "_rgb" + suffix;
+        std::string param_rgba = param_name.substr(0,param_name.length()-2) + "_rgba" + suffix;
+        if (m_initCondTree.find(param_rgba) != m_initCondTree.end() || m_initCondTree.find(param_rgb) != m_initCondTree.end() ||
+            m_perFrameInitEqnTree.find(param_rgba) != m_perFrameInitEqnTree.end() || m_perFrameInitEqnTree.find(param_rgb) != m_perFrameInitEqnTree.end())
+            return;
```
revmischa: maybe warn?

src/libprojectM/MilkdropPresetFactory/InitCondUtils.hpp
```diff
@@ -48,14 +59,37 @@ inline void LoadUnspecInitCond::operator() (Param * param) {
 	if (m_perFrameInitEqnTree.find(param->name) != m_perFrameInitEqnTree.end())
 		return;
 
+	// _rgba and _rgb hide corresponding _r, _g _b, _a parameters
+	std::string param_name = param->name;
+	std::string suffix;
+	if (ends_with(param_name, "2"))
```
revmischa: comment?

mbellew:
BTW, I don't understand why Travis CI failed...

mbellew:
I just realized, I didn't handle the matrix version of this which is handled directly in PerPointEqn.cpp and PerPixelEqn.cpp.  Hmm, encapsulation is good...

mbellew:
I'm going to postpone this for a short bit.  I think it makes sense to do some Param/PerPointEqn cleanup separately (and first), instead of doing with the rgb feature as one pull request.

PR #142, opened 2019-01-04
Per point bug

mbellew:
Still tracking down broken presets.  Found similar presets where one work and one didn't and tracked down to the per_point parameters "value1" and "value2".  They were not being passed properly.

To see the fix in action try "shifter + geiss - neon pulse (glow mix).milk"


struktured:
I thought we put that for a specific reason- are you sure it doesn't break another preset?

mbellew:
The previous code _never_ wrote to the values1[] and values2[], so that couldn't be right.  The presets I debugged were just reading uninitialized data.

Issue #118, opened 2018-10-24
Complex presets with problems


revmischa:
http://mischa.lol.s3.amazonaws.com/projectM/presets-broken-martin.zip

These Martin presets have serious problems in projectM, and supposedly push milkdrop and are cool. Maybe can help us find some issues with parsing and shaders.

psperl donated his Parser.cpp which he says adds some assertions that help track down actual parsing problems. We should merge it in as well. http://mischa.lol.s3.amazonaws.com/projectM/Parser.cpp

mbellew:
5 of these presets use value1 and value2  (per point variables).  Pull 142 may have addressed these.

https://github.com/projectM-visualizer/projectm/pull/142

mbellew:
Three of the presets are broken because they refer to a function that doesn't exist: gmegabuf()

martin - carpet weaver.milk
martin - massif central 2.milk
martin - sphery tales.milk

and since 'martin - tiling the tube.milk' seems to work, that just leaves 'martin - dark galaxy.milk' as the last mystery.

revmischa:
From some forum post:
> The way to transfer them upwards to the "per frame" section is via the 100 global "reg" variables (reg00 to reg99), or via the unlimited global array gmegabuf[].

revmischa:
Also see #186 for more problematic presets

korfuri:
I've found another possibly bad preset: `Rovastar & Krash - Interwoven (Contra Mix).milk`

I don't know if the preset itself is bad, but it's causing a whole bunch of undefined behaviors in the parser and in the execution. Here's a valgrind trace of me running that preset: https://pastebin.com/GUixatJm

korfuri:
Also happening in:
* `Rovastar & Krash - Interwoven (Contra Mix).milk` (https://pastebin.com/GUixatJm)
* `Mstress & Darius - Pursuing The Sunset.milk` (https://pastebin.com/fCESn1Ab)

I'll edit this comment if I find more, but this seems to be all.

PR #143, opened 2019-01-07
Perf cleanup

mbellew:
refactor to consolidate all reading/writing Param in various places (InitCond, Eqn's, Expr, Parser) and put it all back into Param

This is just less confusing, and also makes it cleaner to create new param types, such as "rgb" and "rgba".


revmischa:
All this is great but kind of scary. What do you think about adding some unit tests to make refactoring less scary?
Furthermore if you want to help fix issues with the parser, I do suggest merging the fixed up Parser.cpp from the closed-source app fork that was donated before doing any more changes to Parser - see https://github.com/projectM-visualizer/projectm/issues/118

mbellew:
I really like the idea of unit tests (hence the new presets/tests directory).  I wasn't sure what's up with the projectM-test directory since it doesn't seem to be built currently.

revmischa:
I have no idea what's in that directory

src/libprojectM/MilkdropPresetFactory/InitCond.cpp
```diff
@@ -58,59 +54,23 @@ void InitCond::evaluate()
...
-  if ( param->type == P_TYPE_INT) {
- 	assert(param->engine_val);
-	 *((int*)param->engine_val) = init_val.int_val;
-     return;
-  }
-
-  /* Parameter is of a float type, just like C */
+void InitCond::evaluate(bool evalUser)
+{
+    assert(this);
```
revmischa: what does this do?

src/libprojectM/MilkdropPresetFactory/Param.cpp
```diff
@@ -191,3 +190,228 @@ Param * Param::new_param_string(const char * name, short int flags, void * engin
 }
 
 
+/*
+ * I made Param extend Expr just to avoid inheritence of multiple virtual classes,
+ * however, this is really the subclass that knows how to interact with Expr
+ */
+
+struct _Param : public Param
```
revmischa: all of this code needs comments

src/libprojectM/MilkdropPresetFactory/InitCond.cpp
```diff
@@ -58,59 +54,23 @@ void InitCond::evaluate()
 }
 
 /* Evaluate an initial conditon */
-void InitCond::evaluate(bool evalUser) {
-
-
-
-   assert(this);
```
revmischa: https://stackoverflow.com/questions/40783056/safely-check-if-this-is-null

Issue #147, opened 2019-01-11
projectM::renderFrame() synchronization


mbellew:
I think I tracked down another ghost.  If I leave projectM running it will crash after an hour (or four).  I believe this is due to bad synchronization in projectM::renderFrame().  Here's my reasoning:

It makes the following call to wake up the background thread to call projectM::evaluateSecondPreset()

`    pthread_cond_signal(&condition);`

At which point the background thread presumably will wake up and start running evaluateSecondPreset().  Later the foreground thread waits for the background thread to finish its work by making this call

`    pthread_mutex_lock( &mutex );`

That seems correct, because while the background thread is running, it holds the lock so the foreground thread will be forced to wait to acquire this lock.

I think there are two bugs here, let me know if I've got this right before I embarrass myself with a bad PR.

1) The foreground thread is assuming that the background thread will wake up in response to the pthread_cond_signal() _before_ the foreground thread can finish its work and get to its pthread_mutex_lock().  This seems to be true 999 times out of 1000.  But what if the background thread is really sleepy, and does not wake up for a long time.  It could still be asleep (even though it is eligible to wake up), and not have taken hold of the mutex yet.  Then the foreground thread will blow thorough its lock/unlock.  Then the background thread finally wakes up... and toes are gonna get stepped on.

2) My undestanding is that it's not best practice to assume the world is in the state you expect when a thread awakes from a pthread_cond_wait().  So it's probably a good idea to put use explicit state variable to communicate between the two threads that there is work to be done (or work is finished).

What finally tipped me of was seeing a stack trace with the background thread running evaluateSecondPreset(), and the main thread off doing something completely unrelated instead of being in renderFrame() like you'd expect.

Thread 9 
#4  0x000055ed3913b12c in projectM::evaluateSecondPreset (this=this@entry=0x55ed3aa66cb0) at projectM.cpp:321
#5  0x000055ed3913b1e2 in projectM::thread_func (this=0x55ed3aa66cb0, vptr_args=<optimized out>) at projectM.cpp:304
#6  0x000055ed3913b23c in thread_callback (prjm=<optimized out>) at projectM.cpp:287
#7  0x00007f75ce07f6db in start_thread (arg=0x7f759cba0700) at pthread_create.c:463
#8  0x00007f75cdda888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 1 
[...]

PR #149, opened 2019-01-12
147

mbellew:
Fix for https://github.com/projectM-visualizer/projectm/issues/147
Seems to resolve the crashes.

I just noticed that this included a lot of whitespace changes.  Sorry about that.  The main change is in the header file anyway, plus update to renderFrame() and evaluateSecondPreset()

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -57,9 +57,10 @@
 #ifdef USE_THREADS
 #include "pthread.h"
 
-pthread_mutex_t mutex;
-pthread_cond_t  condition;
+#include "BackgroundWorker.h"
+
 pthread_t thread;
+BackgroundWorkerSync worker_sync;
 
 #ifdef SYNC_PRESET_SWITCHES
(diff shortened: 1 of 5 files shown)


src/libprojectM/projectM.cpp
```diff
@@ -70,29 +71,22 @@ projectM::~projectM()
 {
     void *status;
 
-    #ifdef USE_THREADS
+#ifdef USE_THREADS
     std::cout << "[projectM] thread ";
     printf("c");
```
revmischa: whats with the printfs

revmischa: Remove imo

revmischa: I mean are they printed? Is this code run?

mbellew: There is uncommented debug code in QPulseAudioThread.cpp and TextureManager.cpp as well.  Might be nice to have a runtime DEBUG flag as well as a compile time flag.  When I have a few minutes I could try to do a pass.

PR #151, opened 2019-01-13
Perf cleanup

mbellew:
Here's the PR again.  Same as before, but added a few tests and ProgramExpr.

This tees up the fun work, which would be to JIT the per_pixel and per_point expressions.  I think it's pretty easy (well not easy, but not rocket science either).


src/libprojectM/MilkdropPresetFactory/MilkdropPreset.cpp
```diff
@@ -387,94 +394,55 @@ void MilkdropPreset::evaluateFrame()
...
-
-
-      for (x=0;x<presetInputs().gx;x++){
-	for(y=0;y<presetInputs().gy;y++){
-	  _presetOutputs.warp_mesh[x][y]=presetOutputs().warp;
-	}}
+#ifdef __SSE2__
+inline void init_mesh(float **mesh, const float value, const int gx, const int gy)
+{
+  __m128 mvalue = _mm_set_ps1(value);
```
revmischa: 1. shouldn't the compiler do this stuff anyway?
2. what about memset?

src/libprojectM/MilkdropPresetFactory/MilkdropPreset.cpp
```diff
@@ -387,94 +394,55 @@ void MilkdropPreset::evaluateFrame()
...
-  for (std::map<int, PerPixelEqn*>::iterator pos = per_pixel_eqn_tree.begin();
-       pos != per_pixel_eqn_tree.end(); ++pos)
-    pos->second->evaluate(mesh_x, mesh_y);
-
+    if (nullptr == per_pixel_program)
+    {
+        std::vector<Expr *> steps;
+        for (std::map<int, PerPixelEqn*>::iterator pos = per_pixel_eqn_tree.begin(); pos != per_pixel_eqn_tree.end(); ++pos)
+            steps.push_back(pos->second->assign_expr);
+        per_pixel_program = new ProgramExpr(steps,false);
```
revmischa: comment?

revmischa:
tests ought to be run by travis

src/libprojectM/MilkdropPresetFactory/MilkdropPreset.cpp
mbellew: Right as usual...

src/libprojectM/MilkdropPresetFactory/MilkdropPreset.cpp
mbellew: On one machine it seemed to make a big difference.  Though you'd think the compiler would do this. I haven't dug that deep into in gcc optimizations.  BTW, I've found gperftools really helpful.  Though I try to be careful to use builds with -DNDEBUG to make sure I'm profiling something meaningful.  Is there an easy way to do that using ./configure?

revmischa: I was profiling with xcode and Instruments. clang will emit different optimizations. Are you compiling with -O2/-O3/-Ofast?
and since this is a contiguous block of memory can we not just memset the region?

mbellew: I thought about using memset, but I don't think it will escalate to using 128bit writes (the compilers/libraries are very clever though).  I've just been compiling with the defaults, for profiling I just add -DNDEBUG to remove asserts().

mbellew:
Running the junit tests in Travis would be good.  I assume that means creating a simple top level app that it can simply run?

PR #153, opened 2019-01-16
projectM-unittest

mbellew:
revive projectM-test directory
* new executable src/projectM-test/projectM-unittest
* still need to update existing code SDL->SDL2

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -69,11 +69,10 @@ pthread_mutex_t preset_mutex;
 
 projectM::~projectM()
 {
-    void *status;
-
 #ifdef USE_THREADS
+    void *status;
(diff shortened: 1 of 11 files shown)


src/libprojectM/projectM.cpp
```diff
@@ -69,11 +69,10 @@ pthread_mutex_t preset_mutex;
 
 projectM::~projectM()
 {
-    void *status;
-
 #ifdef USE_THREADS
+    void *status;
     std::cout << "[projectM] thread ";
-    printf("c");
+    printf("cl");
```
revmischa: maybe just delete these

mbellew: Will try to get to as a part of finding debug printing code in general.

PR #159, opened 2019-01-26
JIT compilation of  equations using LLVM

mbellew:


revmischa:
llvm-config doesn't come with the dev libraries, only from package `llvm`

PR #164, opened 2019-02-10
Add a nix file

sage-michaels:
This makes it easy to install all the relevant dependencies
on OSX. You can then build the project with:
```bash
nix-shell nix/shell.nix --run \
  './autogen.sh && ./configure.sh --enable-sdl && make'
```


revmischa:
Cool... I'll give it a try
why not add a travis CI step for it?

PR #166, opened 2019-02-22
Use proper includedir for libprojectM

luigino:


luigino:
After a short inquiry I can confirm the rule is:
`#include <libprojectM/libprojectM.h>`, therefore I'll retract this PR.

PR #167, opened 2019-03-02
Testing using AX_HAVE_QT for configuring Qt compilation

revmischa:
Using autoconf extension for checking for Qt5 instead of pkg-config
Please test and tell me if it works
It makes travis unhappy for unknown reasons - travis thinks Qt isn't installed ???

Apparently everyone's Qt5 setup is totally different


mrturcot:
Build is successful

PR #170, opened 2019-03-03
Allocate PerlinNoise dynamically / Fix thread detach after join

peak3d:
1.) There are some devices around (AFTV 2017 for example) which have limited stack size.
PerlinNoise is huge in size and leads to stack overflow -> crash if used in kodi.

2.) if pthread_join is called the thread is in some implementations in a not joinable mode and cannot be detached. Because there is no need to detach a thread after join, the detach was removed.

### DIFF
--- a/src/libprojectM/Renderer/TextureManager.cpp
+++ b/src/libprojectM/Renderer/TextureManager.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <vector>
+#include <memory>
 
 #include "projectM-opengl.h"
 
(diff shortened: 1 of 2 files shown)


src/libprojectM/Renderer/TextureManager.cpp
```diff
@@ -84,55 +84,57 @@ TextureManager::TextureManager(const std::string _presetsURL, const int texsizeX
...
     GLuint noise_texture_hq_vol;
     glGenTextures( 1, &noise_texture_hq_vol );
     glBindTexture( GL_TEXTURE_3D, noise_texture_hq_vol );
-    glTexImage3D(GL_TEXTURE_3D, 0, GL_RGBA, 32, 32, 32, 0, GL_RGB, GL_FLOAT, noise.noise_hq_vol);
+    glTexImage3D(GL_TEXTURE_3D, 0, GL_RGBA, 32, 32, 32, 0, GL_RGB, GL_FLOAT, noise->noise_hq_vol);
     Texture * textureNoise_hq_vol = new Texture("noisevol_hq", noise_texture_hq_vol, GL_TEXTURE_3D, 32, 32, false);
     textureNoise_hq_vol->getSampler(GL_REPEAT, GL_LINEAR);
     textures["noisevol_hq"] = textureNoise_hq_vol;
+
+    delete noise;
```
revmischa: Isn’t there an auto ptr thing that will delete automatically when going out of scope?

peak3d: If projectM is stdc++11 we can use https://de.cppreference.com/w/cpp/memory/unique_ptr

PR #171, opened 2019-03-03
Add GLES into automake build / fix android compile string

peak3d:
This PR changes adds GLES detection in configure step.


peak3d:
yeah, sure, but tbh. I have no experience with autoconf and don't see me as a good sparring partner for such kind of discussions.
We can leave this PR simply open without merging for ppl. who are common with the projectm build system. If they have a solution for GLES, pls. let me know, I will test this then in our (kodi) build environment.

Issue #109, opened 2018-10-01
Preset subdirectories


revmischa:
It would be really useful for organizing presets if the presets directory could contain sub-directories.

mancoast:
Do we need to find an alternative for dirent.h that uses recursive directory scan for presets path?

revmischa:
Perhaps?

mancoast:
This is a great task for someone new that wants to get involved and requires exact specifics for what needs done

AlwinEsch:
Have added for Kodi a Patch to change prefix install way, see second commit here: https://github.com/xbmc/visualization.projectm/pull/36

But I'm not sure how this can be used on other systems.

Also, I noticed that some of the same file names were in different folders and then overwritten during installation to a previous folder.

revmischa:
Agreed, so you would like a config option to enable recursively scanning the preset directory for all preset files? And deduplicating by filename?

revmischa:
Done in #385

PR #176, opened 2019-03-11
Sort input file list

bmwiedemann:
so that projectM builds in a reproducible way
in spite of indeterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

Without this patch, multiple .milk files varied between builds
in interesting ways:
```diff
--- "old//usr/share/projectM/presets/Geiss - All-Spark Polar.milk"
+++ "new//usr/share/projectM/presets/Geiss - All-Spark Polar.milk"
@@ -3,7 +3,7 @@
 PSVERSION_WARP=2^M
 PSVERSION_COMP=2^M
 [preset00]^M
-fRating=4.000000^M
+fRating=2.0^M
```

Maybe it is also a bug and either duplicate input .milk files should 

* be dropped or
* be synchronized or
* all of them should be installed into subdirs so no overwrites happen


revmischa:
Agreed, there's a semi-related ticket: #109

PR #181, opened 2019-03-21
Msvc

mancoast:
this enables build using visual studio 2017


mancoast:
there are still many things to update and fix, however, i would like to get these build files into the centralized code base and establish continuous integration for windows using appveyor to prevent future bitrot for the win32 build

labkey-matthewb:
This is great.  Does Travis support building under Windows?

mancoast:
Travis does not support windows, the repo owner will need to activate appveyor. I will make another PR for win32 CI after these build files exist in repo to reference in the CI config yaml

src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -11,6 +11,11 @@
 #define FAKE_AUDIO          0
 #define TEST_ALL_PRESETS    0
 
+#ifdef WIN32
+// cannot get SDL audio with win32
```
revmischa: really? why not? doesn't this make the windows SDL build pointless?

PR #187, opened 2019-03-30
Temporarily move bad presets to avoid disrupting users' trippy visuals

rothn:
This PR would move bad presets to a `bad_presets` folder in the repository root so that they don't disrupt the visuals seen by users while the community works to fix them. This also raises awareness of the bad presets and makes them easily accessible to maintainers.


rothn:
- The two CI runs with LLVM ran fine.
- The CI run with GCC failed because ```The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install g++-7 libsdl2-dev libglm-dev libc++-dev" failed and exited with 100 during .[0m```
- The CI run with XCode failed because the build tried to update XCode Command-Line Tools and ran out of time

PR #190, opened 2019-04-08
XDG Base Directory Specification compliance

agireud:
Currently the configuration file for projectM is created under a user's home directory specified by the environment variable `$HOME`.

I searched the repository for any discussions on the matter and failed to find anything so I went ahead and drafted an initial fix given recent trends to avoid cluttering a user's home directory.

This pull request would change the location of the configuration file to comply with the XDG base directory specification and create the configuration file in a directory specified by `$XDG_CONFIG_HOME` instead.

See: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
+++ b/src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
@@ -163,6 +163,8 @@ std::string read_config()
 	FILE *out;
 
 	char* home;
+        char* xdg_home;
+        
 	char projectM_config[1024];
 	char default_config[1024];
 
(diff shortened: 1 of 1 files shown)


revmischa:
Might mess up people who use the old path and their config stops working? No idea if anyone has config files in HOME at present

src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
```diff
@@ -173,30 +173,30 @@ std::string read_config()
 	strcpy ( projectM_config+strlen ( PROJECTM_PREFIX ), CONFIG_FILE );
 	projectM_config[strlen ( PROJECTM_PREFIX ) +strlen ( CONFIG_FILE ) ]='\0';
 	printf ( "dir:%s \n",projectM_config );
-	home=getenv ( "HOME" );
+	home=getenv ( "XDG_CONFIG_HOME" );
```
revmischa: is this guaranteed to be set? about on say, macOS?

agireud: As of macOS version 10.14.4, `$XDG_CONFIG_HOME` is not set.

We can do a check to see if the variable is set, and in the case that it isn't use a sane default. (Maybe continue to use `$HOME` and leave the current behavior as is.)

agireud:
Went ahead and gave it another go.  This time around, if the legacy configuration file is found, it will default to using it.

In the case that `~/.projectM/config.inp` is not found, if the environment variable `XDG_CONFIG_HOME` is set, it will attempt to create the config directory and file there.

If it is not set, it will attempt to create the configuration file at `~/.projectM/config.inp`.

src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
```diff
@@ -158,50 +158,58 @@ int main ( int argc, char*argv[] )
...
-		strcpy ( projectM_home, home );
-		strcpy ( projectM_home+strlen ( home ), "/.projectM/config.inp" );
-		projectM_home[strlen ( home ) +strlen ( "/.projectM/config.inp" ) ]='\0';
-
-		if ( ( out = fopen ( projectM_home,"w" ) ) !=0 )
-		{
-
-			if ( ( in = fopen ( projectM_config, "r" ) ) != 0 )
+	else {
+                if (xdg_home) {
```
revmischa: it looks like this still ignores the old config file if XDG home is set even if the old config file exists?
or am I misreading?
sort of hard to follow because of all the branches and lack of commenting...

agireud: Yeah, the whole function could use some cleanup to be honest. Given my C and C++ are a bit rough I didn't want to move things around too much.

The old config file is checked for and read here at line 181.

I considered reworking the entire function to make things much easier to read but I held off because of various concerns. My initial thought had been to start using things like std::string instead of C char arrays and fixing the branching and adding comments as well.

The main thing that stopped me however was concerns over portability of using things like std::string and changing the fopen calls to something like fstream which provides other things like type safety. As I've been out of the C/C++ game for a while now, I wasn't sure about the implications of making these sort of sweeping changes.

If updating the function as a whole and using more modern libraries isn't a big concern, I can go that route and just rewrite the whole thing to clean it up to make function much more readable. I just didn't want to clobber the whole function without at least having a conversation about it.

revmischa: we're using some modern C++, C++14 I think? using fstream etc is totally fine. rewriting things is totally fine. can merge this now if you want, or can wait for you

agireud: Why don't we go ahead and get this merged if you're satisfied with it.

That will give me some time to brush up on my C++ and follow up with a second pull request to tackle the rewrite in one go.  It should also serve as a good reference for me during refactoring.

PR #191, opened 2019-04-11
projectM Android app

PeterMalkin:
I think I got the basics all hooked up. libprojectM is compiled using NDK, and then jni wrapper links it properly in, rendering into the native activity's glSurfaceView. 

Please take a look. Its definitely not the cleanest. But i'd rather merge it now, since there has been a few interested people here who wanted to see an android port. May be some of them will improve things.


revmischa:
Okay but please remove the duplicated presets first. This PR adds 2,000 files

PeterMalkin:
Sure. I only store the test presets in case someone will want to skip the native compilation step. 

The presets are actually copied over by `make install`, so its easy to get to the same state.

teotigraphix:
I am building Android right now, things work in a Nexus 7. :)

Thanks you so much for this work!

revmischa:
@psperl do you have any android bits you'd like to contribute?

PR #194, opened 2019-04-13
Use real input on Win/SDL

dzarda:
- Use real input in SDL app on Win (no FAKE_AUDIO)
- Update docs for Win
- Move all building-related info to BUILDING.md


mancoast:
The SDL project is a good example of the graphics, but the SDL audio component is not well supported by different hardware. That is why I have the fake_audio set by default with windows preprocessors. It might work on your mobo with your specific audio driver, but it crashes on my windows systems. I suggest we leave fake_audio for windows continuous integration testing.

dzarda:
@mancoast I feel ya, but I don't fully agree:
- If you're a developer, you don't use the CI build and will compile with whatever macro required
- If you're a Windows user that just wants to give it a try, you want to download and test drive the thing on real data, that's what it's for, nevermind it crashing on some platforms

revmischa:
Yes, we should publish a pre-release and let people download it and try it out and report issues. If it fails, does it output any useful debug information people can share with us?

dzarda:
I expect some SDL-related failure with little value for us

revmischa:
want to do a pre-release for windows?

mancoast:
Sure! I am also working to finalize a new client for windows that dynamically links to libprojectm

PR #173, opened 2019-03-05
Reduce Qt5 requirements to a minimum

luigino:
Fixes https://github.com/projectM-visualizer/projectm/issues/172

PR #197, opened 2019-04-25
 4 commits 

sergiomb2:
Let me now if you want that I do several pull requests.

Commit 1 : copied from https://github.com/projectM-visualizer/projectm/pull/173
Commit 2 : projectM-jack don't need hard coded -lSDL I use SDL2 so we don't need to link with SDL1 and we can build withou SDL-devel files
Commit 3: /usr/bin/projectM_jack => /usr/bin/projectM-jack
Commit 4: Remove unneeded files, they are regenerated with autotools


configure.ac
```diff
@@ -164,6 +164,11 @@ AX_CHECK_COMPILE_FLAG([-std=c++11], [
 AC_ARG_ENABLE([qt], AS_HELP_STRING([--enable-qt], [Enable Qt: needed for pulseaudio and jack GUIs]), [], [enable_qt=check])
 AS_IF([test "$enable_qt" != "no"],
   [
+    case $host_os in
+      linux*)
+        PATH="$PATH:`pkg-config --variable=host_bins Qt5Core`"
```
revmischa: comment this please... i have no idea what this is doing

src/projectM-jack/Makefile.am
```diff
@@ -22,14 +22,13 @@ ${QT_CXXFLAGS} \
 
 # TODO: build qprojectM-jack and projectM-jack as separate programs
 
-bin_PROGRAMS = projectM_jack
+bin_PROGRAMS = projectM-jack
```
revmischa: this is okay i just hope it doesn't trip up downstream maintainers

configure.ac
sergiomb2: hello , I'm studying this [1], while the work it is and hack ! 
I will try improve it today 
[1]
https://autotools.io/pkgconfig/pkg_check_modules.html

src/projectM-jack/Makefile.am
sergiomb2: I forgot to mention this also is align with desktop file configuration , which try run projectM-jack

sergiomb2:
I dropped Qt patch (Commit 1), it will be to review later in other PR

Commit 2 : projectM-jack don't need hard coded -lSDL I use SDL2 so we don't need to link with SDL1 and we can build withou SDL-devel files
Commit 3: /usr/bin/projectM_jack => /usr/bin/projectM-jack  also align with desktop file configuration , which try run projectM-jack
Commit 4: Remove unneeded files, they are regenerated with autotools

PR #201, opened 2019-04-28
Added VR/Multiple screen support

hibengler:
I broke the RenderFrame into three separate methods in order to allow pass2 to be rendered multiple times.  This is to satisfy the requirements of VR, but might have other uses as well.

Made Renderer also have separate methods, and made it possible to do pass2 in an offset on the screen (like for left eye and right eye).

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -304,154 +304,14 @@ void projectM::evaluateSecondPreset()
 
 void projectM::renderFrame()
 {
-#ifdef SYNC_PRESET_SWITCHES
-    pthread_mutex_lock(&preset_mutex);
-#endif
-
-#ifdef DEBUG
-    char fname[1024];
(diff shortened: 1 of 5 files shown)


src/libprojectM/projectM.cpp
```diff
@@ -454,6 +454,228 @@ void projectM::renderFrame()
...
+            // printf("Hard Cut\n");
+            if (settings().shuffleEnabled)
+                selectRandom(true);
+            else
+                selectNext(true);
+        }
+    }
+
+
+    if ( timeKeeper->IsSmoothing() && timeKeeper->SmoothRatio() <= 1.0 && !m_presetChooser->empty() )
```
revmischa: perhaps this code could be refactored? it looks copied and pasted

hibengler: It is done.

Issue #205, opened 2019-05-10
How can I render an mp3 to a video file?


NightMachinery:
I like to render an mp3 to a video file. I took a look at https://github.com/projectM-visualizer/projectm/issues/10, but I didn’t find a way to do that? Was support of projectM ever added to https://github.com/djfun/audio-visualizer-python? Though that project seems dead anyways.

hsantanna:
You mean, you want to export the visualization animation generated while playing a song (encoded as mp3) to a video file (encoded like a h264 mp4 or mkv).

Maybe try something like this:
http://blog.scene.ro/posts/capturing-opengl-video-on-linux/

Iiridayn:
Yeah, OBS can be set up to record the window + audio easily enough - but the setup runs at the rate of 1 second per second of output, and requires post-processing to trim the start/end.

Been digging into the code/API. Looks like it uses the batch approach. It might be simple to write an application which takes an audio stream in PCM s16le format over stdin, and writes a "video" stream of binary ppm images (p6 - trivial to implement w/o dependencies) over stdout from an OpenGL surface. This then could be used with a `ffmpeg` pipeline as `ffmpeg -i <audio file> -f s16le -acodec pcm_s16le | <new program name> | ffmpeg -i <audio file> -f ppm_pipe -i - -map 0:a:0 -map 1:v:0 out.mkv` or something (_not_ great at `ffmpeg` cli, the second command is a guess, but something w/`ppm_pipe` should work).

Great example of the Unix philosophy overall - do one thing well - just missing pipe support :).

kblaschke:
Closing this issue, as we currently have no resources to implement it. We'll create a page in the Wiki with all sorts of suggested, but unimplemented applications of libprojectM.

msuiche:
@Iiridayn This seems promising, anyway you managed to make this command work with [`projectMSDL`](https://github.com/projectM-visualizer/frontend-sdl2) ?

Iiridayn:
> @Iiridayn This seems promising, anyway you managed to make this command work with [`projectMSDL`](https://github.com/projectM-visualizer/frontend-sdl2) ?

Sorry to say it hasn't been a priority for me. As I mentioned, it shouldn't be terribly difficult - a CS Sophmore might be able to take it on as a semester project for example. (The code'd be bad compared to a senior, but hey, if it works...). If you build it, please let me know :).

PR #207, opened 2019-05-19
GStreamer (WIP)

revmischa:
### Work in progress
* wrapper application which lets you specify an input file/url to read from
* gstreamer element which renders the visual

## Help Wanted
If you are familiar with gstreamer, please help make this work!


coast77777777777:
what is gstreamer? does this work with appleTV? does this work with chromecast? Some leverage chromecast and appletv for music. How can we support those end users?

revmischa:
GStreamer is an amazing multimedia library that can do a lot with setting up routing of audio and video https://gstreamer.freedesktop.org/documentation/?gi-language=c

Personally I'd love to have this working so I can feed in an audio file and output a movie file (for #205). It would enable all kinds of uses for projectM, allowing it to be placed anywhere in any sort of media pipeline someone could dream up.

coast77777777777:
Ok this makes sense to me now. I have been recording the hdmi signal using hardware. I am curious the system requirements to record some 4K while generating a smooth 60fps.

kblaschke:
Might still make sense to add this as an external support library in the project, closing for now since it's stale & hard to rebase on current master.

PR #216, opened 2019-06-01
init Stereo

mancoast:
@revmischa are you OK with these headers?

### DIFF
--- a/src/projectM-sdl/pmSDL.cpp
+++ b/src/projectM-sdl/pmSDL.cpp
@@ -1,9 +1,32 @@
-//
-//  pmSDL.cpp
-//  SDLprojectM
-//
-//  Created by Mischa Spiegelmock on 2017-09-18.
-//
+/**
+* projectM -- Milkdrop-esque visualisation SDK
+* Copyright (C)2003-2019 projectM Team
+*
+* This library is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License as published by the Free Software Foundation; either
+* version 2.1 of the License, or (at your option) any later version.
+*
+* This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/projectM-sdl/pmSDL.cpp
```diff
@@ -144,7 +167,10 @@ void projectMSDL::keyHandler(SDL_Event *sdl_evt) {
         case SDLK_f:
             if (sdl_mod & KMOD_LGUI || sdl_mod & KMOD_RGUI || sdl_mod & KMOD_LCTRL) {
                 // command-f: fullscreen
-                toggleFullScreen();
+				// Stereo requires fullscreen
+#ifndef STEREOSCOPIC_SBS
```
revmischa: maybe this check would be better in `toggleFullScreen()` - just make it always toggle to true or no-op if stereo

src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -1,16 +1,35 @@
...
+* Authors: Created by Mischa Spiegelmock on 6/3/15.
+*
+* 
+* experimental Stereoscopic SBS driver functionality by
+*	RobertPancoast77@gmail.com
+*/
 
 #include "pmSDL.hpp"
 
-#define FAKE_AUDIO          0
```
revmischa: why remove? FAKE_AUDIO is at least useful for testing. is it documented anywhere else?

mancoast: I moved theses definitions to the header file so they can be effective across multiple files.

src/projectM-sdl/pmSDL.cpp
mancoast: Ok i will rework these

PR #239, opened 2019-09-18
There was a bug that existed in open GL ES 2.0 implementations and random textures

hibengler:
where glTexImage2D could not convert GL_RGB formed textures into GL_RGBA
while generating the noise textures.   The GL error
was GL_INVALID_OPERATION(0x502)
This probably made the noise textures blank.
So a new class was added that has the alpha channel for the values.

Also - added a hook so that the time of day can be manipulated in case milkdrop is being used for non-real time rendering.  This hook is not implemented on WIN32 variants.

### DIFF
--- a/src/libprojectM/Renderer/PerlinNoiseWithAlpha.hpp
+++ b/src/libprojectM/Renderer/PerlinNoiseWithAlpha.hpp
@@ -1,0 +1,176 @@
+/*
+ * PerlinNoise.hpp
+ *
+ *  Created and based on PerlinNoise.hpp
+ *  Created on: Sep 14, 2019
+ *      Author: hibengler
+ */
+
+
+/* The reason for this cousin class is that in Open GLES 2.0 environments,
+the glTexImage2d cannot convert from GL_RGB into GL_RGBA
+so the TextureManager has to use something that is pre-RGBA
+*/
+#ifndef PERLINNOISEWITHALPHA_HPP_
+#define PERLINNOISEWITHALPHA_HPP_
+
+#include <math.h>
+
+class PerlinNoiseWithAlpha
+{
+public:
+
+    float noise_lq[256][256][4];
+    float noise_lq_lite[32][32][4];
+    float noise_mq[256][256][4];
+    float noise_hq[256][256][4];
+    float noise_lq_vol[32][32][32][4];
+    float noise_hq_vol[32][32][32][4];
+
+
+	PerlinNoiseWithAlpha();
+	virtual ~PerlinNoiseWithAlpha();
+
+private:
+
+	static inline float noise( int x)
+	{
+	    x = (x<<13)^x;
+	    return (((x * (x * x * 15731 + 789221) + 1376312589) & 0x7fffffff) / 2147483648.0);
+	   }
+
+	static inline float noise(int x, int y)
+	{
+		 int n = x + y * 57;
+		 return noise(n);
+	}
+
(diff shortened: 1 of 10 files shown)


src/libprojectM/Renderer/PerlinNoiseWithAlpha.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+
+	PerlinNoiseWithAlpha();
+	virtual ~PerlinNoiseWithAlpha();
+
+private:
+
+	static inline float noise( int x)
+	{
+	    x = (x<<13)^x;
```
revmischa: i truly actually have no idea what is going on here. comment? make some constants?

revmischa:
you can merge master in to (hopefully) get tests to pass - i just made some fixes

src/libprojectM/Renderer/PerlinNoiseWithAlpha.hpp
hibengler: No, it is generating random noise data used by some of the visualizers. 

The noise functions convert the standard random numbers to follow different 
ranges, styles, etc.

inline float noise looks to be generating a psuedo random number that is seeded with X.


 This file is basically a copy of PerlInNoise.hpp, and does the same thing as PerlInNoise class, except that it generates the noise in an RGBA buffer instead of an RGB buffer.

OpenGL ES 2.0 cannot transform RGB into RGBA in hardware,   but it can do RGBA to RGBA.  This is why I me the step-sibling class.

revmischa:
The windows build still [appears to be failing](https://ci.appveyor.com/project/revmischa/projectm/builds/27535554)

```
       "C:\projects\projectm\msvc\projectM.sln" (default target) (1) ->
       "C:\projects\projectm\msvc\projectM.vcxproj.metaproj" (default target) (4) ->
       "C:\projects\projectm\msvc\projectM.vcxproj" (default target) (8) ->
       (ClCompile target) -> 
         C:\projects\projectm\msvc\../src\libprojectM\timer.cpp(30): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int [C:\projects\projectm\msvc\projectM.vcxproj]
         C:\projects\projectm\msvc\../src\libprojectM\timer.cpp(30): error C2146: syntax error: missing ';' before identifier 'pprojectm_gettimeofday' [C:\projects\projectm\msvc\projectM.vcxproj]
         C:\projects\projectm\msvc\../src\libprojectM\timer.cpp(66): error C2065: 'pprojectm_gettimeofday': undeclared identifier [C:\projects\projectm\msvc\projectM.vcxproj]
         C:\projects\projectm\msvc\../src\libprojectM\timer.cpp(70): error C3861: 'projectm_gettimeofday': identifier not found [C:\projects\projectm\msvc\projectM.vcxproj]
```

hibengler:
I think I got it. Unable to test on win32 so I reverted that section in timer.h (and cpp) to what it was.

revmischa:
Thanks for the effort! Appreciate it.

>         C:\projects\projectm\msvc\../src\projectM-sdl\projectM_SDL_main.cpp(112): error C3861: 'time': identifier not found [C:\projects\projectm\msvc\projectMSDL.vcxproj]

https://ci.appveyor.com/project/revmischa/projectm/builds/27573266

hibengler:
Fixed projectM_SDL_main.cpp for windows platforms.

PR #249, opened 2019-10-05
allow separate folder structure of presets during install

AlwinEsch:
This is intended to install the folders as deposited at the source. If the backend is used then the desired folder will be selected.

This allows a better overlay of the available presets and avoids overwriting the same filenames in different folders during installation.

New configure value "--enable-preset-subdirs"

Related to here https://github.com/xbmc/visualization.projectm/pull/36 and to allow without patch there.

Hope it is largely correct, my experience with "configure" is not quite as good.

Also, I have nothing better as a name, do you still have ideas for it?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,6 +32,7 @@ AS_IF([test "x$enable_emscripten" = "xyes" || test "x$EMSCRIPTEN" = "xyes"], [
   enable_threading=no
   enable_gles=yes
   enable_sdl=yes
+  enable_preset_subdirs=no
 ], [
   dnl Running in a normal OS (not emscripten)
   AX_CHECK_GL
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -129,6 +130,13 @@ AC_ARG_ENABLE([gles],
     AC_DEFINE([USE_GLES], [1], [Define USE_GLES])
 ])
 
+AC_ARG_ENABLE([separate_preset_install],
+  AS_HELP_STRING([--enable-separate-preset-install], [Keep the presets folder structure defined in the source during installation.]),
```
revmischa:
```suggestion
  AS_HELP_STRING([--enable-preset-subdirs], [Organize presets into subdirectories.]),
```
my suggestion

AlwinEsch: That sounds better, thanks. Is updated.

PR #255, opened 2019-10-22
added 52kpresets from sourceforge

ravermeister:
added the milkdrop preset pack collection as split zip of 20MB files


ravermeister:
is that fine?

revmischa:
Really appreciate you taking the effort to contribute this, I'm just really wary about adding hundreds of megabytes to this source repo forever. I propose to host them on AWS here: https://mischa.lol/projectM/52k-presets.zip
What do you think?
Again sorry to be so slow here and negate the effort you did. I just wonder if increasing the size of this repo tremendously is the best approach. I've been seeding the torrent of these presets for a long time to try to help get it spread around.

ravermeister:
They are already on source forge, and now as a zip too 🙂. I once tried to create a Drupal milkdrop site with custom addon's sothat your was thinking about a website where they are

coast77777777777:
New git repo for plain text Community milk drops would be ideal. GitHub.com/Projectm-visualizer/presets/ Git submodule . Also then the file names can be tidied up for everyone to enjoy so out apps can reach broader audiences on the stores. I want these presents 🎁 in the Xbox/iPhone store. Someone please take control of this library of creations and curate the the inappropriate stuff so we can maintain our store app rates as “E for everyone”.

revmischa:
Dumb question - do the textures need to go with the presets? Are they tied together?

ravermeister:
indeed some presets load the textures as background and do their stuff on top of them.

PR #256, opened 2019-10-23
Parser update

revmischa:
Related to #118 -
Incorporating a few changes (I assume these are fixes) to Parser.cpp from @psperl's fork.

I don't really know if any of these are fixing anything or helping.


psperl:
These changes fix some parsing errors with newer android NDKs. Maybe it
fixes other ARM platforms as well. These changes should have no effect on
platforms that are currently functioning well.

kblaschke:
From what I see, most of the issues addressed in this PR have already been merged/fixed in other changesets. The biggest part, supporting the missing functions like megabuf and loop, will be added later in the parser rewrite. The existing parser isn't up to the task as it has some limitations, e.g. it expects a variable on the left side of all equations. Internally, Milkdrop converts assignments like `x = 1;` to `_set(a, 1);`, so there's not even an assignment operator anymore.

Closing the PR as rebasing it would probably be an impossible task due to the amount of changes accumulated since then.

PR #257, opened 2019-10-24
projectMSDL user interface using ImGui.

r-hackett:
The goal of this is to provide a user interface for projectM-sdl but currently it's pretty much a demonstration to get some feedback on whether or not it might be useful to use [Dear ImGui](https://github.com/ocornut/imgui). What's implemented is a menu with some simple options (F3 toggles the menu) including the option to select an audio capture device (without WASAPI loopback and fake PCM only) as well as one to change the preset (double click the item in the list) from the playlist. To achieve this some changes were made to the code which selects an audio capture device, opting for an init function to enumerate the devices and functions to release and open a device.

ImGui supports a variety of graphics API's and provides an implementation for each; their [SDL](https://github.com/ocornut/imgui/tree/master/examples/example_sdl_opengl3) and [OpenGL 3](https://github.com/ocornut/imgui/blob/master/examples/imgui_impl_opengl3.cpp) implementations are used. The GL3 implementation as well as the platform-agnostic stuff is added to the libProjectM renderer and SDL to projectM-sdl. They advise to start out using their premade implementation which works out the box so I started from there, although I modified imgui_impl_opengl3.h/.cpp slightly to point IMGUI_IMPL_OPENGL_LOADER_CUSTOM to projectM-opengl.h and to define IMGUI_IMPL_OPENGL_ES3 if USE_GLES is defined, as well as commenting out some code which selects GL headers. Todo, need to change that to IMGUI_IMPL_OPENGL_ES2 for emscripten, will do.

A few points.

- ImGui can load and render TrueType fonts so could be a replacement for FTGL. I can provide a wrapper class in libProjectM for drawing text if you think it's a good idea.

- Of course there's more options and features that could be added and it could be made consistent with the Qt settings in the pulseaudio project (ImGui supports drop down menus, text input, etc).

- It builds for *nix and android but haven't been able to confirm the windows build yet until that issue is resolved (should be perfectly fine). 

- Hopefully the coding style is appropriate. I tried to keep it in line with the majority of libProjectM. The ImGui files might want renaming to fit projectM's style but for clarity I left them as they are for now.

- Uncomment line 338 in pmSDL.cpp to view ImGui's demo window. As I'm writing this I noticed I deleted the show_demo_window variable, oops, but easy fix. `bool show_demo_window = true;`

So I leave it up to you. I think ImGui ticks all the boxes, performs well, looks good, and is MIT licensed which is GPL compatible. With some work the menu might be really useful and could provide an OpenGL ES friendly way to render .ttf fonts.

Much love to this awesome project and big thanks to you devs. 

p.s. Weekdays are a bit busy so apologies for any delays. Also, first PR from me, hopefuly not too ambitious.

### DIFF
--- a/src/projectM-sdl/projectM_SDL_main.cpp
+++ b/src/projectM-sdl/projectM_SDL_main.cpp
@@ -297,7 +297,8 @@ srand((int)(time(NULL)));
 
 
 	SDL_GLContext glCtx = SDL_GL_CreateContext(win);
-
+    SDL_GL_MakeCurrent(win, glCtx);
+    SDL_GL_SetSwapInterval(1); // Enable vsync
 
     SDL_Log("GL_VERSION: %s", glGetString(GL_VERSION));
     SDL_Log("GL_SHADING_LANGUAGE_VERSION: %s", glGetString(GL_SHADING_LANGUAGE_VERSION));
@@ -355,9 +356,12 @@ srand((int)(time(NULL)));
 #endif
 
 #if !FAKE_AUDIO && !WASAPI_LOOPBACK
-    // get an audio input device
-    if (app->openAudioInput())
+    // Init audio capture.
+    if (app->initAudioInput())
 	    app->beginAudioCapture();
+    else {
+        // Exit gracefully?
+    }
 #endif
 
 #if TEST_ALL_PRESETS
--- a/.directory
+++ b/.directory
@@ -1,0 +1,6 @@
+[Dolphin]
+Timestamp=2019,10,24,21,18,44
+Version=4
+
+[Settings]
+HiddenFilesShown=true
--- a/src/libprojectM/Renderer/Makefile.am
+++ b/src/libprojectM/Renderer/Makefile.am
@@ -2,6 +2,11 @@
 noinst_LTLIBRARIES = libRenderer.la
 
 libRenderer_la_SOURCES = \
+  imgui/imgui_demo.cpp \
+  imgui/imgui_draw.cpp \
+  imgui/imgui_widgets.cpp \
+  imgui/imgui.cpp \
+  imgui/imgui_impl_opengl3.cpp \
   SOIL2/image_DXT.c \
   SOIL2/image_helper.c \
   SOIL2/SOIL2.c \
--- a/src/libprojectM/Renderer/Makefile.am
+++ b/src/libprojectM/Renderer/Makefile.am
@@ -7,6 +7,8 @@ libRenderer_la_SOURCES = \
   imgui/imgui_widgets.cpp \
   imgui/imgui.cpp \
   imgui/imgui_impl_opengl3.cpp \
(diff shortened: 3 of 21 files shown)


r-hackett:
Forgot that I placed imgui in vendor, need to remove that and the reference to it in `src/libprojectM/Renderer/Makefile.am`. Not sure why clang and gcc in the travis build can't find imgui.h when compiling the imgui source, not a great start.

r-hackett:
Thanks for the AppVeyor fix coast.

r-hackett:
[Remade](https://github.com/projectM-visualizer/projectm/pull/260). AppVeyor working, hurrah.

coast77777777777:
Hey Ryan,

This looks Fantastic! 
Feedback:

1. The GUI needs to be implemented in the projectm library, not the SDL implementation. 

2. Try to implement the GUI in a new CPP/HPP file and make a connection to the projectm renderer. A good place to launchpad is in Renderer.cpp 

3. The new GUI needs to hook keyboard events at the library level.  There is already machinery in place within projectm that needs updated/investigated. 

4. Don’t use someone overly complex for your GUI. Try to find a text based rendering library that supports OpenGL and opengles because the Xbox and phone are pushing limits of hardware in the current state without a GUI. 

5. The GUI is a fantastic contribution and I hope your implementation functions with the eyetune build for windows store on appveyor. Please try to build something text based. See if you can reimplement the exact Winamp GUI using a new text based library. (Download and install Winamp for the fully worked example of what the community expects with Milkdrop.) it was text based. Please consider a recreation of this. 

6. Appveyor needs an SDL build configuration that uses opengles so we can confirm that the new projectm GUI design supports both open gles and OpenGL. 

Thanks!

Rob

coast77777777777:
@r-hackett you have already met most of the constraints for project-m GUI that were listed from my email in the comment above. Please post screen shots.  Can you push another commit to rebuild the continuous integration?

coast77777777777:
Looks fantastic! Screenshots please

revmischa:
"Immediate Mode Graphical User Interface" - correct me if I'm wrong but OpenGL ES doesn't support Immediate Mode? Only vertex buffer objects
That means this would never work on mobile, raspberry pis, nvidia shield, or other devices?

src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -297,7 +297,8 @@ srand((int)(time(NULL)));
 
 
 	SDL_GLContext glCtx = SDL_GL_CreateContext(win);
-
+    SDL_GL_MakeCurrent(win, glCtx);
```
revmischa: what do these two lines change or do? why do we want vsync enabled?

revmischa:
I think a GUI would be really wonderful, especially if it was done in openGL and in the core. The issue here is that immediate mode is super deprecated and doesn't work anywhere except old deprecated desktop GL implementations as I understand it. We had a huge task getting rid of all immediate mode support in projectM so that it was possible to run on android, raspi, shield, etc..

PR #230, opened 2019-07-08
include the 52k preset pack from winamp forums

otherdev98:
This pull request includes the milkdrop text files from the mega preset pack, this PR omits all the textures.
also, some/many of these presets have inappropriate names, this should be discussed before merging into the code.

PR #259, opened 2019-10-25
52k preset link

revmischa:
Instead of #230/#255


coast77777777777:
@revmischa https://docs.microsoft.com/en-us/azure/cognitive-services/content-moderator/api-reference

I have been scanning the file names using the content moderation cognitive api services from azure. It is slow, but steady and offloads the cognition.  I am ready to add push many of these presets now.

revmischa:
Coolio

PR #262, opened 2019-10-30
Update keybindings similar to MilkDrop

srp3003:


coast77777777777:
Greetings @srp3003 

Can you review to verify that your implementation matches the classic milkdrop keybindings?

image

src/libprojectM/KeyHandler.cpp
```diff
@@ -169,7 +169,7 @@ void projectM::default_key_handler( projectMEvent event, projectMKeycode keycode
 	        break;
 	    case PROJECTM_K_b:
 	      break;
-            case PROJECTM_K_n:
+            case PROJECTM_K_H:
```
revmischa: can we keep both?

revmischa:
opened in favor of #279

Issue #271, opened 2019-11-11
Compile fails with "DSO missing from command line"


scruloose:
It's entirely possible that this is some configuration error on my part; I don't do a lot of installing from source. My desired outcome is to have a working install of projectM-pulseaudio in particular.

My system is Debian Bullseye, and the version of projectm in the Debian repos crashes as soon as it tries to load any visualization, as described in this Debian [bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=924925). I have successfully built and installed the Sid version using `apt -b source projectm-pulseaudio`, but even in Sid it's an old 2.1.0 version, and it seems the patch has not been incorporated there, as I still get the same behaviour (runs as long as the 'ProjectM' placeholder visualization is active, instantly crashes on switching to any other).

I have downloaded and extracted the [3.1.1 RC6](https://github.com/projectM-visualizer/projectm/releases) release, installed the dependencies listed in [BUILDING.md](https://github.com/projectM-visualizer/projectm/blob/master/BUILDING.md), including the four packages listed for `--enable-qt`.

I did not do the `autoreconf` step, as this is a release tarball, not a git clone.

`./configure --enable-qt` seems to complete successfully, concluding with the following output:

    Applications:
    =====
    
    libprojectM:            yes
    Threading:              yes
    SDL:                    no
    Qt:                     yes
    Pulseaudio:             yes
    Jack:                   yes
    OpenGLES:               no
    Emscripten:             no
    llvm:                   no
    Preset subdirs:         no

`make` produces a bunch of output, concluding with this:

    libprojectM_qt.a -ljack -lpthread ../libprojectM/.libs/libprojectM.a -lGL -ldl -pthread
    /usr/bin/ld: ../projectM-qt/libprojectM_qt.a(qprojectm_mainwindow.o): undefined reference to symbol '_ZN5QIconC1Ev@@Qt_5'
    /usr/bin/ld: /usr/lib/x86_64-linux-gnu/libQt5Gui.so: error adding symbols: DSO missing from command line
    collect2: error: ld returned 1 exit status
    make[3]: *** [Makefile:491: projectM-jack] Error 1
    make[3]: Leaving directory '/home/chris/projectm/projectM-3.1.1-rc6/src/projectM-jack'
    make[2]: *** [Makefile:457: all-recursive] Error 1
    make[2]: Leaving directory '/home/chris/projectm/projectM-3.1.1-rc6/src'
[...]

PR #273, opened 2019-11-12
Restore QtWidgets from m4 script

revmischa:
Multiple people are reporting issues with building Qt (https://github.com/projectM-visualizer/projectm/issues/271 , also on IRC)

We believe this change is the culprit: https://github.com/projectM-visualizer/projectm/commit/33ca67c8aee841c0ddb5cc1b868628eef0eeed79#diff-6cd33f1420898619d28f3dd2efb5535eL78


revmischa:
Just fixed in master because this commit is based on old history

PR #276, opened 2019-11-13
fix for freeze (issue 243)

mbellew:
don't assume ProjectM is allocated only once in application, reset BackgroundWorker in projectM_init()


mbellew:
PS I have a separate change coming to make the SDL test app work a bit more like the kodi app which seems more realistic for library consumers (pointer to projectM instead of inherit), plus a key shortcut to test delete/new.

PR #279, opened 2019-11-16
Update keybindings similar to MilkDrop

revmischa:
#262 but keeping the existing key bindings
From @srp3003


revmischa:
Key bindings should really be documented somewhere...

PR #290, opened 2019-12-12
use updates synchronized with vertical retrace if adaptive vsync fails

jas-per:
got projectM-SDL compiled on a raspberry pi 4 using the SDL kmsdrm-backend instead of X - works but with lots of flicker after a short while. seems that you need vsync, but adaptive vsync is not supported (yet)

the SDL documentation for `SDL_GL_SetSwapInterval` suggests this
> If application requests adaptive vsync and the system does not support it, this function will fail and return -1. In such a case, you should probably retry the call with 1 for the interval. 

works nicely on the rpi4 now!


revmischa:
Oh fantastic, I need to give this a try.

Issue #283, opened 2019-11-30
3.1.1-rc7 Compile fatal error: QGLWidget: No such file or directory


carsayao:
I'm new to building projects, but I'm having the same issue [#280](https://github.com/projectM-visualizer/projectm/issues/280#issue-526049987) compiling projectm-pulseaudio. I'm following `BUILDING.md` on Ubuntu 19.04. I ran: 

    ./configure --enable-pulseaudio --enable-qt
    make

Then:

`fatal error: QGLWidget: No such file or directory`

[comp-err.txt](https://github.com/projectM-visualizer/projectm/files/3906747/comp-err.txt)

I might try some older releases. Thanks!

revmischa:
I would love to know why this build passes on Travis on ubuntu https://travis-ci.org/projectM-visualizer/projectm/jobs/624274159

nuess0r:
same on debian unstable.

QGLWidget headers are provided by the packet "libqt5opengl5-dev" which is installed on my system.

They end up in "/usr/include/x86_64-linux-gnu/qt5/QtOpenGL/"

The file QGLWidget contains only one line, to include qgl.h which is in the same folder.

carsayao:
@revmischa I'm not sure what I did differently, but this build is now stable for me. I know I experimented with different package dependencies, and followed the same instructions I posted above on the same system. Thanks.

Issue #280, opened 2019-11-20
Compile still fails on Ubuntu 19.10 (QGLWidget)


Russ-Holmes:
Sorry for late response to rc7 fix.
Just attempted to compile release 3.1.1-rc7 and still fails on the widget module.

[projectm-comp-err.txt](https://github.com/projectM-visualizer/projectm/files/3870961/projectm-comp-err.txt)

revmischa:
Sorry, why close the issue?

Russ-Holmes:
I suspect the problem might be me not passing the correct library name when running the configure script (as for the Debian). I will try this evening and report back.

carsayao:
I'm new to building projects, but I'm having the same issue. I'm following `BUILDING.md` on Ubuntu 19.04. I ran:

    ./configure --enable-qt
    make

[comp-err.txt](https://github.com/projectM-visualizer/projectm/files/3900453/comp-err.txt)

Issue #274, opened 2019-11-12
Does not compile on Ubuntu 19.10


Russ-Holmes:
Compile fails with:

`/usr/bin/ld: ../projectM-qt/libprojectM_qt.a(qprojectmwidget_moc.o):(.data.rel.ro+0x0): undefined reference to `QGLWidget::staticMetaObject'
collect2: error: ld returned 1 exit status
`
On both RC4 and RC6
Installed QT version 5.12.4+dfsg-4ubuntu1

Any suggestions?

revmischa:
#273 - can you test this fix please?

revmischa:
This should be fixed on master now, please let us know if you have any further issues., Thanks!

mamantoha:
Hi @revmischa . Unfortunately, it's not fixed on master. Please reopen this issue.

revmischa:
Same issue as #280 #283

mamantoha:
I was able to make it work on Ubuntu 19.10

```
./configure --enable-qt --enable-gles LIBS="-lQt5Gui -lQt5OpenGL"
make
sudo make install
```

<img width="812" alt="Screenshot_20191218_135644" src="https://user-images.githubusercontent.com/61285/71084754-a41af180-219e-11ea-88c7-8fc228b30179.png">

Hope this helps someone.

Button4Button:
That's impressive... is that the latest RC or stable release?  I've been
trying on Linux Mint 19.3 (beta) and failing consistently.

mamantoha:
@Button4Button from the master branch

Russ-Holmes:
Yes, I was able to compile 3.1.1-rc7 on ubuntu 19.10 with:

`sudo apt install libtool pkg-config libglm-dev qt5-default qtdeclarative5-dev libgles2-mesa-dev
./configure --enable-qt --enable-gles LIBS="-lQt5Gui -lQt5OpenGL"
make
sudo make install`

revmischa:
I am on ubuntu 19.04 and I can simply run
```
./configure --enable-qt
make
```
And it builds without errors. On master (3.1.1-rc7)

revmischa:
I just upgraded to ubuntu 19.10 and got:
```
depbase=`echo ConfigFile.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
[...]

PR #293, opened 2019-12-29
Hardcode missing Qt5 libs (libQt5Gui and libQt5OpenGL) 

revmischa:
To fix missing DSO linkage with Qt5 on Ubuntu 19.10

Should fix the error: 
```
/usr/bin/ld: ../projectM-qt/libprojectM_qt.a(qprojectm_mainwindow.o): undefined reference to symbol '_ZN5QIconC1Ev@@Qt_5'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libQt5Gui.so: error adding symbols: DSO missing from command line
```

I don't know if there is a better way to solve this but hardcoding these needed libraries seems to fix the issue, at least for me, on ubuntu 19.10.

Should address #274 #283 #271 #280

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,10 @@ AS_IF([test "$enable_qt" != "no"],
         # we have at least qt5 if $have_qt is true
         enable_qt=yes
         export QT_SELECT=qt5
+
+        # we depend on lQt5Gui and lQt5OpenGL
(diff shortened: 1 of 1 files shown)


revmischa:
Going to merge this and release it and see if it helps. Let me know if this helped!

PR #297, opened 2020-01-03
Display FPS

milkdropper:
Display FPS

### DIFF
--- a/src/libprojectM/Renderer/Renderer.cpp
+++ b/src/libprojectM/Renderer/Renderer.cpp
@@ -238,9 +238,9 @@ void Renderer::SetupPass1(const Pipeline& pipeline, const PipelineContext& pipel
 	totalframes++;
 
 	/*
-	If FPS is displayed (by pressing F5 or config):
+	If FPS is displayed (by pressing F5 or by config):
 		- check if 250 milliseconds has passed (1/4 of a second)
-		- multiply the total rendered frames (totalframes) by four to get the approximate frames per second cound.
+		- multiply the total rendered frames (totalframes) by four to get the approximate frames per second count.
 		- reset the totalframes and timer (lastTime) so we don't trigger for another 250 milliseconds.
 	*/
 	if (this->showfps)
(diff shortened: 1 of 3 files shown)


src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -229,6 +236,16 @@ void Renderer::ResetTextures()
 void Renderer::SetupPass1(const Pipeline& pipeline, const PipelineContext& pipelineContext)
 {
 	totalframes++;
+	this->currentTime = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
```
revmischa: perhaps we could avoid doing this work if we're not outputting FPS?

src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -229,6 +236,16 @@ void Renderer::ResetTextures()
...
+	milliseconds ms = std::chrono::duration_cast<std::chrono::milliseconds>(this->currentTime - this->lastTime);
+	double diff = ms.count();
+	if (diff >= 250)
+	{
```
revmischa: comment explaining this block would be nice

milkdropper: There is already a total frame count (totalframes). The idea is to track from the initial run until we reach 250 milliseconds (1/4 of a second). To estimate the FPS I multiply the totalframes rendered in 1/4 of a second by 4. That way we get an update every 1/4 of a second.

milkdropper: I will put in a comment block to explain the logic.

src/libprojectM/Renderer/Renderer.cpp
milkdropper: You are right. If FPS is not being displayed there is no need to check this timer. I will make that change.

revmischa:
Also, isn't there a FPS counter in the SDL app? Wonder if there's a way to clean that up and have it use this code

milkdropper:
> Also, isn't there a FPS counter in the SDL app? Wonder if there's a way to clean that up and have it use this code

Can you point me in the direction of this? I looked into projectM-sdl and only found a placeholder to display FPS. The lack of something working is why I submitted this.

Here is what projectM-sdl\Release\projectM.exe looks like after pressing F5 now.
image

revmischa:
There sort of is in SDL but it's really more for timing the frame delays - https://github.com/projectM-visualizer/projectm/blob/master/src/projectM-sdl/projectM_SDL_main.cpp#L454

PR #303, opened 2020-01-18
Menu and FPS changes

milkdropper:
Hopefully the commit messages are clear, but the summary is:
- more accurate FPS (resetting timers when the menu is first loaded and doing more exact math).
- removing the F1 help menu no longer removes the preset name from the screen.
- the F1 help menu is moved down and aligned with the preset name or FPS. Before you could have the help menu and preset name up at the same time and they go in the way of each other.
- the idle preset name is now set when you display it (before it displayed "None").
- changed the default Windows SDK version so all 10.* soany minor version of 10 should compile right away.


revmischa:
Looks good, thanks coast

PR #307, opened 2020-01-20
Emscripten

revmischa:
Very very rough sketch of running projectM in a browser. 
It works in Chrome, Safari doesn't seem to like WebGL2. Safari can kiss my ass.

image


kblaschke:
I'll just close this PR as the changes are either already merged or outdated and we now have a working demo app, including audio recording (with a good amount of lag, but probably not much we can do about it).

PR #313, opened 2020-02-01
Fix off by 1 qvar errors, var names

struktured:
This PR fixes an off by 1 bug where the variable `q0` is assumed to exist instead of `q32`. It also fixes possible memory bound violations when iterating over the `q` variables.


revmischa:
Thanks

psperl:
FYI, this isn't just a memory bound violation.  This fixes almost any complicated shader preset, because before this patch, all the q variables are off by 1 when entering the shader code.  So all shader program using a q variable (LOTS OF THEM) were broken before this.

PR #315, opened 2020-02-02
Beatdetect 309

mbellew:
I think I got getPCMScale() wrong.  Does using 1.5 instead of 4.0 seem better?  Feel free to experiment the scale is arbitrary.


mbellew:
Note: moved implementation of getPCMScale() into the cpp file to avoid recompiling the work just to change the constant.

PR #324, opened 2020-02-09
Remove CPPFLAGS from moc command lines. moc isn't a C++ compiler and some C++ flags caus it to fail (ex. -isystem).

yurivict:


revmischa:
Cool! I did the same https://github.com/projectM-visualizer/projectm/pull/323

Issue #330, opened 2020-03-12
keyboard shortcut F


VulkanLogic:
Readme does not mention keyboard shortcut
F: fullscreen

revmischa:
Feel free to correct it!

kblaschke:
It's now documented in the Wiki:

https://github.com/projectM-visualizer/projectm/wiki/User-Manual-(Standalone)#application

PR #333, opened 2020-03-13
Update README.md

VulkanLogic:
#330


revmischa:
It's command/ctrl-F, right? at least on projectMSDL it is

kblaschke:
> It's command/ctrl-F, right? at least on projectMSDL it is

Yup. Same for preset lock toggle, which is also documented without a modifier.

kblaschke:
Keyboard shortcuts were removed from the README, and are now available in the wiki.

PR #337, opened 2020-04-13
Automatically update version string in projectM.hpp

s09bQ5:
There is a copy of that header for the Android JNI, but the macros are not used in that case.


s09bQ5:
Oh, forget this PR. Didn't notice that you want to support building with Microsoft tools.

PR #343, opened 2020-05-01
SDL Monitor Fixes

milkdropper:
**Cmd/Ctrl-S: stretch monitors**
This should make stretching much more flexible. Before it was basic and would only work for monitors perfectly aligned vertically or horizontally.

**Cmd/Ctrl-M: change monitor**
There was a bug where it would try to go to a monitor that doesn't exist (diff between displayCount which starts at 1 and displayIndex which starts at 0)

### DIFF
--- a/src/projectM-sdl/pmSDL.cpp
+++ b/src/projectM-sdl/pmSDL.cpp
@@ -187,29 +187,28 @@ void projectMSDL::stretchMonitors()
 			SDL_GetDisplayBounds(i, &displayBounds.back());
 		}
 
-        int mostX = 0;
+		int mostXLeft = 0;
+		int mostXRight = 0;
 		int mostWide = 0;
-		int mostY = 0;
+		int mostYUp = 0;
+		int mostYDown = 0;
 		int mostHigh = 0;
 
 		for (int i = 0; i < displayCount; i++)
 		{
-			if (displayBounds[i].x < mostX) mostX = displayBounds[i].x;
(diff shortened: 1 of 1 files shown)


src/projectM-sdl/pmSDL.cpp
```diff
@@ -187,55 +187,29 @@ void projectMSDL::stretchMonitors()
...
-				else if (vertical) // perfectly aligned vertical monitors.
-				{
-					spanHeight = spanHeight + displayBounds[i].h;
-					spanWidth = widest;
-				}
-			}
-			SDL_SetWindowPosition(win, spanX, spanY);
-			SDL_SetWindowSize(win, spanWidth, spanHeight);
+			if (displayBounds[i].y < mostYUp) mostYUp = displayBounds[i].y;
+			if ((displayBounds[i].y + displayBounds[i].h) > mostYDown) mostYDown = displayBounds[i].y + + displayBounds[i].h;
```
revmischa, line 205: typo here? `+ +`?

src/projectM-sdl/pmSDL.cpp
```diff
@@ -187,55 +187,29 @@ void projectMSDL::stretchMonitors()
...
 		}
+
+        mostWide = abs(mostXLeft) + abs(mostXRight);
```
revmischa, line 208: weird indentation here

src/projectM-sdl/pmSDL.cpp
milkdropper: Yes - a typo. I'll fix.

PR #344, opened 2020-05-02
SDL: FPS cap based on display.

milkdropper:
**SDL Change:**
Before - max FPS was capped at 60 FPS.
Now - when there is no config (built-in defaults), the app will use the monitor max refresh rate as FPS cap.
Result - on my 144 hz monitor I am now getting 144 fps for many presets and it looks beautiful!

### DIFF
--- a/src/projectM-sdl/projectM_SDL_main.cpp
+++ b/src/projectM-sdl/projectM_SDL_main.cpp
@@ -339,7 +339,7 @@ srand((int)(time(NULL)));
 				if (current.refresh_rate > maxRefreshRate) maxRefreshRate = current.refresh_rate;
 			}
 		}
-		if (maxRefreshRate <= 0) maxRefreshRate = 60;
+		if (maxRefreshRate <= 60) maxRefreshRate = 60;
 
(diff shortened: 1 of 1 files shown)


src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -328,13 +328,26 @@ srand((int)(time(NULL)));
...
+		int maxRefreshRate = 0;
+		SDL_DisplayMode current;
+		for (i = 0; i < SDL_GetNumVideoDisplays(); ++i)
+		{
+			if (SDL_GetCurrentDisplayMode(i, &current) == 0)
+			{
+				if (current.refresh_rate > maxRefreshRate) maxRefreshRate = current.refresh_rate;
+			}
+		}
+		if (maxRefreshRate <= 0) maxRefreshRate = 60;
```
revmischa, line 342:
```suggestion
		if (maxRefreshRate < 60) maxRefreshRate = 60;
```

milkdropper: Agreed and changed in latest commit.

PR #345, opened 2020-05-02
SDL and libprojectM updates on config handling.

milkdropper:
**SDL Fix 1 of 2:**
_Problem:_
SDL only looks at /usr/local/share/projectM/ for config.inp.
If this fails it falls back on the built-in config, which users can't easily edit.

_Solution:_
SDL will now:
1. Look in /usr/local/share/projectM/
2. Look in %APPDATA%\projectM\ on Windows and ~/.projectM/ on other OS's.
3. Look in the current working directory.
If all of the above fails it falls back on the built-in config.

**libProjectM Fix**
_Problem:_
Once I got SDL easily loading a config file, I realized:
- Some configs are best set by SDL because it has awareness of your monitor settings.
- A user might only want to tweak 1 setting. By doing this, libProjectM will fall back on it's internal defaults which are really bad.

_Solution:_
libProjectM can be initialized two ways:
1. With a config file
2. With predefined settings.

The solution / fix in this commit is to give libProjectM a third option to initialize with:
3. With a config file AND predefined settings. 

With this new way to load libProjectM you can provide it with dynamic settings like height/width but let users modify only something like FPS or meshX/meshY.

This should not break libProjectM for any projects as the first two options are still in-tact.

**SDL Fix 2 of 2:**
SDL now supplies both a config file and built-in settings. If no config file was found during the load libProjectM will ignore the empty config file and just use the built-in settings.

### DIFF
--- a/src/projectM-sdl/projectM_SDL_main.cpp
+++ b/src/projectM-sdl/projectM_SDL_main.cpp
@@ -1,104 +1,131 @@
 /**
-* projectM -- Milkdrop-esque visualisation SDK
-* Copyright (C)2003-2019 projectM Team
-*
-* This library is free software; you can redistribute it and/or
-* modify it under the terms of the GNU Lesser General Public
-* License as published by the Free Software Foundation; either
-* version 2.1 of the License, or (at your option) any later version.
-*
-* This library is distributed in the hope that it will be useful,
-* but WITHOUT ANY WARRANTY; without even the implied warranty of
-* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-* Lesser General Public License for more details.
-*
-* You should have received a copy of the GNU Lesser General Public
-* License along with this library; if not, write to the Free Software
-* Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
-* See 'LICENSE.txt' included within this release
-*
-* projectM-sdl
-* This is an implementation of projectM using libSDL2
-* 
-* main.cpp
-* Authors: Created by Mischa Spiegelmock on 6/3/15.
-*
-* 
-*	RobertPancoast77@gmail.com :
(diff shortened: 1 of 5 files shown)


src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -1,104 +1,131 @@
...
+	/*if (type != GL_DEBUG_TYPE_OTHER)*/
 	{
-        std::cerr << " -- \n" << "Type: " <<
-           type << "; Source: " <<
-           source <<"; ID: " << id << "; Severity: " <<
-           severity << "\n" << message << "\n";
-       }
- }
+		std::cerr << " -- \n"
+							<< "Type: " << type << "; Source: " << source << "; ID: " << id
```
revmischa, line 44: what happened here?

revmischa:
The whitespace cleanup is appreciated but would be more readable if it was a separate PR
Travis builds are failing: https://travis-ci.org/github/projectM-visualizer/projectm/jobs/682307418#L1662

```
projectM.cpp:140:14: error: variable has incomplete type 'struct stat'
        struct stat sb;
```

src/projectM-sdl/projectM_SDL_main.cpp
milkdropper: It looks like Visual Studio tried to be "helpful" and changed the formatting. I'm not sure how to easily undo that unless I just submit a new PR and manually / copy paste the changes to the relevant sections.

milkdropper:
> The whitespace cleanup is appreciated but would be more readable if it was a separate PR
> Travis builds are failing: https://travis-ci.org/github/projectM-visualizer/projectm/jobs/682307418#L1662
> 
> ```
> projectM.cpp:140:14: error: variable has incomplete type 'struct stat'
>         struct stat sb;
> ```

I'm not great with github so I'll try submitting it all over again. I think the whitespace problem will be confusing.

PR #284, opened 2019-12-02
manage home dir for conf file

ycollet:
The projectM-sdl binary only look in install system directory for config.inp.
Let's look for this file in ~/.projectM/config.inp like for others projectM version.

PR #346, opened 2020-05-02
Sdl config

milkdropper:
**SDL Fix 1 of 2:**
_Problem:_
SDL only looks at /usr/local/share/projectM/ for config.inp.
If this fails it falls back on the built-in config, which users can't easily edit.

_Solution:_
SDL will now:
1. Look in /usr/local/share/projectM/
2. Look in %APPDATA%\projectM\ on Windows and ~/.projectM/ on other OS's.
3. Look in the current working directory.
If all of the above fails it falls back on the built-in config.

**libProjectM Fix**
_Problem:_
Once I got SDL easily loading a config file, I realized:
- Some configs are best set by SDL because it has awareness of your monitor settings.
- A user might only want to tweak 1 setting. By doing this, libProjectM will fall back on it's internal defaults which are really bad.

_Solution:_
libProjectM can be initialized two ways:
1. With a config file
2. With predefined settings.

The solution / fix in this commit is to give libProjectM a third option to initialize with:
3. With a config file AND predefined settings. 

With this new way to load libProjectM you can provide it with dynamic settings like height/width but let users modify only something like FPS or meshX/meshY.

This should not break libProjectM for any projects as the first two options are still in-tact.

**SDL Fix 2 of 2:**
SDL now supplies both a config file and built-in settings. If no config file was found during the load libProjectM will ignore the empty config file and just use the built-in settings.

### DIFF
--- a/src/projectM-sdl/projectM_SDL_main.cpp
+++ b/src/projectM-sdl/projectM_SDL_main.cpp
@@ -58,54 +58,50 @@ std::string getConfigFilePath(std::string datadir_path)
 	struct stat sb;
 	const char *path = datadir_path.c_str();
 
-	SDL_LogInfo(
-		SDL_LOG_CATEGORY_APPLICATION, "Looking for configuration file in data directory: %s\n", path);
+	// first check if settings file is in DATA_DIR
+	std::string configFilePath = path;
+	std::string configFullPath = configFilePath;
+	configFullPath += "config.inp";
 
-	std::string configFilePath = "";
-
-	// check if datadir exists.
-	// it should exist if this application was installed. if not, try the settings directory.
-	if (stat(path, &sb) != 0)
+	if (stat(configFullPath.c_str(), &sb) == 0)
 	{
-#ifdef WIN32
-		configFilePath += getenv("APPDATA");
-		configFilePath += "\\projectM\\config.inp";
-#else
-		configFilePath += getenv("home");
-		configFilePath += "/.projectM/config.inp";
-#endif
-		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION,
-			"Looking for configuration file in settings directory: %s\n", configFilePath.c_str());
-	}
-	else
-	{
-		configFilePath = path;
-		configFilePath += "config.inp";
+		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "Config: %s was found!\n", configFullPath.c_str());
+		return configFilePath;
 	}
 
-	// check if settings config file exists.
-	// if not, assume we're developing it and use the current working directory.
-	if (stat(configFilePath.c_str(), &sb) != 0)
-	{
-		configFilePath = "";
+	// now check settings directories (windows / other)
+	configFilePath = "";
 #ifdef WIN32
-		configFilePath += ".\\config.inp";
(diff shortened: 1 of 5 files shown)


src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -53,28 +53,55 @@ void DebugLog(GLenum source,
...
+	if (stat(configFullPath.c_str(), &sb) == 0)
+	{
+		SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "Config: %s was found!\n", configFullPath.c_str());
+		return configFilePath;
+	}
+
+	// now check settings directories (windows / other)
+	configFilePath = "";
+#ifdef WIN32
+	configFilePath += getenv("APPDATA");
```
revmischa, line 75: what does this return?

src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -315,53 +342,58 @@ srand((int)(time(NULL)));
...
-        // get path to our app, use CWD or resource dir for presets/fonts/etc
-        settings.presetURL = base_path + "presets";
+	}
+	if (maxRefreshRate <= 60) maxRefreshRate = 60;
+
+    float heightWidthRatio = (float)height / (float)width;
+    projectM::Settings settings;
+    settings.windowWidth = width;
+    settings.windowHeight = height;
+    settings.meshX = 128;
```
revmischa, line 371: can we define these defaults in one place? seemed like they were specified above too

milkdropper: > can we define these defaults in one place? seemed like they were specified above too

It will only be defined once in SDL. Those settings will be used as the built-in fall back ORalso sent with your config file. That way if your config file is missing anything (most likely on purpose because you commented out things like Width / Height) then it will fall back on the built-in SDL settings.

Defaults are also defined in libProjectM and I kept those the same.

src/projectM-sdl/projectM_SDL_main.cpp
milkdropper: Previously it would return the full path to the config file (if it exists), or an empty string if it does not exist.

Now it returns the **folder** of the config file (if it exists), or an empty string if does not exist. 

I changed it so the folder where the config file was found can easily be used as the new base_path for built-in config settings like the presets folders.

revmischa:
Some similar work was done in https://github.com/projectM-visualizer/projectm/pull/284/files

milkdropper:
Thanks for pointing that out. If you prefer to use that please do.

But the more important update (in my opinion) is allowing the library to accept a config file AND settings (so you can use a partial config file).

Would it make sense for that to be a separate PR?

revmischa:
Can you please document in the README what config paths are loaded? This really should be documented

kblaschke:
These changes still look useful, despite the SDL app being turned into a "dev UI". IMO we should not invest much time into improving the config file path logic in the test UI, as it's only for devs working on the library who don't want to use the external end-user SDL app for some reason.

Would you mind updating the commits with the new project structure, so we can merge it?

kblaschke:
Probably solved with the new SDL UI. If there's still a motion to merge this change, please rebase & reopen the PR as you see fit.

PR #347, opened 2020-05-03
Display status

milkdropper:
Display stats (Keybind: F4). I grabbed some immediately useful stuff that was exposed to the renderer. I think it is useful to learn / see how beat sensitivity is working and will allow people to confirm the audio device is being "heard" by projectM.

See screenshot below.

image

### DIFF
--- a/src/libprojectM/Renderer/Renderer.hpp
+++ b/src/libprojectM/Renderer/Renderer.hpp
@@ -179,8 +179,10 @@ private:
   void draw_title();
   void draw_title_to_screen(bool flip);
   void draw_title_to_texture();
-
-  int nearestPower2( int value );
+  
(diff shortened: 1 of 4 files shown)


src/libprojectM/Renderer/Renderer.hpp
```diff
@@ -179,8 +179,10 @@ class Renderer
   void draw_title();
   void draw_title_to_screen(bool flip);
   void draw_title_to_texture();
-
-  int nearestPower2( int value );
+  
+  std::string float_stats(float stat);
+  
+int nearestPower2( int value );
```
revmischa, line 185:
```suggestion
  int nearestPower2( int value );
```

PR #348, opened 2020-05-03
Beat sensitivity implemented (?)

milkdropper:
By showing the current beat sensitivity with the new stats key (F4), I noticed that it had no effect if I used the Up and Down arrow to adjust it.

I looked into the BeatDetect source and saw it tracks this value but does nothing with it. I also noticed the beat sensitivity defaults are very inconsistent with hard coded settings have it at 10.0, SDL defaulting to 0.8, and KeyHandler.cpp putting a limit of 5 with Up and Down arrow keys, 

I came up a simple idea: make beat sensitivity a multiplier for the key areas in beat detect like bass, mid range, and treble.

By doing this I immediately noticed how amazing beat sensitivity is! My PR adjusts the following:

- Default is now 1.0.
- Anything below 1.0 will decrease sensitivity of beat detect.
- A sensitivity of 0 will make it seem like there is no beat / audio (what I expected originally).
- Anything from 1.0 to 5.0 increases the beat detect (as originally intended).
- 5.0 looks wild (seizure warning)! 

Note: a fun visualization to test is Flexi - infused with the spiral.milk

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -589,6 +589,7 @@ void projectM::projectM_resetengine()
     {
         beatDetect->reset();
     }
+    beatDetect->beat_sensitivity = _settings.beatSensitivity;
 
 }
 
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -215,12 +215,21 @@ void projectM::readConfig (const std::string & configFile )
     _settings.softCutRatingsEnabled =
             config.read<bool> ( "Soft Cut Ratings Enabled", false);
 
+    // Hard Cuts are preset transitions that occur when your music becomes louder. They only occur after a hard cut duration threshold has passed.
+    _settings.hardcutEnabled = config.read<bool> ( "Hard Cuts Enabled", false );
+    // Hard Cut duration is the number of seconds before you become eligible for a hard cut.
+    _settings.hardcutDuration = config.read<int> ( "Hard Cut Duration", 60 );
+    // Hard Cut sensitivity is the volume difference before a "hard cut" is triggered.
+    _settings.hardcutSensitivity = config.read<float> ( "Hard Cut Sensitivity", 1.0 );
+    
+    // Beat Sensitivity impacts how reactive your visualizations are to volume, bass, mid-range, and treble. 
+    // Preset authors have developed their visualizations with the default of 1.0.
+    _settings.beatSensitivity = beatDetect->beatSensitivity = config.read<float> ( "Beat Sensitivity", 1.0 );
+
+
     projectM_init ( _settings.meshX, _settings.meshY, _settings.fps,
-                    _settings.textureSize, _settings.windowWidth,_settings.windowHeight);
-
-    _settings.beatSensitivity = beatDetect->beat_sensitivity = config.read<float> ( "Hard Cut Sensitivity", 1.0 );
-
-
+                _settings.textureSize, _settings.windowWidth,_settings.windowHeight);
+    
     if ( config.read ( "Aspect Correction", true ) )
     {
         _settings.aspectCorrection = true;
@@ -256,13 +265,17 @@ void projectM::readSettings (const Settings & settings )
 
     _settings.easterEgg = settings.easterEgg;
 
+    _settings.hardcutEnabled = settings.hardcutEnabled;
+    _settings.hardcutDuration = settings.hardcutDuration;
+    _settings.hardcutSensitivity = settings.hardcutSensitivity;
+    
+    _settings.beatSensitivity = settings.beatSensitivity;
+    
     projectM_init ( _settings.meshX, _settings.meshY, _settings.fps,
                     _settings.textureSize, _settings.windowWidth,_settings.windowHeight);
+    
 
-
-    _settings.beatSensitivity = settings.beatSensitivity;
     _settings.aspectCorrection = settings.aspectCorrection;
-
 }
 
 #ifdef USE_THREADS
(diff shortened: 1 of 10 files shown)


milkdropper:
Here is a video of it in action (you can turn off volume, it's annoying). 

Keep an eye on the "Sensitivity" value going down from 1.0 to 0.0 then back up to 5.0.

https://streamable.com/e7qunw

src/libprojectM/projectM.cpp
```diff
@@ -218,7 +218,7 @@ void projectM::readConfig (const std::string & configFile )
     projectM_init ( _settings.meshX, _settings.meshY, _settings.fps,
                     _settings.textureSize, _settings.windowWidth,_settings.windowHeight);
 
-    _settings.beatSensitivity = beatDetect->beat_sensitivity = config.read<float> ( "Hard Cut Sensitivity", 10.0 );
+    _settings.beatSensitivity = beatDetect->beat_sensitivity = config.read<float> ( "Hard Cut Sensitivity", 1.0 );
```
revmischa, line 221: I feel like "Hard Cut Sensitivity" should be changed

milkdropper:
One final touch: I realized waveforms rely on a separate function to get volume data (getPCMScale) for producing visuals. This is now also impacted by beat sensitivity. A good example: Rovastar & Rocke - Answer.42 (Trippy S. Mix)

src/libprojectM/projectM.cpp
milkdropper: > I feel like "Hard Cut Sensitivity" should be changed

I would agree. I can rename it to Beat Sensitivity and update configs. 

It seems Milkdrop used to have Soft Cuts and Hard Cuts for transitions (I didn't know this).

- Soft cuts were more frequent and based on a duration (default 16 seconds) + some randomness (up to 10 seconds by default).
- Hard cuts were less frequent and based on a loudness event **(disabled by default).**

My guess is hard cuts aren't needed as they were disabled by default anyway.

image

milkdropper: I did more digging and it seems projectM supports soft cuts and hard cuts, and it's doing them both.

Right now the hard cut duration is fixed to 3 (HARD_CUT_DELAY) and timekeeper is checking if we passed 3 seconds with a check on the difference between current volume vs. the previous volume is greater than the Hard Cut Sensitivity (what we now call Beat Sensitivity).

`else if ((beatDetect->vol-beatDetect->vol_old>beatDetect->beat_sensitivity ) &&
                 timeKeeper->CanHardCut())`

I recommend:
- Keep Hard Cut Sensitivity and document what it is. I could also make it a number between 0 to 100 but on my system volume is normally only showing as a number very low. I would like feedback from others once they start testing the stats PR.
- Add a new variable for Hard Cut Duration (default 60 seconds).
- Add a new variable / setting for Beat Sensitivity.

revmischa: 👍 please document your findings!

labkey-matthewb:
By my reading of the code I think this is combining two ideas.
a) How sensitive the "hard-cut" logic is to a beat, so that it will tend to change on  a beat (which is kinda cool) rather than just running out of time. see projectM::renderFrameOnlyPass1()
b) How large the bass/mid/treb values are.  The values for these average 0.66 and go up/down as a ratio from there (why not 1? I don't know that's how the code was when I first looked at it).

labkey-matthewb:
BTW.  I've thought it would be cool if the beat sensitivity in renderFrameOnlyPass1() became progressively more sensitve after the CanHardCut() time passed, to increase the chance of a synchronized switch instead of running out the clock (PresetProgressA()>=1.0)

milkdropper:
> By my reading of the code I think this is combining two ideas.
> a) How sensitive the "hard-cut" logic is to a beat, so that it will tend to change on a beat (which is kinda cool) rather than just running out of time. see projectM::renderFrameOnlyPass1()
> b) How large the bass/mid/treb values are. The values for these average 0.66 and go up/down as a ratio from there (why not 1? I don't know that's how the code was when I first looked at it).

Yes this all makes sense to me now. I don't think I was the only one who thought beat sensitivity had to do with the "impact" of the beat. At least by doing this we have a new feature. I'll separate these features and document them.

> BTW. I've thought it would be cool if the beat sensitivity in renderFrameOnlyPass1() became progressively more sensitve after the CanHardCut() time passed, to increase the chance of a synchronized switch instead of running out the clock (PresetProgressA()>=1.0)

That's a good idea. It will make more sense after HARD_CUT_DELAY is made configurable.

labkey-matthewb:
I think making both configurable is interesting (as long as the preset writers target the default sensitivity).  Another thing to verify with your change is how the waveforms react with your slider.  Do you want them to scale as well?  A while ago someone noticed that the waveforms did not autoscale, depending on the sound source/configuration.  If you turned down the volume, the waveforms could become very flat.  I fixed that with BeatDetect.getPCMScale().  So you may want to test how/if the waveforms react to your new setting.  (see some of the presets in presets/tests)

milkdropper:
Thanks again for your feedback. I think internally I will rename the current beat_sensitivity to **hardcut_**beat_sensitivity and this new setting (around this PR) will be named **beatdetect_**sensitivity.

For waveforms I modified getPCMScale here https://github.com/projectM-visualizer/projectm/pull/348/commits/a11112abafdb1f998cba0158cf0225a19ec1da97 . I noticed not impacted by beat_sensitivity (future: beatdetect_sensitivity), and I think they should be.

milkdropper:
I committed more changes:

**Hard Cuts:**
Hard cuts are now more clearly documented in the config.inp and in the code comments. It's much more flexible as you can enable/disable it and set a duration (the default was 3 seconds before, and Milkdrop's default was 60).
```
Hard Cuts Enabled = false  # Hard Cuts are preset transitions that occur when your music becomes louder. They only occur after a hard cut duration threshold has passed. 
Hard Cut Duration = 60     # Number of seconds before you become eligible for a hard cut.
Hard Cut Sensitivity = 1.0 # Volume sensitivity before a hard cut is triggered.
```

**Beat Sensitivity:**
What I originally described in this PR is now separately tracked and better documented.
`Beat Sensitivity = 1.0 # Beat Sensitivity impacts how reactive your visualizations are to volume, bass, mid-range, and treble. Range: 0 - 5 (from "dead" to VERY reactive)`

revmischa:
This good to merge?

milkdropper:
> This good to merge?

Yes, I am happy with it.

milkdropper:
Can you merge this into master?

revmischa:
@milkdropper with pleasure!

Issue #350, opened 2020-05-05
Config not documented anywhere


goosewobbler:
I've been messing around with ProjectM (SDL on MacOS) for a while and finally found the config file, after searching google and github issues.  I was using ProjectMilkSyphon - which does exactly what I want - if I ignore the dead presets, the crashes on manually switching presets, the heavy CPU load, and the semi-functional options.  So I switched to projectM, ditching the convenience of Syphon in favour of more stability and better functionality.

Basically I am wanting to tweak the preset duration as the SDL default (10s I believe) is too hyperactive for my liking.  I increased the preset duration to 60 and smooth transition duration to 5, which results in the project M logo visual not being replaced in a reasonable time (> 2 days).  So I am guessing that this combination of settings is unreasonable.  Perhaps - or does it just need more time?  Hopefully you can see where I'm going with this.  Is there any documentation for configuring project M?

Although I'd like to get projectM working for my particular use case I'd like to help the project out too.  I'm not a C dev but as part of figuring out my problem I'd like to offer my help in documenting the configuration options - assuming people with more knowledge than me are willing to answer my questions in this issue.  If there is some documentation on config, I'm sure it could be made easier to find...

goosewobbler:
Oh also logging.  If they exist, where are logs located on the different platforms?

milkdropper:
Hi @goosewobbler 

My guess is that your config doesn't have the presents folder set, or it's not a full path.

If you put your config file in /usr/local/share/projectM and your presets config value is "presets" it will look relative to the config file, not relative to the projectM binary. So maybe your presets are in /Users/goose/Downloads/projectM/presets but SDL will look in /usr/local/share/projectM.

I am planning to make new config options and will create clear documentation at the same time.

revmischa:
Do take a look at #346

goosewobbler:
Thanks guys, I knew about `/usr/local/share/projectM`.  My presets config value is `presets` and I seem to have presets in `/usr/local/share/projectM/presets`, which should work?  Does it need a particular ownership or...?  this is why I was also asking about the log file haha, might help with debug.

Offer to help still stands, I'll try to aggregate some information about the config file in this thread at least.

revmischa:
Help is greatly appreciated! On macOS for logs try launching the app binary from terminal, run the whole thing from xcode, or maybe check in Console.app

goosewobbler:
[...]

PR #351, opened 2020-05-07
Implement toast messages

milkdropper:
For lack of a better word, this PR implements toast notifications. For now they are displayed centered in the screen with a hard coded timeout of 2 seconds (I can't imagine a reason for this being configurable).

For immediate use: lock/unlock in messages are displayed in libprojectM and audio toggle in SDL will produce a useful message.

Obligatory demo video as POC here: https://streamable.com/4mgxx8. 

With fast transitions the feedback of locking is really helpful. Also audio toggle from SDL is implemented showing the device name on the screen as per feedback here https://github.com/projectM-visualizer/projectm/issues/350.

image

### DIFF
--- a/src/libprojectM/Renderer/Renderer.cpp
+++ b/src/libprojectM/Renderer/Renderer.cpp
@@ -55,7 +55,7 @@ void Renderer::drawText(GLTtext* text, const char* string, GLfloat x, GLfloat y,
 	// - GLT_TOP (default)
 	// - GLT_CENTER
 	// - GLT_BOTTOM
-	gltDrawText2DAligned(text, x, y, scale, GLT_LEFT, GLT_TOP);
+	gltDrawText2DAligned(text, x, y, scale, horizontalAlignment, verticalAlignment);
 
 	// Finish drawing text
 	gltEndDraw();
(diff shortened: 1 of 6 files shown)


src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -76,10 +76,13 @@ Renderer::Renderer(int width, int height, int gx, int gy, BeatDetect* _beatDetec
 	title_fontURL(_titlefontURL), menu_fontURL(_menufontURL), presetURL(_presetURL)
 {
 	this->totalframes = 1;
-	this->lastTime = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
-	this->currentTime = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+	this->lastTimeFPS = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+	this->currentTimeFPS = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+	this->lastTimeToast = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+	this->currentTimeToast = duration_cast<milliseconds>(system_clock::now().time_since_epoch());
```
revmischa, line 82: Maybe refactor `duration_cast<milliseconds>(system_clock::now().time_since_epoch())` to a function

src/libprojectM/Renderer/Renderer.hpp
```diff
@@ -102,13 +108,28 @@ class Renderer
   std::string fps() const {
 		return m_fps;
   }
+
+  void setToastMessage(const std::string& theValue)
+  {
+    // Initialize counters
+	lastTimeToast= duration_cast<milliseconds>(system_clock::now().time_since_epoch());
```
revmischa, line 115: wild indentation here

milkdropper:
- In the last commit I tried to address the indentation. I am using Visual Studio on Windows. Any idea how I can make it adapt to the existing indentation? In my view there is really inconsistent indentation and I would prefer to make everything the same. I assume Visual Studio could automate fixing this.

- I refactored all the timing for FPS and toast messages into functions.

- I tried to address a problem with displaying text (more below).

Currently if your toast, preset name, or status is larger than the draw window it will just overflow. See below:
image

The current commit will cut off text and make it obvious it was and overflow by adding "...". Except when it's multi-line then it will cut it off with some buffer. See below:
image

I also left behind (but commented out) option #2: instead of cutting off the text it scales it down. See below:
image

milkdropper:
Not sure why the travis build is showing incomplete. It looks done/passed when you click details.

PR #352, opened 2020-05-07
SDL bugfix for settings smaller than your monitors resolution.

milkdropper:
SDL FIX:
If the loaded config file or your hard coded settings for height/width are smaller than your monitors resolution it will currently display something really ugly in windows like this:

image

This bug fix will recognize that your settings are smaller than the monitor, and center them onto your screen. In either case you can use CTRL/CMD+F to go to full screen (which is probably what you wanted in the first place!)

Fix example below:
image


revmischa:
What about changing the projectM config based on the window size instead of vice versa?

milkdropper:
> What about changing the projectM config based on the window size instead of vice versa?

For SDL we could just ignore the resolution config settings altogether and default full screen.

But tasks to make config management better like https://github.com/projectM-visualizer/projectm/pull/284/files or https://github.com/projectM-visualizer/projectm/pull/346 could be updated write back settings on last run. Resolution and beat sensitivity are the only two things configurable during run time in SDL that I can think of.

milkdropper:
I have an idea: with SDL we could render multiple projectM instances (small like the first image I posted above, with low fps and meshx) tiled across a full screen SDL window, each with a different preset. This could be a "live" preset browser / previewer. If we could use arrow keys for navigator / pagination and mouse clicks to enter a desired preset I think it would be really cool.

Could anyone give guidance on how to position projectM in the sdl window? I tried paying with glViewPort on the renderer and the shader engine and offset did change but it was not behaving as expected.

revmischa:
As far as I know the only limit to having multiple instances of projectM or projectMSDL is CPU power. pmSDL isn't really set up to have multiple instances running without each having their own window. Maybe you could pass in a window or GL context when creating one. What's your idea exactly?

milkdropper:
projectM doesn't seem too CPU intensive right now (on modern computers). So I was hoping if multiple instances were started with reduce FPS/meshx/size it might not impact performance too much.

I found that multiple SDL windows is straight forward, but then they are separate windows in the task bar. To explain my idea more fully here is a copy/paste job into mspaint:
image

The idea is you can use arrow keys or mouse buttons to enter and navigate the grid. If SDL could do something like that, adjustable based on your resolution or customizable based on your preference, it would be a really slick way to preview and navigate presets.

revmischa:
My DREAM is to have a web UI to browse, edit, preview, and share presets. That would be amazing. It is part of the motivation behind the efforts in #307 
What you have here would be really terrific for browsing

milkdropper:
Yes that does sound fantastic and I would be happy to contribute.

Have you considered sending anonymous telemetry and rating data to the same web site? Or even let users sign-in / sign up to synchronize playlists / ratings / the presets they created?

With all of this the end user could get "good" presets at a priority. There is also a neat concept here https://github.com/LydiaSevelt/GeneticM/blob/master/README where someone was creating mutated presets based on a directory of "good" presets. This means the number of presets could be unlimited.

revmischa:
Can we change the projectM config to match the window size instead?

milkdropper:
SDL will now write the existing config back to the same file when you close it. That means if you open at a small resolution but then go full screen it will save that resolution.

writeConfig should probably updated to also write a comment into the config.

revmischa:
Good to merge?

milkdropper:
Yes - it's only a small change and I don't see any issues.

PR #359, opened 2020-05-24
Implement touching / clicking (interaction with presets).

milkdropper:
This PR implements touching / clicking functionality in libProjectM to dynamically generate a new waveform into the renderer. It's been implemented into SDL. 

Obligatory demo video: https://streamable.com/oe76ql

Notes:
- When you click the screen a random waveform is generated (bias to circles). The colour is also random. 
- The waveform will stay on screen for 5 seconds (or until you click / touch again).
- The waveform will act differently depending on what preset is loaded.
- There is future consideration for touch pressure (I think mobile apps can register how hard you push.) I tracked this but did not implement it yet. If implemented we could make the waveform thicker, bigger, etc if you pushed harder.
- Right now projectM doesn't accept absolutely X and Y coordinates for new waveforms. It wants a scale between 0 and 1. So I converted the pixel coordinates in SDL into the scale of X/Y for the approximate location where you clicked.
- Issue: In SDL the function SDL_ShowCursor should show the mouse cursor but I could not see it in Windows when in Full Screen (Ctrl-F). The ideal solution is that when you are in fullscreen it should only show the mouse cursor when moving your mouse or clicking, and have a timeout to hide it afterwards.


src/projectM-sdl/pmSDL.cpp
```diff
@@ -426,13 +431,30 @@ void projectMSDL::pollEvent() {
...
             case SDL_QUIT:
                 done = true;
                 break;
         }
     }
 }
 
+void projectMSDL::touch(float x, float y, int pressure, int touchtype) {
+    if (y > 1)
+        y = 1.0;
```
revmischa, line 452: what's this check for? is it possible?
comment this method

milkdropper: It's not possible anymore. I wrote that when I made a mistake and was going math off the wrong axis and didn't understand what I was getting a positive value. I'll remove it now.

revmischa:
Merge when ready 😀

milkdropper:
Much better this way! https://streamable.com/f1y1hl

src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -633,6 +654,121 @@ bool Renderer::timeCheck(const milliseconds currentTime, const milliseconds last
...
+			return;
+		}
+	}
+
+	touchx = x;
+	touchy = y;
+	touchp = pressure;
+	// If we touched randomly, then assign type to a random number between 0 and 8
+	if (type == 0) {
+		type = (rand() % 9) + 1;
```
revmischa, line 675: remove magic number; use constant

src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -633,6 +654,121 @@ bool Renderer::timeCheck(const milliseconds currentTime, const milliseconds last
...
+	waveformList.push_back(wave);
+}
+
+// Move a waveform when dragging X, Y, and Pressure can change. We also extend the counters so it will stay on screen for as long as you click and drag.
+void Renderer::touchDrag(float x, float y, int pressure)
+{
+	// if we left clicked & held in the approximate position of a waveform, snap to it and adjust x / y to simulate dragging.
+	// For lines we don't worry about the x axis.
+	for (int i = 0; i < waveformList.size(); i++) {
+		if (waveformList[i].x > (x-0.05f) && waveformList[i].x < (x+0.05f) && ((waveformList[i].y > (y-0.05f) && waveformList[i].y < (y+0.05f)) || waveformList[i].mode == Line || waveformList[i].mode == DoubleLine || waveformList[i].mode == DerivativeLine ))
```
revmischa, line 744: can we refactor this condition to a function?

src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -763,8 +899,7 @@ void Renderer::CompositeOutput(const Pipeline& pipeline, const PipelineContext&
 
 	glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
-	for (std::vector<RenderItem*>::const_iterator pos = pipeline.compositeDrawables.begin(); pos
-	     != pipeline.compositeDrawables.end(); ++pos)
+	for (std::vector<RenderItem*>::const_iterator pos = pipeline.compositeDrawables.begin(); pos!= pipeline.compositeDrawables.end(); ++pos)
```
revmischa, line 902: `for (auto drawable : pipeline.compositeDrawables)`

psperl:
The Android and iOS branches have a similar feature but it's more designed for touch interfaces.  The waveforms only stay while your fingers are touching the screen.  If you touch with multiple fingers, it connects each touch point with line waveforms.  So 2 fingers make a line, 3 fingers would make a triangle, and so on. As soon as you take your hand away it stops.  

I'm only mentioning this because it seems like this version is based around non-touch interfaces.  I bet if you wrote this on a phone you'd have chosen different semantics. Something to think about.

milkdropper:
I gave some thought to touch screen interfaces, but only the possibility of factoring pressure. My idea was that pressure could impact the size and alpha of a waveform. But I have not done any mobile development so you will know better if this is realistic / feasible.

Until someone ports projectM to IOS (where your app seems to be dead in the water based on reviews), the focus is going to likely be on devices with a keyboard/mouse or a remote/controller.

But things can get really interesting if we start thinking about VR where controllers and buttons can define your interaction with the renderer. There is a really amazing game called Chroma Lab which is inspiring: https://store.steampowered.com/app/587470/Chroma_Lab The photos and videos don't show it but you can open up pallette-like menus around your controllers that are used to change how you interact with visualizations. I highly recommend the app/game if you enjoy projectM / milkdrop.

src/libprojectM/Renderer/Renderer.cpp
milkdropper: done

milkdropper: (saved lots of unnecessary code)

src/libprojectM/Renderer/Renderer.cpp
milkdropper: done - also saved lots of unnecessary & not well-documented code

src/libprojectM/Renderer/Renderer.cpp
milkdropper: done

Issue #365, opened 2020-06-03
Input device toggle does not work on OSX 10.14.6


felder:
Using release v3.13 on a 2017 Macbook Pro running OSX 10.14.6, the input device toggle does not seem to actually use different input devices.

Hitting CMD-i shows output that suggests that the device is being switched.  However, the visual effects continue to respond only to mic input.  I can verify this by toggling the input volume via the sound preferences for the mic and watching the resulting visual effects changes.  Doing a similar thing for the currently selected input device, when it is not the internal mic, has no effect.

mackemint:
I can confirm this behavior on 10.15.4 and High Sierra too.

milkdropper:
What audio devices do you see when you toggle through the list? My understanding is that right now the default loopback is WASAPI, which is for Windows only.

mackemint:
I see the input devices I expect to find and in the order they appear under System Preferences -> Sound->Input.

milkdropper:
I can't run the v3.13 release on my Mac to confirm due to https://github.com/projectM-visualizer/projectm/issues/363, but I looked into it on Windows and found some problems. 

I only have one input device and if I toggle over and over my devId keeps incrementing until it crashes. Can you tell me if you see the devID incrementing like below?
```
INFO: Opened audio capture device index=0 **devId=2**: Mic in at rear panel (Pink) (Realtek High Definition Audio)
INFO: Opened audio capture device index=0 **devId=3**: Mic in at rear panel (Pink) (Realtek High Definition Audio)
INFO: Opened audio capture device index=0 **devId=4**: Mic in at rear panel (Pink) (Realtek High Definition Audio)
...
CRITICAL: Failed to open audio capture device: **Too many open audio devices**
```
I think I have this fixed now, and I've got some more bug fixes related to audio input toggling I'll commit tomorrow after I get your feedback.

mackemint:
This is what I get on starting the application and stepping through the inputs with CMD+I:
> 2020-06-07 23:03:11.084 SDLprojectM[2174:95573] INFO: GL_VERSION: 4.1 NVIDIA-14.0.32 355.11.11.10.10.143
> 2020-06-07 23:03:11.084 SDLprojectM[2174:95573] INFO: GL_SHADING_LANGUAGE_VERSION: 4.10
> 2020-06-07 23:03:11.084 SDLprojectM[2174:95573] INFO: GL_VENDOR: NVIDIA Corporation
[...]

PR #368, opened 2020-06-07
Little things

milkdropper:
Some minor things:

- disabled the compilation of native presets in msvc.
- suppressed some unnecessary debug output so the console is more clear.
- made error messages about missing textures more clear (it failed when looking for multiple file extensions).
- separated the presets from textures folder. Note: loadTextureDir checks dataDir/presets dataDir/textures and the presetURL itself. It's more logical to separate them, and that's what milkdrop did as well.


src/libprojectM/Renderer/TextureManager.hpp
```diff
@@ -20,13 +20,13 @@ class TextureManager
...
   TextureSamplerDesc loadTexture(const std::string name, const std::string imageUrl);
   void ExtractTextureSettings(const std::string qualifiedName, GLint &_wrap_mode, GLint &_filter_mode, std::string & name);
-  std::vector<std::string> extensions;
 
 public:
   TextureManager(std::string _presetsURL, const int texsizeX, const int texsizeY,
                  std::string datadir = "");
   ~TextureManager();
 
+  std::vector<std::string> extensions;
```
revmischa, line 29: pls comment

milkdropper:
I did a lot of testing last night with WASAPI enabled, WASAPI disabled, a single input microphone, and multiple input microphones. Based on this testing I made the following changes (some little, and others bigger):

**removed some unnecessary debug from console*

**Issue: selectAudioInput is legacy**
It does nothing and returns 0 by default. I removed this.

**Issue: input toggling does not work on Windows**
On WASAPI devices input toggling will show a toast message and will output to console you are changing devices, but **it doesn't actually start listening to your mic or stop listening to the WASAPI loopback**. I made changes so you can jump from WASAPI loopback to microphone and back when toggling devices in Windows. From the end user perspective it will seem like there is an additional input source called "capture loopback".

Here is an example of console output you will see on screen (with some lines redacted). The toast messages are similar:
`INFO: GL_VERSION: 3.2.0 NVIDIA 446.14
INFO: GL_SHADING_LANGUAGE_VERSION: 1.50 NVIDIA via Cg compiler
INFO: GL_VENDOR: NVIDIA Corporation
INFO: Opened audio capture loopback.
INFO: Opened audio capture device index=0 devId=2: Headset Microphone (HyperX 7.1 Audio)
INFO: Opened audio capture device index=1 devId=2: Rear Mic (Realtek High Definition Audio)
INFO: Opened audio capture device index=2 devId=2: Stereo Mix (Realtek High Definition Audio)
INFO: Opened audio capture loopback.
INFO: Opened audio capture device index=0 devId=2: Headset Microphone (HyperX 7.1 Audio)
INFO: Opened audio capture device index=1 devId=2: Rear Mic (Realtek High Definition Audio)
INFO: Opened audio capture device index=2 devId=2: Stereo Mix (Realtek High Definition Audio)
INFO: Opened audio capture loopback.`

**Issue: input toggling will crash Windows (eventually)**
Each time you input toggle you will register a new device ID (it keeps incrementing). If you keep toggling eventually it will crash with the error "Too many open audio devices". I changed it so it will end the capture with that device ID and reinitialize so it doesn't keep incrementing. I tested a lot to be sure you can reuse the same device ID and change sources with the device Index you pass to it.

**Issue: input toggling is pointless when you are not using WASAPI loopback and there is only one audio source**
I changed it so if you only have one audio source then input toggling will do nothing and just produce a message to console.

milkdropper:
So it seems like audio input toggling never worked. The mac issue reported https://github.com/projectM-visualizer/projectm/issues/365 is legitimate and also exists in Windows (when WASAPI is not used). You see information about toggling but never stop listening to the first device.

msvc/projectMSDL.vcxproj
```diff
@@ -205,6 +205,7 @@
   </Target>
   <Target Name="Build-Presets" BeforeTargets="PrepareForBuild">
     <Exec Command="xcopy /Y /Q /I &quot;$(MSBuildThisFileDirectory)../presets\presets_tryptonaut&quot; &quot;$(OutDir)presets&quot;" />
+    <Exec Command="xcopy /Y /Q /I &quot;$(MSBuildThisFileDirectory)../presets\textures_tryptonaut&quot; &quot;$(OutDir)textures&quot;" />
```
revmischa, line 208: any chance we could install all the presets?

src/projectM-sdl/pmSDL.cpp
```diff
@@ -61,27 +61,45 @@ void projectMSDL::audioInputCallbackS16(void *userdata, unsigned char *stream, i
...
-    initAudioInput();
+        selectedAudioDevice = CurAudioDevice;
+        this->wasapi = false; // Track wasapi as off so projectMSDL will stop listening to WASAPI loopback in pmSDL_main.
+    }
+    else {
+        this->endAudioCapture(); // end current audio capture.
+        CurAudioDevice++; // iterate device index
+        if (CurAudioDevice >= NumAudioDevices) { // We reached outside the boundaries of available audio devices.
+            CurAudioDevice = 0; // Return to first audio device in the index.
+#ifdef WASAPI_LOOPBACK
```
revmischa, line 78: would really love to see this code live in "driver" subclasses rather than all `#ifdef`'d in between the rest of the code in this big file someday

src/projectM-sdl/pmSDL.cpp
```diff
@@ -61,27 +61,45 @@ void projectMSDL::audioInputCallbackS16(void *userdata, unsigned char *stream, i
...
+                return 1;
+            }
+            // If WASAPI_LOOPBACK is not enabled and we have multiple input devices, return to device index 0 and let's listen to that device.
+            selectedAudioDevice = CurAudioDevice;
+            initAudioInput();
+            this->beginAudioCapture();
+#endif
+        }
+        else {
+            // This is a normal scenario where we move forward in the audio device index.
```
revmischa, line 97: thank you for all the comments!

src/projectM-sdl/pmSDL.cpp
```diff
@@ -61,27 +61,45 @@ void projectMSDL::audioInputCallbackS16(void *userdata, unsigned char *stream, i
...
+        this->wasapi = false; // Track wasapi as off so projectMSDL will stop listening to WASAPI loopback in pmSDL_main.
+    }
+    else {
+        this->endAudioCapture(); // end current audio capture.
+        CurAudioDevice++; // iterate device index
+        if (CurAudioDevice >= NumAudioDevices) { // We reached outside the boundaries of available audio devices.
+            CurAudioDevice = 0; // Return to first audio device in the index.
+#ifdef WASAPI_LOOPBACK
+            // If we are at the boundary and WASAPI is enabled then let's load WASAPI instead.
+            projectM::setToastMessage("Listening to loopback audio");
```
revmischa, line 80:
```suggestion
            projectM::setToastMessage("Loopback audio input selected");
```

msvc/projectMSDL.vcxproj
milkdropper: Yes of course - I was just going with what was done previously. I am not sure if there was a reason to use tryptonaut presets by default.

src/projectM-sdl/pmSDL.cpp
milkdropper: I'm happy to do that. But to be consistent toggling inputs should be changed too. Right now it will read "Listening to Microphone rear" but could be changed to "Microphone rear selected"

src/projectM-sdl/pmSDL.cpp
milkdropper: I totally agree the #ifdef stuff looks sloppy and confusing when nested like this. I didn't want to undo what someone else did so I kept with it. But hopefully the comments will keep it OK for now. 

I did A LOT of testing so it should work on windows / mac without issue.

milkdropper:
I accidentally committed the icon change in https://github.com/projectM-visualizer/projectm/pull/368/commits/c870da3c7e8f67f00cc09ec0486c37dcfda23744 , but since I did I thought it's worth raising.

@coast77777777777 - I assume you made the icon based on the EyeTune concept for the Windows Store. Since you renamed it back from EyeTune to projectM would you consider using this Icon instead? I can share the PSD. I think it's a better idea since it uses the M .tga file from the idle preset and github logo.

revmischa:
I refactored the texture and preset loading code in #385

src/projectM-sdl/projectM_SDL_main.cpp
```diff
@@ -393,6 +394,10 @@ modKey = "CMD";
 
     app->init(win, &glCtx);
 
+	// if using built in settings then default to full screen.
+	if (configFilePath.empty())
```
revmischa, line 398: I think pmSDL mostly always uses built-in config. It doesn't install the config file by default

src/libprojectM/Renderer/ShaderEngine.cpp
```diff
@@ -590,7 +590,22 @@ GLuint ShaderEngine::compilePresetShader(const PresentShaderType shaderType, Sha
 
             if (texDesc.first == NULL)
             {
-                std::cerr << "Texture loading error for: " << sampler << std::endl;
+                std::string extensionList = "";
+                // prepare a list of extensions we looked for, so it's more clear what a texture is in error logs.
```
revmischa, line 594: any extension will work, just the first part of the name should match. i suggest removing this code

milkdropper:
I reverted:
- the extension stuff
- the preset / texture folder move stuff (since you are working on that)
- the auto fullscreen (that's part of the issue you saw on macOS with Steam - it slows the load and if you don't see the microphone access request it will be a bad first impression).

src/projectM-sdl/pmSDL.hpp
```diff
@@ -88,6 +88,8 @@ class projectMSDL : public projectM {
 
 
     bool done;
+    bool wasapi = false; // Used to track if wasapi is currently active. This bool will allow us to run a WASAPI app and still toggle to microphone inputs.
+    bool fake_audio = false; // Used to track fake audio, so we can turn it off and on.
```
revmischa, line 92: i guess it should be `fakeAudio` for consistency
```suggestion
    bool fakeAudio = false; // Used to track fake audio, so we can turn it off and on.
```

PR #382, opened 2020-07-24
Fix bug in Windows.

milkdropper:
Can you merge this into master? The config PR submitted by @ycollet seems to have problems. It causes crashes for me on Windows. I think it's because the homedir was backwards, and it aborts without a config when it should continue with built-in config.

### DIFF
--- a/src/projectM-sdl/projectM_SDL_main.cpp
+++ b/src/projectM-sdl/projectM_SDL_main.cpp
@@ -69,9 +69,9 @@ std::string getConfigFilePath(std::string datadir_path) {
   projectM_config = datadir_path;
 
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


revmischa:
Feel free to merge if tests pass

milkdropper:
Should I be able to merge it in? I don't see the option. I think I let to your request to collaborate expire (I didn't notice it).

milkdropper:
Great - I see it now! What standard operating procedures would you like me to follow (beyond checking tests)?

revmischa:
Up to you. We could always use more automated tests in CI. Adding a test for checking if static linking of SDL2 or compilation of aligned alloc works would be super cool. More Qt compile tests

Issue #381, opened 2020-07-22
projectM Qt PulseAudio: configure: error: Failed to find matching components of a complete Qt installation. Try using more options, see ./configure --help.


fvdfgrgtg42etgfw:
As it has been for at least a year, compiling projectM on Ubuntu is a rabbit hole of incompatibilities, manual package downloads, and browser tabs. Today, I tried compiling projectM Qt PulseAudio (latest version 3.1.3) on Ubuntu 20.04. I fixed the libglm error, but here comes this vague "checking correct functioning of Qt installation... failure/configure: error: Failed to find matching components of a complete Qt installation. Try using more options, see ./configure --help." with qt5-default apparently already the latest version.

Button4Button:
Lol I tried exactly the same on Mint 20 this morning since I’ve just
upgraded all my dozen or so machines.

I got as far as the libglm error and sighed. I then recalled that there’s a
set of command line parameters that just worked before that aren’t in the
Readme or webpage.   I can’t find them and need to go back through emails
to find them and try again.

I’d be very happy to update some docs or similar if it means less hassle
for ubu Linux folks in future.

revmischa:
Can you please share what you are running and the output around the error message?
I am running Ubuntu 20.04. I can build v3.1.5 with simply running `./configure && make -j8`

```
checking for shmat... yes
checking for IceConnectionNumber in -lICE... yes
checking for Qt... Info: creating stash file /tmp/tmp.y5VMV6RYv7/.qmake.stash
rmdir: failed to remove '/tmp/tmp.y5VMV6RYv7': Directory not empty
yes:
    QT_CXXFLAGS=-pipe -O2 -Wall -W -D_REENTRANT -fPIC -DQT_NO_DEBUG -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -I. -isystem /usr/include/x86_64-linux-gnu/qt5 -isystem /usr/include/x86_64-linux-gnu/qt5/QtOpenGL -isystem /usr/include/x86_64-linux-gnu/qt5/QtWidgets -isystem /usr/include/x86_64-linux-gnu/qt5/QtGui -isystem /usr/include/x86_64-linux-gnu/qt5/QtCore -I. -I/usr/lib/x86_64-linux-gnu/qt5/mkspecs/linux-g++ 
    QT_DIR=/usr/lib/x86_64-linux-gnu
[...]

PR #387, opened 2020-07-26
Don't require user to provide libglm

revmischa:
We already provide the glm headers in `vendor/`, they just need to be added to the autoconf include path so it can find them.

Should help with #381 glm errors (I assume)


revmischa:
@milkdropper can remove glm from nuget and use the included headers

PR #388, opened 2020-07-28
More toasts... and one more little thing

milkdropper:
I never noticed the Y key can enable/disable the shuffle setting, so I added a toast notification for a better user-experience. I separately added toast messages for manipulating beat sensitivity.

In my documentation review I noticed that N/P (a/k/a lowercase) are reported to load next/previous presets with soft-cut transitions, and SHIFT+N/P are supposed to be hard-cut transitions. Currently SDL treats lowercase and SHIFT+N/P the same: they are hard cut transitions.

I think I fixed this up and changed the documentation (it was wrong as lowercase should be hard cut and SHIFT should be soft cut).

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,10 @@ See [BUILDING.md](BUILDING.md)
 * Down: decrease beat sensitivity (min 0)
 * Y: toggle shuffle enabled
 * R: jump to random preset
-* N: next preset
-* P: previous preset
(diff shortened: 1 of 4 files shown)


milkdropper:
Looks like AppVeyor failed because I had special characters in the title. I changed them.

revmischa:
Would be great to get toasts working on the macOS build

PR #360, opened 2020-05-24
Upgrade LLVM JIT

revmischa:
For new LLVM version (9+)
Doesn't quite work. Need help.

Issue #168, opened 2019-03-03
Support different GLSL versions


revmischa:
*Problem*: different systems and drivers have support for different versions of GLSL. ProjectM will not run unless the shader language matches the supported driver version.

As an example, my raspberry pi outputs:
```
INFO: GL_VERSION: OpenGL ES 2.0 Mesa 13.0.6
INFO: GL_SHADING_LANGUAGE_VERSION: OpenGL ES GLSL ES 1.0.16
INFO: GL_VENDOR: Broadcom
Failed to compile shader 'Vertex: v2f_c4f'. Error: 0:2(1): error: syntax error, unexpected NEW_IDENTIFIER
Failed to compile shader 'Fragment: v2f_c4f'. Error: 0:3(1): error: `in' qualifier in declaration of `fragment_color' only valid for function parameters in GLSL ES 1.00
0:4(1): error: `out' qualifier in declaration of `color' only valid for function parameters in GLSL ES 1.00
```

And @servimo [reports](https://github.com/projectM-visualizer/projectm/issues/148#issuecomment-468976691) the error: 
`error: GLSL 3.30 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.00 ES, and 3.00 ES`  
And again in #101 

How can we support multiple versions of GLSL, ES and non-ES?

Also important: detect if GLES is needed automatically in configure.ac so that users don't have to do `--enable-gles` manually. A decent number of issues on here are related to people not specifying that they need GLES.

More info: #115 #101 

@deltaoscarmike could really use your help here

champagj:
Thank you @revmischa for the new thread. 

Trying to run ProjectM on pi 3 and finding it difficult with my limited programming skills.. Back to linux after many years. 

Looks like it may be possible to run via QT but also difficult to install/config to run ProjectM.

If you need an "intermediate" tester, I'm here.

peak3d:
@revmischa gl version is usuallly detected during runtime and not compile time.
But for my GLSL 100 (GLES 2.0) devices there are more changes in shaders needed to get it working

The easiest thing is the version ( 5 minutes work) and the in / out (what is not supported by GLSL 100)
Mor interesting are the   layout statements (IIRC) which are not part of GLSL 100

revmischa:
Yes, i agree it would be great if it could detect the right shader code to load at runtime rather than compile time.

sergiomb2:
[...]

PR #399, opened 2020-08-05
Backport GLSL shaders to version 1.2 to support older hardware.

fughilli:
This PR backports the GLSL hardcoded shaders in ShaderEngine.cpp to GLSL1.2 and also adds another target version to hlslparser to export GLSL1.2. These changes enable running ProjectM on hardware which supports only older GLSL versions, such as the VideoCore IV GL implementation used by the Raspberry Pi (versions <=3).


revmischa:
Yo, this is terrific! Thanks!

I think older Pis may have issues anyway due to CPU limitations (mine certainly did running without shaders)

If this means we can require a lower minimum GL version this should be updated - https://github.com/projectM-visualizer/projectm/blob/master/src/projectM-sdl/projectM_SDL_main.cpp#L307

fughilli:
CPU limitations are definitely an issue. I'm using the Pi to run an addressable LED array with ~3000 pixels, so much smaller raster and grid size is OK, plus I'm curating presets that run smoothly. I mentioned in #168 that async would help with this. Another potential plus would be supporting vectorization on ARM (I see that only SSE intrinsics are used for acceleration currently). One way to do this scalably would be to rewrite the fixed-function parts of the per-frame and per-vertex equations in [Halide](https://halide-lang.org/).

fughilli:
Re: keeping things in sync with upstream; any reason this project is not making use of submodules and/or a build system with imports? I used Bazel for my [fork](https://github.com/fughilli/libprojectm-bazel-glsl2) and that also allowed me to support cross-compilation (compiling projectM from source every time on the Pi itself is absolutely miserable).

revmischa:
Async would help. 
Enabling LLVM JIT compilation of presets may speed things up for the CPU and utilize things like vectorization.  A little bit of work is needed to make `--enable-llvm` work with recent versions of `llvm`  - https://github.com/projectM-visualizer/projectm/pull/360

revmischa:
Upgrading build system is fine, though it may make downstream projects 😞 
It was CMake long ago and it was a huge pain in the ass to deal with

revmischa:
Tested on macOS Catalina:
```
2020-08-08 15:07:41.885 projectMSDL[52080:1382685] INFO: GL_VERSION: 4.1 ATI-3.9.15
2020-08-08 15:07:41.885 projectMSDL[52080:1382685] INFO: GL_SHADING_LANGUAGE_VERSION: 4.10
2020-08-08 15:07:41.885 projectMSDL[52080:1382685] INFO: GL_VENDOR: ATI Technologies Inc.
2020-08-08 15:07:41.886 projectMSDL[52080:1382685] INFO: Using data directory: /usr/local/share/projectM
2020-08-08 15:07:41.886 projectMSDL[52080:1382685] INFO: Config file not found, using built-in settings. Data directory=/Users/cyber/dev/projectm/src/projectM-sdl/
Failed to compile shader 'Vertex: v2f_c4f'. Error: ERROR: 0:1: '' :  version '120' is not supported
ERROR: 0:2: '' :  #version required and missing.
ERROR: 0:2: 'attribute' : syntax error: syntax error

Failed to compile shader 'Fragment: v2f_c4f'. Error: ERROR: 0:1: '' :  version '120' is not supported
ERROR: 0:2: '' :  #version required and missing.
ERROR: 0:2: 'varying' : syntax error: syntax error

Failed to link program: ERROR: One or more attached shaders not successfully compiled
```

fughilli:
Maybe we could query the available GL version using [glGetString](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glGetString.xml) and then select a variant of the shaders that way? If we go that route, I would propose taking the shader code out of string literals and instead either load them in at runtime or compile them in as binary resources.

revmischa:
Yeah makes sense

fughilli:
I ran a small test program on both my RPi 3A and my Razer Blade 14" 2017:

```
... GL context setup ...
std::cout << "GL version: " << glGetString(GL_VERSION) << std::endl;
...
```

Razer Blade:
GL version: 4.6.0 NVIDIA 390.132

Raspberry Pi:
GL version: 2.1 Mesa 13.0.6

Looking at [this table](https://en.wikipedia.org/wiki/OpenGL#Version_history), I can map the version number to a GLSL version. I'm thinking as a first pass that we can select the GLSL1.2 code for GL version <= 3.2, and the original GLSL3.3 code for GL version >= 3.3.0.

fughilli:
Added another commit to the PR which refactors the shaders out into a separate manager. The manager queries the supported GLSL version and picks either the original GLSL3.3 shaders or the backported GLSL1.2 shaders depending on the result of the query.

fughilli:
I also opened #402

fughilli:
It looks like the builds are failing because the new sources are not included in the project manifests for the XCode and Visual Studio builds. Does someone else have the time to add them?

milkdropper:
I'll do the visual studio bits tonight.

fughilli:
What remains to be done to get this merged? @milkdropper do you have a fixed VS project file? Also I think the XCode project file needs to be updated--I don't have a Mac, so I can't do it.

milkdropper:
> What remains to be done to get this merged? @milkdropper do you have a fixed VS project file? Also I think the XCode project file needs to be updated--I don't have a Mac, so I can't do it.

I'll give it another shot with msvc. I actually had problems last time I tried, but will take a look with a fresh set of eyes.

fughilli:
Nice! Thanks for the assist!

[gif](https://media.giphy.com/media/9eHzCa8QO0UTK/giphy.gif)

PR #400, opened 2020-08-07
Preset browser

milkdropper:
- [M] or [m] key loads a list of presets.
- If you are on the idle preset, nothing is highlighted.
- You can navigate presets with Up/Down Arrows and Page Up/Page Down. (Page Up/Page Down moves up by half the screen & Up/Down Arrows by 1 preset)
- If you navigate to the top or the bottom of the list it will rotate to the top or bottom.
- You can tell what is selected by the color.
- You can verify with F3.
- If a ransom preset is loaded by a timer or by [R], it will load the expected page at the expected offset.
- The menu impacts FPS. But it shouldn't be a problem since it's only temporary until you find what you want.
- Toasts are now on top of any text menu, as this has a greater chance of filling the middle of the screen with text. But the help menu was already conflicting.


milkdropper:
And the demo: https://streamable.com/8xf3wk

PR #412, opened 2020-08-28
Fix blank preset bug

psperl:
I haven't tested this because my only Linux box is headless, but it compiles and is the same fix I applied to the Android branch in Feb 2020.  It makes sure than an interrupted preset blend doesn't accidentally leave the master alpha set to a non-1.0 value.  The fix is not very elegant, but it is quite effective.

### DIFF
--- a/src/libprojectM/PipelineMerger.cpp
+++ b/src/libprojectM/PipelineMerger.cpp
@@ -5,6 +5,14 @@
 const double PipelineMerger::e(2.71828182845904523536);
 const double PipelineMerger::s(0.5);
 
+void PipelineMerger::ensureAlphaIsNotBlended(const Pipeline & a)
+{
+	for ( std::vector<RenderItem*>::const_iterator pos = a.drawables.begin(); pos != a.drawables.end(); ++pos )
+	{
+		( *pos )->masterAlpha = 1.0;
+	}
+}
+
 void PipelineMerger::mergePipelines(const Pipeline & a, const Pipeline & b, Pipeline & out, RenderItemMatcher::MatchResults & results, RenderItemMergeFunction & mergeFunction, float ratio)
 
 {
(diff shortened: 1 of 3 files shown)


src/libprojectM/PipelineMerger.cpp
```diff
@@ -5,6 +5,14 @@
 const double PipelineMerger::e(2.71828182845904523536);
 const double PipelineMerger::s(0.5);
 
+void PipelineMerger::ensureAlphaIsNotBlended(const Pipeline & a)
+{
+	for ( std::vector<RenderItem*>::const_iterator pos = a.drawables.begin(); pos != a.drawables.end(); ++pos )
```
revmischa, line 10:
```suggestion
	for (auto &pos : a.drawables)
```
I think is equivalent?

revmischa (approved):
Thanks for this!

src/libprojectM/PipelineMerger.cpp
psperl: It is.  Up to you if you want to require C++11.

revmischa: I believe we do already and I don't see a reason not to. It's fine without this change though

milkdropper:
Hi @psperl - I am memory access violations in Windows linked to ( *pos )->masterAlpha = 1.0 in ensureAlphaIsNotBlended. Any recommendations? If not, how can I reproduce this bug to see I can tweak your PR to be effective without causing this issue.

psperl:
I'm sorry about that. I just ported this straight over from the android
branch and only confirmed it compiled.  I don't have a windows dev env set
up and my Linux server is headless.

Do we have confirmation that it works on other platforms? Does it crash for
you at startup or sporadically?  This is kind of a safe thing to do, so
it's a bit curious.

Issue #413, opened 2020-08-31
Shader compilation failure breaks preset


fughilli:
Right now, if a shader fails to compile, we continue to run it anyways. Alternatively, we could check the status of the shader compilation and skip the preset if it fails to compile.

kblaschke:
This issue is still present in current master, so some additional measures need to be taken. I'd also suggest removing the offending preset from the playlist (or mark it "broken"), so it's not being loaded again. If it is loaded over and over again, that might result in projectM getting stuck in an endless loop and possibly hang if the playlist only contains broken presets.

audreyspency:
is it solved yet ?

kblaschke:
> is it solved yet ?

Not yet. We're currently pushing to the 4.0 release, with focus on the API and frontends. We'll get back to bug fixing immediately after the release is out. This one looks like a low-hanging fruit, so it might be solved shortly after.

kblaschke:
Some HLSL shaders in presets use global variables like this:

```
warp_2=`sampler sampler_pw_noise_lq;
warp_3=`float streetx, streety,streetr,street; 
warp_4=`float2 uv0, uv1, uv2, uv3;
warp_5=`static const float slim = 0.99;
warp_6=`
...
warp_15=`streetx = 1- min(sat(p1*abs(uv2.x) -p2),sat(2*p1*abs(abs(uv1.x)-0.25)-p2)); 
warp_16=`streety = 1- min(sat(p1*abs(uv2.y) -p2),sat(2*p1*abs(abs(uv1.y)-0.25)-p2));
warp_17=`streetr = 1- sat(p1*abs(rad/aspect.y*1.07-0.5) -p2*1.75);
```

I'm not proficient enough with HLSL to tell if this is alllowed, but in OpenGL at least there are no shader-global variables. Anything declared outside a function is considered a read-only uniform and thus can't be assigned within a function.

This breaks a good number of presets, e.g. `martin - city lights v2 c.milk`. If anyone has a good solution on how to convert this properly to GLSL, I'd be very interested.

kblaschke:
At the current point, I think the best approach for these presets would be identifying them (e.g. loading each preset with shaders and see if they compile in projectM), then put in some manual work and rewrite the shaders so they don't use global variables anymore. Most presets I've seen wouldn't even need them as they only have the `shader_body` function, others a bit more complicated.

Rewriting the shader code wouldn't affect Milkdrop parsing these files, and allow projectM to play them properly.

[...]

PR #414, opened 2020-08-31
Add mitigation for shader compilation failure

fughilli:
Re: #413

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -67,6 +67,10 @@ pthread_mutex_t preset_mutex;
 #endif
 #endif
 
+namespace {
+constexpr int kMaxSwitchRetries = 10;
+}
+
 projectM::~projectM()
 {
 #ifdef USE_THREADS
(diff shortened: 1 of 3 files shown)


src/libprojectM/projectM.cpp
```diff
@@ -837,20 +844,26 @@ void projectM::switchPreset(const bool hardCut) {
     }
 
     populatePresetMenu();
+
+    return result;
```
revmischa, line 848: Not really clear to me why returning a string instead of bool.

revmischa (approved):

src/libprojectM/projectM.cpp
fughilli: Yeah I also don't know why I did this. Changing it to bool.

fughilli: Ah I remember. It was to keep the return type consistent with the other overload. Why do we have two overloads for this function anyways? Seems like the overload taking `unique_ptr<Preset>&` should be named `switchToPreset`, and the other overload should be called `startTransition` or something like this.

PR #415, opened 2020-08-31
Revert "Fix blank preset bug (#412)"

milkdropper:
This reverts commit f8cea68105a858186c9ef0cd3c510505fbd7c753.

### DIFF
--- a/src/libprojectM/PipelineMerger.cpp
+++ b/src/libprojectM/PipelineMerger.cpp
@@ -5,14 +5,6 @@
 const double PipelineMerger::e(2.71828182845904523536);
 const double PipelineMerger::s(0.5);
 
-void PipelineMerger::ensureAlphaIsNotBlended(const Pipeline & a)
-{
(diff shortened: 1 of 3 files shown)


src/libprojectM/PipelineMerger.cpp
```diff
@@ -5,14 +5,6 @@
 const double PipelineMerger::e(2.71828182845904523536);
 const double PipelineMerger::s(0.5);
 
-void PipelineMerger::ensureAlphaIsNotBlended(const Pipeline & a)
-{
-	for ( std::vector<RenderItem*>::const_iterator pos = a.drawables.begin(); pos != a.drawables.end(); ++pos )
-	{
-		( *pos )->masterAlpha = 1.0;
```
revmischa, line 12: i don't really understand the error, so I don't know if a null ptr check here is sufficient or that's just hiding the real problem 
@psperl

PR #430, opened 2020-11-10
On linux with --enable-gles, the obtaining of the shader major and mi…

hibengler:
…nor version had a bug.

The old code assumed <major> '.' <minor> [' ' other stuff ]
But on Linux the string being parsed has other stuff at the beginning.
This would cause the atoi calls to raise an exception.

I solved it with inline C code that handles all cases.

### DIFF
--- a/src/libprojectM/Renderer/StaticGlShaders.cpp
+++ b/src/libprojectM/Renderer/StaticGlShaders.cpp
@@ -1,5 +1,7 @@
 #include "StaticGlShaders.h"
 #include "projectM-opengl.h"
+#include <cstring>
+
 
 namespace {
 // Variants of shaders for GLSL1.2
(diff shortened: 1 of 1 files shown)


hibengler:
Fixed a problem with linux desktop --enable-gles where an atoi caused a raised exception

src/libprojectM/Renderer/StaticGlShaders.cpp
```diff
@@ -770,16 +772,74 @@ StaticGlShaders::StaticGlShaders(bool use_gles) : use_gles_(use_gles) {
...
+    if (version_len) { // find the number
+      size_t position = 0; 
+      char *cstr = new char [version_len+1];
+
+      strcpy(cstr,glsl_version_string.c_str()); 
+      
+      /* scan the anything before the number */
+      while (position<version_len) {
+        char ch = cstr[position];
+	if ((ch >= '0') || (ch <= '9')) {
```
revmischa, line 802: I think this is always true?

hibengler: OOPS.
The || should be && here and two other places.

PR #431, opened 2020-11-11
 StaticGlShaders::QueryGlslVersionTry 2: --enable-gles 

hibengler:


hibengler:
If --enablke-gles is configured, on a linux desktop, the GL_SHADING_LANGUAGE_VERSION parameter is "OpenGL ES GLSL ES 3.10"

However, StaticGlShaders::QueryGlslVersion assumed that the string would be more like "[blah ]N.NN blah"
This was causing exceptions in the atoi calls to retrieve the major and minor numbers, just for GLES2

So I revamped the scanning code.


I had minor code issues, so revamped the code and walked through it with the debugger.

revmischa:
Sucks that https://www.khronos.org/opengl/wiki/Detecting_the_Shader_Model lies
Indentation looks very inconsistent

hibengler:
I agree.

PR #435, opened 2020-11-25
Use C locale for floats in HLSLParser

tatokis:
The HLSL parser uses locale-aware functions to parse the shaders, which results in errors parsing all shaders under certain locales.

For example, when a decimal point on a float is denoted by a `,` instead of a `.`.
This results in 
```
projectM/presets/presets_milkdrop_200/Geiss - Cosmic Dust 2 - cloud journey.milk(88) : Syntax error: expected ',' near '.'
Failed to parse HLSL(step2) Comp shader
```

This PR does something similar to https://github.com/projectM-visualizer/projectm/commit/dfd335f853b8711262647ecf67e16d8b790d9969 but for HLSLParser.

### DIFF
--- a/src/libprojectM/Renderer/hlslparser/src/Engine.cpp
+++ b/src/libprojectM/Renderer/hlslparser/src/Engine.cpp
@@ -3,8 +3,9 @@
 
 #include <stdio.h>  // vsnprintf
 #include <string.h> // strcmp, strcasecmp
-#include <stdlib.h>	// strtod, strtol
-
+#include <stdlib.h>	// strtol
+#include <locale>
+#include <sstream>
 
 namespace M4 {
 
@@ -40,7 +41,11 @@ int String_Printf(char * buffer, int size, const char * format, ...) {
 }
 
 int String_FormatFloat(char * buffer, int size, float value) {
-    return String_Printf(buffer, size, "%f", value);
+    std::ostringstream oss;
+    oss.imbue(std::locale("C"));
+    oss << value;
+
+    return String_Printf(buffer, size, "%s", oss.str().c_str());
 }
 
 bool String_Equal(const char * a, const char * b) {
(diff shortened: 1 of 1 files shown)


src/libprojectM/Renderer/hlslparser/src/Engine.cpp
```diff
@@ -59,8 +64,31 @@ bool String_EqualNoCase(const char * a, const char * b) {
 #endif
 }
 
+static inline double iss_strtod(const char * in, char ** end) {
+    double df;
+    std::istringstream iss(in);
+    iss.imbue(std::locale("C"));
+    iss >> df;
+    if(iss.fail()) {
+        *end = (char*)in;
```
afeinberg, line 73: consider using something other than (char *) -- static_cast<> is ideal, thgen dynamic_cast<> then, reinterpret_cast<> (equivalent to doing a C style cast.)

src/libprojectM/Renderer/hlslparser/src/Engine.cpp
```diff
@@ -59,8 +64,31 @@ bool String_EqualNoCase(const char * a, const char * b) {
 #endif
 }
 
+static inline double iss_strtod(const char * in, char ** end) {
```
afeinberg, line 67: Some third party libraries may already do this:

folly:::Conv https://github.com/facebook/folly/blob/master/folly/docs/Conv.md
boost::lexical_cast https://www.boost.org/doc/libs/1_42_0/libs/conversion/lexical_cast.htm

afeinberg, line 67: nit: let the compiler decide for itself if it wants to inline this or not: is my advice.

alternativelhy you can use an anonymous namespace instead of static

```namespace  {
// what would be a static before
} // anonymous namespace
```

tatokis: If I understand correctly, using either of those would bring in another dependency, which isn't really a good idea for only this specific thing. I'm also not sure if those libraries support getting the end of the float in the string, which is needed for the parser.
I may be wrong though as I am not familiar with either of those, and I only had a quick glance at the links.

Additionally, as mentioned, the rest of the code does the same thing, which is why I think it's the best course of action.
https://github.com/projectM-visualizer/projectm/blob/fd1d235a059714a3cdf1504d7ffe4ef4c286fe02/src/libprojectM/MilkdropPresetFactory/Parser.cpp#L1285-L1300

tatokis: Well, inline is a suggestion to the compiler. It doesn't force it to do anything. The static was there to make inlining it easier.
I do agree though. It was only done like that because it's used in that specific place.

If you think it's better, I can make it a normal function that's part of the class, private or public.

revmischa:
Good to merge?

revmischa:
Ok, thanks a lot for this improvement!

PR #436, opened 2020-11-26
Fix various warnings

tatokis:
This fixes almost all warnings in the codebase when compiling on Linux under gcc 9.3.

Only one I couldn't figure out, which is possibly a false compiler warning, is
```
qprojectm_mainwindow.cpp: In member function ‘void QProjectM_MainWindow::updateFilteredPlaylist(const QString&)’:
qprojectm_mainwindow.cpp:1258:68: warning: ‘activePresetId’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 1258 |    if (activePresetId.hasValue() && data.id == activePresetId.value()) {
      |                                                ~~~~~~~~~~~~~~~~~~~~^~
```

It complains about `activePresetId` being uninitialized, however it's a class instance and the default constructor is called.
`activePresetId.value()` above won't be called if it's never been assigned (as `hasValue()` only returns true if `operator=` has been called).


revmischa:
Wow cool!

PR #443, opened 2021-01-08
Search menu fix

milkdropper:
This fixes a bug where the search menu would not move forward with down arrow key.

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -929,6 +929,7 @@ void projectM::selectNext(const bool hardCut) {
         }
         else {
             // otherwise move forward 
+            renderer->m_activePresetID++;
             selectPresetByName(renderer->m_presetList[renderer->m_activePresetID].name,true);
         }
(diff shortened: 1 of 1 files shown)


sandiMexicola:
Wait.  There's a search menu?  :P

milkdropper:
Yeah - I thought I updated the menu in master but guess I forgot! 

- Use [m] to open preset menu / navigator. 
- Use [enter key] to search the presets (if the preset list wasn't already up, it will open)

sandiMexicola:
Wow!  Thanks for the info!  This might make the handling of 45k presets a tad bit easier! 💯

sandiMexicola:
I should also mention: thanks for the whole project!  It is really cool! :)

src/libprojectM/projectM.cpp
```diff
@@ -929,6 +929,7 @@ void projectM::selectNext(const bool hardCut) {
         }
         else {
             // otherwise move forward 
+            renderer->m_activePresetID++;
```
revmischa, line 932: can it exceed the bounds?

milkdropper: No that's covered already. A few lines above there is a check to see if we hit the boundary. If so, it will rotate back to the top when you reach the bottom.

revmischa (approved):

revmischa:
This good to merge @milkdropper ?

PR #312, opened 2020-01-27
new BeatDetect implementation

mbellew:
This is a proposed fix for https://github.com/projectM-visualizer/projectm/issues/309

I think this simpler implementation does what it says it does.  At least it makes more sense to me.  It doesn't really seem to have that much noticeable affect how the presets render.  I wrote a new test preset so you can more directly evaluate the output.  

https://github.com/projectM-visualizer/projectm/blob/master/presets/tests/300-beatdetect-bassmidtreb.milk

It would be great if the regular contributors could 

  a) review the code change carefully (not that big) 
  b) verify your favorite presets
  c) if you like, play with various parameters (e.g. freq cutoffs) and see if the current version seems best.

PR #450, opened 2021-02-06
Fix #408: Parser can blow up if no [preset00] header

### LINKED ISSUE #408: ProjectM Freezes/Deadlocks with Certain Presets
22nd of August 2020

I am reporting four separate issues I am experiencing with ProjectM. I started having problems when I installed the massive community preset pack available on the ProjectM GitHub home page. Previously, I was not experiencing any problems. It seems that there are certain presets in there that cause problems. In some cases, I was able to determine a few of these that were causing problems, and reproduce the errors. In other cases, I was not able to do so.

This is the second of these issues – certain presets that cause ProjectM to freeze/deadlock. (See the attached Freezing.Presets.zip.)

I did some fairly extensive investigations into these problems in June this year, and was able to get as far as discovering that there are four separate issues, according to backtraces that I obtained with GDB.

My use case is that I consider myself to be a Kodi power user. I use Kodi to play music on my media server most of the time that I am home. I have the ProjectM addon installed for Kodi, and I love it. I only use the ProjectM PulseAudio executable when, for example, I am streaming music via a browser. But I was able to determine that these problems originate in the ProjectM library itself, not within Kodi, and they seem to affect both Kodi with the ProjectM visualiser installed, as well as the ProjectM PulseAudio executable.

The problems fall into two classes, crashes (three problems found), and freezes/deadlocks (one problem found). As of this writing, with the latest available software, both these problems still exist.

I love the new presets I have installed, and even though I installed them over two months ago, I am still surprised by the occasional preset that I have never seen before. I have been collecting presets for a long time now, but some of the presets in the community pack are stunning!

Nevertheless, it is annoying when my music occasionally stops and I have to kill/restart Kodi and try to figure out where it was before it crashed. Most annoying when I am listening to DJ sets of an hour or longer!

Kodi reports that I now have 40585 presets installed, so, as you can imagine, it would be an extremely painful exercise for me to manually test every single preset individually and remove the ones that cause problems, certainly, I don’t have that kind of time to spend. Also, I imagine that all of these presets used to work at some point in the Milkdrop/ProjectM lifecycle, and ideally, this should be fixed inside the ProjectM library itself, or in cases where there is a bad preset that does not parse, the exception it causes should be handled, and ProjectM should simply move on to the next preset, possibly adding the faulty preset to some kind of banned list. That would be a far more robust way of handling the problem, to my mind.

Hardware:

My media server is based on an HP Proliant MicroServer, Gen10 (WW Entry BTO Model), specs here:

https://h20195.www2.hpe.com/v2/GetDocument.aspx?docname=a00008701enw&doctype=quickspecs&doclang=EN_US&searchquery=&cc=us&lc=en

Briefly, the hardware is as follows:

CPU: AMD Opteron(TM) X3216 APU - Dual-core, x86_64
Memory-Main Memory-phys_mem: 8053063680
Monitor: LG TV
Network Cards: NC332i Adapter x2, RTL8192EE PCIe Wireless Network Adapter
Graphics Card: ATI Radeon R5 Graphics
Sound Card: Kabini HDMI/DP Audio (built into graphics card).

The hardware info as reported by YaST may be found here:

https://paste.kodi.tv/lilaqabiqe.kodi

Sample Xorg.0.log may be found here:

https://paste.kodi.tv/serijohada

I am experiencing this problem on OpenSUSE Tumbleweed, latest version, updated today (22nd of August 2020), with Kodi Leia 18.8, installed from the packman repo. (Tumbleweed is OpenSUSE’s rolling release.)

However, I don’t think this is a problem with either the hardware or the operating system.

Sample ProjectM PulseAudio backtrace:

https://paste.kodi.tv/ubecosajew

Sample Kodi backtrace:

https://paste.kodi.tv/texoniquha.kodi

Sample Kodi log:

https://paste.kodi.tv/agasapuyut.kodi

[Freezing.Presets.zip](https://github.com/projectM-visualizer/projectm/files/5112695/Freezing.Presets.zip)

revmischa:
I loaded up the broken presets from #408 and debugged the `Parser::parse_top_comment` function. Turns out the parser can freeze or crash if a preset file does not contain the standard `[preset00]` header. If no `[` is found in the file it scans until the buffer is full (which results in an infinite loop) or it overflows something and blows up.
I changed the parser to just proceed normally if there is no header. We don't do anything with the header anyway.

Tested and the bad presets no longer crash or freeze for me. 😃 

Also fixed:
* skip paths in hidden system `__MACOSX` directories
* add `--enable-debug` configure flag which enables debug mode and disables optimizations
* ensure fft samples are a power of two


src/libprojectM/PCM.cpp
```diff
@@ -338,7 +338,6 @@ void PCM::getPCM(float *PCMdata, int samples, int channel, int freq, float smoot
    if (freq)
      {
        // NOTE some presets set bSpectrum=1 and samples!=2^n, not sure what rdft() does with that
-       assert(samples <= 1024);
```
revmischa, line 341: https://github.com/projectM-visualizer/projectm/pull/312/files#r571438718
This is causing a crash for me, I'm not sure what the solution is here, but the next line seems to handle it sort of gracefully? @mbellew

revmischa: After removing this line, I can trigger a crash in the fft code :(
```
Process 93461 stopped
* thread #1, name = 'projectMSDL', stop reason = signal SIGSEGV: invalid address (fault address: 0x7fff9ac26d98)
    frame #0: 0x00005555555838e4 projectMSDL`cftmdl1(n=540, a=0x00007fffffffa0e0, w=0x0000555555a79cc0) at fftsg.cpp:2505:19
   2502	        x0r = x1r + x3i;
   2503	        x0i = x1i - x3r;
   2504	        a[j3] = wk3i * x0r + wk3r * x0i;
-> 2505	        a[j3 + 1] = wk3i * x0i - wk3r * x0r;
   2506	    }
   2507	    j0 = mh;
   2508	    j1 = j0 + m;
```

revmischa: Possible fix: 8e46235f

src/libprojectM/PCM.cpp
```diff
@@ -333,20 +333,40 @@ void PCM::getPCM(float *PCMdata, int samples, int channel, int freq, float smoot
...
+				 // fft input
+				 temppcm[i] = (double)PCMdata[i];
+			 }
+			 // fft
+			 rdft(samples_pow2, 1, temppcm, ip, w);
+			 for (int j = 0; j < samples; j++)
+			 {
+				 // get fft output, if we had to get a fewer number of samples than requested
+				 // then we just fill in the rest with 0. this is lame.
+				 // TODO: better solution if requested number of samples is not a power of 2
```
revmischa, line 366: I feel like we're sacrificing some accuracy here

labkey-matthewb: I see.  This happens when a waveform has bspectrum==true and samples set to a strange size.  You know this class is constantly recomputing the spectrum data on every call to addPCM().  It's sitting in vdataL and vdataR.  

Maybe it makes more sense to change CustomWave read directly from pcmdataL/R or vdataL?R?

labkey-matthewb: That's not quite right, because Wafeform uses the smoothing parameter.  (PCM has an awkward interface).  In any event, maybe it makes sense to use vdataL when spectrum==true?

Doing an fft on 32 samples doesn't make much sense...  Grabbing the first 32 values out of the larger FFT seems more useful.

labkey-matthewb: I think we can postpone the PCM.cpp part of this PR, and maybe look into changing how Waveform to pulls data.  

I'm reviewing how code uses PCM and will look at that.  BTW, I noticed that BeatDetect is _still_ pulling from pcmdata instead of vdata.  Didn't catch that with my previous beatdetect change... Look for PR maybe next weekend?

revmischa: Gotcha ok. Feel free to edit/amend this PR. Come chat on Discord if you want!

revmischa: Should I undo these changes?

mbellew: Yes, undo the PCM change for now.  I'll look at the whole Waveform/PCM behavior as one piece and with an eye toward milkdrop2 behavior.

revmischa: Ok done. That would be great. I think right now this is a pretty urgent security/stability issue

mbellew (approved):

Issue #407, opened 2020-08-22
ProjectM Crashes with Certain Presets


stevenswart:
22nd of August 2020

I am reporting four separate issues I am experiencing with ProjectM. I started having problems when I installed the massive community preset pack available on the ProjectM GitHub home page. Previously, I was not experiencing any problems. It seems that there are certain presets in there that cause problems. In some cases, I was able to determine a few of these that were causing problems, and reproduce the errors. In other cases, I was not able to do so.

This is the first of these issues – certain presets that cause ProjectM to crash. (See the attached Crashing.Presets.zip.)

I did some fairly extensive investigations into these problems in June this year, and was able to get as far as discovering that there are four separate issues, according to backtraces that I obtained with GDB.

My use case is that I consider myself to be a Kodi power user. I use Kodi to play music on my media server most of the time that I am home. I have the ProjectM addon installed for Kodi, and I love it. I only use the ProjectM PulseAudio executable when, for example, I am streaming music via a browser. But I was able to determine that these problems originate in the ProjectM library itself, not within Kodi, and they seem to affect both Kodi with the ProjectM visualiser installed, as well as the ProjectM PulseAudio executable.

The problems fall into two classes, crashes (three problems found), and freezes/deadlocks (one problem found). As of this writing, with the latest available software, both these problems still exist.

I love the new presets I have installed, and even though I installed them over two months ago, I am still surprised by the occasional preset that I have never seen before. I have been collecting presets for a long time now, but some of the presets in the community pack are stunning!

Nevertheless, it is annoying when my music occasionally stops and I have to kill/restart Kodi and try to figure out where it was before it crashed. Most annoying when I am listening to DJ sets of an hour or longer!

Kodi reports that I now have 40585 presets installed, so, as you can imagine, it would be an extremely painful exercise for me to manually test every single preset individually and remove the ones that cause problems, certainly, I don’t have that kind of time to spend. Also, I imagine that all of these presets used to work at some point in the Milkdrop/ProjectM lifecycle, and ideally, this should be fixed inside the ProjectM library itself, or in cases where there is a bad preset that does not parse, the exception it causes should be handled, and ProjectM should simply move on to the next preset, possibly adding the faulty preset to some kind of banned list. That would be a far more robust way of handling the problem, to my mind.

Hardware:

My media server is based on an HP Proliant MicroServer, Gen10 (WW Entry BTO Model), specs here:

[...]

PR #455, opened 2021-02-07
parse completely empty line

mbellew:
parse completely empty line e.g.

```shape_1_per_frame1=shpt =```

seems to address https://github.com/projectM-visualizer/projectm/issues/407


mbellew:
BTW, what is the best option for  the original milkdrop in action to compare projectm vs milkdrop?

revmischa:
> BTW, what is the best option for installing the original milkdrop to compare against projectm?

That I'm not sure about. I think you need windows? There is a web-based on - https://butterchurnviz.com/
https://github.com/jberg/butterchurn

revmischa (approved):

PR #457, opened 2021-02-08
Wavemode7

mbellew:
I don't know GL very well, but this fix seems to make the double line wave render properly. I just loop through the main body of ::Draw() twice if necessary.

 Added new test presets to verify.


Makefile.am
```diff
@@ -19,7 +19,7 @@ pm_font__DATA = fonts/Vera.ttf fonts/VeraMono.ttf
 # find and install all preset files
 install-data-local:
 	find "$(PRESETSDIR)" -type d -exec $(MKDIR_P) "$(DESTDIR)/$(pm_data_dir)/{}" \;
-	find "$(PRESETSDIR)" -type f -exec $(INSTALL_DATA) "{}" "$(DESTDIR)/$(pm_data_dir)/{}" \;
+	find "$(PRESETSDIR)" -path "$(PRESETSDIR)/tests" -prune -o -type f -exec $(INSTALL_DATA) "{}" "$(DESTDIR)/$(pm_data_dir)/{}" \;
```
revmischa, line 22: What's this do?

revmischa (approved):

Makefile.am
mbellew: I noticed that the test presets get copied when you do "make install".  This skips the tests directory.  Hmm, should probably do that on the line above as well...  

As always opinions/testing appreciated!

mbellew: PS: I followed the link on this page to install winamp/milkdrop2.  http://www.geisswerks.com/milkdrop/

revmischa: Hm ok I'm not an advanced `find` user - I guess that's doing two different things in one statement? wild

PR #458, opened 2021-02-11
Default param values

mbellew:
Some of the current default param values cause bad behavior in the pixel mesh calculation or are markedly different from milkdrop.


revmischa (approved):
Great! I'll have to try this. Really happy to see some comments on these things too 👍🏻

Issue #459, opened 2021-02-11
shapecode_#_textured=1 doesn't seem to work


mbellew:
Shapes that use textured=1 don't draw anything.  Should see a miniature version of the screen copied into the shape.  The code for this is in Renderable.cpp Shape::Draw().

mbellew:
simple example preset (should have a patterned circle in the lower left)

[preset00]
per_frame_1000=// just a double box
per_frame_1001=//

fDecay=1
zoom=0
rot=0.0
warp=1
wave_a=0
ob_size=0.2
ob_r=0.0
ob_g=0.0
ob_b=0.0
ob_a=0.0
ib_size=0.1
ib_r=1.0
ib_g=0.0
ib_b=0.0
ib_a=1.0

// now a shape with textured=1
shapecode_0_enabled=1
shapecode_0_sides=30
shapecode_0_additive=0
shapecode_0_thickOutline=0
shapecode_0_textured=1
shapecode_0_x=0.33
shapecode_0_y=0.33
shapecode_0_rad=0.3
shapecode_0_ang=0.000000
shapecode_0_tex_ang=60
shapecode_0_tex_zoom=0.2
shapecode_0_r=1.000000
shapecode_0_g=0.000000
shapecode_0_b=0.000000
shapecode_0_a=1.000000
shapecode_0_r2=0.600000
shapecode_0_g2=0.600000
shapecode_0_b2=0.900000
shapecode_0_a2=0.000000
shapecode_0_border_r=1.000000
shapecode_0_border_g=1.000000
shapecode_0_border_b=1.000000
shapecode_0_border_a=0.000000
shape_0_per_frame1=tex_ang=sin(time)

johnzarnowski:
Hi I have a project for a software methodology class to contribute to an open source repository and I'm interested in this project. Could I learn a little bit more about this specific issue and potentially be assigned to it?

revmischa:
From what I gather here there is a preset option to mirror the screen as a texture rendered inside a shape in the preset. We're trying to make projectM as close to Milkdrop's behavior as possible, so adding support for this option would get us closer to there.
You could look at Milkdrop's source for inspiration on how it's implemented, but Milkdrop is very windows-specific and projectM uses OpenGL, so some serious translation would have to be done.
Try loading up a preset with `shapecode_0_textured=1` into Milkdrop or Butterchurn if you can to see what it looks like there.

mbellew:
[...]

PR #465, opened 2021-02-15
fix shape.textured=1

mbellew:
https://github.com/projectM-visualizer/projectm/issues/459

This seems to work, but I'm just cutting and pasting GL code.  Please review.


revmischa:
Well that was easy

revmischa (approved):

PR #467, opened 2021-02-17
Cleanup PCM interface

mbellew:
Hide member fields and create public methods.
Handle auto-leveling of input sound in PCM (instead of BeatDetect).
Implement smoothing parameter for data (used by Waveform)
Finish fixing up bass/mid/treb analysis (didn't quite get there with last attempt)
PCM unit tests


src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
+    double updateLevel(size_t samples, double sum, double max)
     {
-      PCMd[0][i]=0;
-      PCMd[1][i]=0;
+        if (sum/samples < 0.00001)
+            return level;
+        level_sum += sum;
+        level_max = fmax(level_max,max*1.02);
+        level_samples += samples;
+        if (level_samples >= 4096 || l0 <= 0)
```
revmischa, line 72: Possible to convert some of these magic numbers to constants?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
-  ip = (int *)wipemalloc(34 * sizeof(int));
-  ip[0]=0;
-
-    /** PCM data */
-//    this->maxsamples = 2048;
-//    this->numsamples = 0;
-//    this->pcmdataL = NULL;
-//    this->pcmdataR = NULL;
+    w  = (double *)wipemalloc(FFT_LENGTH*sizeof(double));
+    ip = (int *)wipemalloc(34 * sizeof(int));
```
revmischa, line 96: Magic number

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
-	free(PCMd[0]);
-	free(PCMd[1]);
-	free(PCMd);
 
+void PCM::addPCMfloat(const float *PCMdata, size_t samples)
+{
+    float a,sum=0,max=0;
+    for (size_t i=0; i<samples; i++)
+    {
+        size_t j=(i+start)%maxsamples;
```
revmischa, line 123: So if samples < maxsamples then it wraps around?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
-	free(pcmdataR);
-	free(w);
-	free(ip);
+#include <iostream>
 
-	free(PCMd[0]);
-	free(PCMd[1]);
-	free(PCMd);
 
+void PCM::addPCMfloat(const float *PCMdata, size_t samples)
```
revmischa, line 118: This is 1ch right?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
+        max = fmax(fmax(max,fabs(a)),fabs(b));
+    }
+    start = (start + samples) % maxsamples;
+    newsamples += samples;
+    level = leveler->updateLevel(samples, sum/2, max);
+}
+
 
-  for(i=0;i<samples;i++)
+void PCM::addPCM16Data(const short* pcm_data, size_t samples)
```
revmischa, line 154: 2ch?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
+}
 
-      }
-      else
-	{
-	  PCMd[0][j % maxsamples] = 0;
-	  PCMd[1][j % maxsamples] = 0;
-	}
+
+void PCM::addPCM8(const unsigned char PCMdata[2][1024])
```
revmischa, line 189: I don't like these hardcoded `1024`s

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
+    start = (start + samples) % maxsamples;
+    newsamples += samples;
+    level = leveler->updateLevel(samples, sum/2, max);
+}
 
-      if (PCMdata[i] != 0 ) {
 
-	PCMd[0][j%maxsamples] = PCMdata[i];
-	PCMd[1][j%maxsamples] = PCMdata[i];
+void PCM::addPCM16(const short PCMdata[2][512])
```
revmischa, line 171: Are we assuming there is always 2048 bytes of samples? Can there be a constant for this?
I want to experiment with fewer samples, maybe it can make the visuals more responsive?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
+        float l2 = 0, l1 =0 , c = 0, r1, r2;
+        r1 = spectrum[0]; r2 = spectrum[0+1];
+        for (size_t i = 0; i < samples; i++)
+        {
+            l2 = l1;
+            l1 = c;
+            c = r1;
+            r1 = r2;
+            r2 = (i + 2) >= samples ? 0 : spectrum[i + 2];
+            data[i] = (l2 + 4 * l1 + 6 * c + 4 * r1 + r2) / 16.0;
```
revmischa, line 317: 😕  ? comment?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
+    }
 }
 
+void PCM::_updateFFT(size_t channel)
+{
+    assert(channel == 0 || channel == 1);
+
+    double *freq = channel==0 ? freqL : freqR;
+    _copyPCM(freq, channel, FFT_LENGTH*2);
+    rdft(FFT_LENGTH*2, 1, freq, ip, w);
```
revmischa, line 338: So we do rdft on `getPCM` and here both?

src/libprojectM/PCM.cpp
```diff
@@ -24,381 +24,528 @@
...
-
-    numsamples = getPCMnew(pcmdataR,1,0,waveSmoothing,0,0);
-    getPCMnew(pcmdataL,0,0,waveSmoothing,0,1);
-    getPCM(vdataL,FFT_LENGTH,0,1,0,0);
-    getPCM(vdataR,FFT_LENGTH,1,1,0,0);
+    return a>mx ? mx : a<mn ? mn : a;
 }
 
+// pull data from circular buffer
+void PCM::_copyPCM(float *to, int channel, size_t count)
```
revmischa, line 356: I think it'd be good to keep any buffers we have as small as possible to not lag between sound and visual
What about using something like memcpy here for efficiency?

src/libprojectM/Renderer/BeatDetect.cpp
```diff
@@ -104,60 +104,36 @@ void BeatDetect::detectFromSamples()
     treb=0;
     vol=0;
 
-    // TODO: get sample rate from PCM?  Assume 44100
-    getBeatVals(44100.0f, FFT_LENGTH, pcm->pcmdataL, pcm->pcmdataR);
+    float vdataL[FFT_LENGTH];
+    float vdataR[FFT_LENGTH];
+    pcm->getSpectrum(vdataL, 0, FFT_LENGTH, 0.0);
+    pcm->getSpectrum(vdataR, 1, FFT_LENGTH, 0.0);
+    getBeatVals(44100.0f, FFT_LENGTH, vdataL, vdataR);
```
revmischa, line 111: We really shouldn't assume 44.1k here; the input sample rate can vary. At least leave the TODO here?

src/libprojectM/Renderer/MilkdropWaveform.cpp
```diff
@@ -192,10 +189,21 @@ void MilkdropWaveform::MaximizeColors(RenderContext &context)
 
 void MilkdropWaveform::WaveformMath(RenderContext &context)
 {
-	const float *pcmdataR = context.beatDetect->pcm->pcmdataR;
-	const float *pcmdataL = context.beatDetect->pcm->pcmdataL;
-	// scale PCM data based on vol_history to make it more or less independent of the application output volume
+    float pcmdataL[512];
```
revmischa, line 192: Why 512?

src/libprojectM/Renderer/Waveform.cpp
```diff
@@ -43,37 +42,38 @@ void Waveform::InitVertexAttrib() {
...
-#else
-	std::transform(&value1[0], &value1[samples_count], &value1[0], std::bind(std::multiplies<float>(), std::placeholders::_1, mult));
-	std::transform(&value2[0], &value2[samples_count], &value2[0], std::bind(std::multiplies<float>(), std::placeholders::_1, mult));
-#endif /** WIN32 */
+    float *value1 = new float[samples_count];
+    float *value2 = new float[samples_count];
+    if (spectrum)
+    {
+        // TODO support smoothing parameter for getSpectrum()
+        context.beatDetect->pcm->getSpectrum( value1, 0, samples_count, 1.0 );
```
revmischa, line 59: Doing something like
```suggestion
        context.beatDetect->pcm->getSpectrum( value1, CHANNEL_L, samples_count, 1.0 );
```
Would be more readable IMO

revmischa (approved):
This is really great, thanks for tackling this stuff. I don't really understand some of what's going on. Try to use constants instead of magic numbers. 
It looks to me like we're treating `maxsamples` as the actual number of samples in some places? I'd like to try experimenting with reducing the samples we process to see if it reduces the visual lag. Right now my biggest problem with the SDL version at least is that it really does lag the beat by a few frames.

Again great stuff 💯

src/libprojectM/PCM.cpp
mbellew: It would be great if the names were really explicit.  I just didn't want to chase down all the existing usages.
Maybe another PR could make new cleaned up names, and make them all consistent?

src/libprojectM/PCM.cpp
mbellew: The call to rdft in getPCM() is actually an inverse fft.  The call here is doing the time->freq transformation.  In getPCM() we filter the high-freq values then do the freq->time conversion.  And voila, smooth!

src/libprojectM/PCM.cpp
mbellew: If you're concerned about how the data is being provided to PCM.  That can be an issue, I have strange behavior with spotify on linux, but not with  browser streaming for instance.  I don't know what that is all about.

Once we have the data, I've never been able to show that any of the audio handling is really a problem.  The data is just so small.

src/libprojectM/Renderer/Waveform.cpp
mbellew: Yeah, I thought about changing that to an enum.

src/libprojectM/PCM.cpp
mbellew: yes

src/libprojectM/Renderer/BeatDetect.cpp
mbellew: added comment

src/libprojectM/Renderer/MilkdropWaveform.cpp
mbellew: added comment

src/libprojectM/PCM.cpp
mbellew: added comment

mbellew:
I definitely recommend pulling the branch, and comparing the test presets master vs branch.

mbellew:
Thanks for the feedback @revmischa.  I agree that the add methods could use some revisiting.  I assume they accumulated over time to match various app or sound apis.  I didn't set out to clean those up because apps directly call them.

revmischa:
Merge it whenever you deem it ready
I think next it would be great to figure out how to fix the lag, at least in the SDL version. We must have some buffer somewhere that is too large or something, because there is visible delay between the beat and the visuals

mbellew:
> Merge it whenever you deem it ready
> I think next it would be great to figure out how to fix the lag, at least in the SDL version. We must have some buffer somewhere that is too large or something, because there is visible delay between the beat and the visuals

It would be helpful to isolate how much is due to the OS sound handling vs processing in the app itself.  I kinda suspect some of this outside the app based on seeing different  quality/behavior based on the sound source.

PR #470, opened 2021-02-22
Llvm 10

mbellew:
LLVM generates a million of declaration of XYZ shadows a member of PDQ warnings
So that's not great.


src/libprojectM/MilkdropPresetFactory/Expr.cpp
```diff
@@ -1415,10 +1415,15 @@ Expr *Expr::jit(Expr *root, std::string name)
     std::vector<Type *> arg_typess;
     arg_typess.push_back(IntegerType::get(Context,32));
     arg_typess.push_back(IntegerType::get(Context,32));
-    Constant* c = jitx.module->getOrInsertFunction<Type*>("Expr_eval",
+#if LLVM_VERSION_MAJOR >= 10
```
revmischa, line 1418: I would be fine with requiring a recent version since LLVM JIT is an optional feature anyway. We need to do some work to use the newer interface though, see #360

revmischa (approved):

src/libprojectM/MilkdropPresetFactory/Expr.cpp
mbellew: That makes sense.

revmischa:
Can `configure.ac` be made to check you have the right version of LLVM?

kblaschke:
Oh, yeah, switching to C++14 is overdue!

For LLVM and CMake, it might be a good thing to check for LLVM being at least version 10 if the user has set ENABLE_LLVM and fail if not. Otherwise, it might leave the user thinking that LLVM support is enabled, while it actually isn't. Just add this after [the find_package(LLVM)](https://github.com/projectM-visualizer/projectm/blob/b2b12a00f06103882c37e130b0d3cb05ffdb6476/CMakeLists.txt#L105)  command:

```CMake
    if(LLVM_VERSION VERSION_LESS 10.0)
        message(FATAL_ERROR "LLVM JIT support requires at least version 10.0, but only ${LLVM_VERSION} was found.")
    endif()
```

mbellew:
> Oh, yeah, switching to C++14 is overdue!
> 
> For LLVM and CMake, it might be a good thing to check for LLVM being at least version 10 if the user has set ENABLE_LLVM and fail if not. Otherwise, it might leave the user thinking that LLVM support is enabled, while it actually isn't. Just add this after [the find_package(LLVM)](https://github.com/projectM-visualizer/projectm/blob/b2b12a00f06103882c37e130b0d3cb05ffdb6476/CMakeLists.txt#L105) command:
> 
> ```cmake
>     if(LLVM_VERSION VERSION_LESS 10.0)
>         message(FATAL_ERROR "LLVM JIT support requires at least version 10.0, but only ${LLVM_VERSION} was found.")
>     endif()
> ```

thanks

kblaschke (approved):

revmischa:
🎉 !!

PR #483, opened 2021-04-07
Added case-insensitive search

ZimM-LostPolygon:
Added case-insensitive search to the `M`-key preset explorer. Now you don't have to memorize the exact case of your favorite presets!

Also, fixed a typo in `PCM.cpp` that resulted in Debug builds failing under MSVC.


revmischa:
Thanks!

PR #487, opened 2021-04-21
Add CMake build system files (only as "autotools" replacement for now)

kblaschke:
Added CMake scripts for all targets that are currently being built by the autotools scripts, recreating the original install structure as closely as possible. Vendor library detections either use CMake-provided find scripts or, if possible, `pkgconfig` just as the autotools scripts. See the `cmake` directory for how it's done.

I've tried not to touch any existing files. Only two changes were made, one of those not affecting the autotools build as the executable wasn't even built:

- Changed a few lines of Qt code that initializes the resources for the Qt UI applications. CMake generates resources with the `-name <QRC basename>` switch, which will add a suffix to the `qInitResources()` function. I've replaced the direct call with the proper macro and added the `-name` switch to the original Makefile.
- A larger change was a quick & dirty upgrade of the SDL-based JACK UI from SDL1 to SDL2. The application now runs with the most recent SDL version and is built by the CMake build system if both `ENABLE_JACK` and `ENABLE_SDL` are set to `ON`.

CMake will display a warning message that informs developers not to use CMake files for production purposes yet. Will remove this message when it works on all supported platforms and the build parameters are no longer changing.

I've tested building on a Gentoo Linux machine. Should work out-of-the-box on other distros as well, but this needs to be tested.

These autotools-supported features not yet fully converted:

- Using GLES
- Building for Emscripten (depends on GLES support)


CMakeLists.txt
```diff
@@ -0,0 +1,182 @@
...
+option(ENABLE_EMSCRIPTEN "Build for web with emscripten" OFF)
+cmake_dependent_option(ENABLE_SDL "Enable SDL2 support" ON "NOT ENABLE_EMSCRIPTEN;ENABLE_TESTING" ON)
+cmake_dependent_option(ENABLE_GLES "Enable OpenGL ES support" OFF "NOT ENABLE_EMSCRIPTEN" ON)
+cmake_dependent_option(ENABLE_THREADING "Enable multithreading support" ON "NOT ENABLE_EMSCRIPTEN" OFF)
+cmake_dependent_option(ENABLE_PULSEAUDIO "Build Pulseaudio-based Qt UI" OFF "ENABLE_QT;NOT ENABLE_EMSCRIPTEN;CMAKE_SYSTEM_NAME STREQUAL Linux" OFF)
+cmake_dependent_option(ENABLE_JACK "Build JACK-based Qt and SDL UIs" OFF "ENABLE_QT;NOT ENABLE_EMSCRIPTEN;CMAKE_SYSTEM_NAME STREQUAL Linux" OFF)
+cmake_dependent_option(ENABLE_LLVM "Enable LLVM JIT support" OFF "NOT ENABLE_EMSCRIPTEN" OFF)
+cmake_dependent_option(ENABLE_LIBVISUAL "Build and install the projectM libvisual plug-in" OFF "NOT ENABLE_EMSCRIPTEN;CMAKE_SYSTEM_NAME STREQUAL Linux" OFF)
+
+option(USE_SYSTEM_GLM OFF "Use system/external GLM library instead of the bundled one")
```
revmischa, line 56: Would be cool if we could detect and use system glm if available

presets/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+            presets_eyetune
+            presets_milkdrop
+            presets_milkdrop_104
+            presets_milkdrop_200
+            presets_mischa_collection
+            presets_projectM
+            presets_stock
+            presets_tryptonaut
+            presets_yin
+            tests
```
revmischa, line 14:
```suggestion
```

src/NativePresets/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+            COMPONENT Presets
+            )
+endmacro()
+
+add_native_preset(MstressJuppyDancer)
+add_native_preset(RLGFractalDrop7c)
+add_native_preset(RovastarDarkSecret)
+add_native_preset(RovastarDriftingChaos)
+add_native_preset(RovastarFractalSpiral)
+add_native_preset(RovastarFractopiaFrantic)
```
revmischa, line 40: Would be nice sometime to have an enable flag for native presets, I don't think most people need or want them

src/libprojectM/CMakeLists.txt
```diff
@@ -0,0 +1,245 @@
...
+# This syntax will pull in the compiled object files into the final library.
+add_library(projectM_static STATIC
+        $<TARGET_OBJECTS:MilkdropPresetFactory>
+        $<TARGET_OBJECTS:NativePresetFactory>
+        $<TARGET_OBJECTS:Renderer>
+        $<TARGET_OBJECTS:hlslparser>
+        $<TARGET_OBJECTS:SOIL2>
+        Common.hpp
+        ConfigFile.cpp
+        ConfigFile.h
```
revmischa, line 20: Does adding header files do anything?

src/libprojectM/CMakeLists.txt
```diff
@@ -0,0 +1,245 @@
...
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/Renderer/hlslparser/src>"
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/MilkdropPresetFactory>"
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/NativePresetFactory>"
+        "$<INSTALL_INTERFACE:${PROJECTM_INCLUDE_DIR}>"
+        )
+
+target_link_libraries(projectM_shared
+        PUBLIC
+        OpenGL::OpenGL
+        OpenGL::GLX
```
revmischa, line 122: What's GLX for? Do we use it?

src/libprojectM/CMakeLists.txt
```diff
@@ -0,0 +1,245 @@
...
+# pkgconfig export
+macro(set_pkg_config_path varname path)
+    if(IS_ABSOLUTE "${path}")
+        set(${varname} "${path}")
+    else()
+        set(${varname} "\${prefix}/${path}")
+    endif()
+endmacro()
+
+set(PKGCONFIG_PREFIX "${CMAKE_INSTALL_PREFIX}")
```
revmischa, line 215: Is it possible to just write
```suggestion
set(PKGCONFIG_PREFIX CMAKE_INSTALL_PREFIX)
```
?

revmischa (approved):
Hey this is terrific, thanks!

CMakeLists.txt
kblaschke: Sure, could just call ‘find_package(GLM)‘ without ‘REQUIRED‘ and then check the ‘GLM_FOUND‘ variable. If it's set to ‘NOTFOUND‘, use the bundled library and output a message.

presets/CMakeLists.txt
kblaschke: Good one. They were installed with the autotools scripts, so wasn't sure if that was intentional. Be gone, then!

src/NativePresets/CMakeLists.txt
kblaschke: Can just add an ‘ENABLE_NATIVE_PRESETS‘ build option, defaulting to ‘OFF‘. Then the user can enable them manually if really desired.

src/libprojectM/CMakeLists.txt
kblaschke: Not required for compilation, but adds them to the "rebuild target and its dependents if these files have changed" list and also displays the files in IDE projects, e.g. Xcode, KDevelop or Visual Studio so one can easily open headers from the project tree. Also makes sure that the configuration phase fails if the file is not present for some reason.

src/libprojectM/CMakeLists.txt
kblaschke: SOIL2 automatically uses it if compiled on an X server platform, so at least on most Linux systems it's required. If I've seen it correctly, linking this target on other platforms is a simple no-op, but have to check that.

src/libprojectM/CMakeLists.txt
kblaschke: Sadly no, only ‘if()‘ interprets variables directly. The ‘set()‘ command only takes the first argument as a variable name (also literally as a string) and then assigns everything after that as a string. The quotes are also required, otherwise spaces in a path can end up as a list separator, as CMake interprets unquoted spaces as such. So using

‘set(SOME_PATH /path/to/dir with/space)‘

will be stored as ‘/path/to/dir;with/space‘ in the variable. This variable weirdness is one of the things I despise most in CMake, especially the scoping mechanisms, but you get used to it 😁

kblaschke:
Another thing that still seems to be broken as I found out while testing is threading support. While threads are used, these seems to be some kind of use-after-free happening in the Renderer class, segfaulting the application. Probably missed another #define required for proper locking, will check the autotools-generated config.h for candidates.

If using recent C++ standards is an option, it'd be a good thing to port the threading support to std::thread instead of pthread (in a future change).

src/libprojectM/CMakeLists.txt
revmischa: Getting this on macOS
```
-- Configuring done
CMake Error at src/libprojectM/CMakeLists.txt:12 (add_library):
  Target "projectM_static" links to target "OpenGL::OpenGL" but the target
  was not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


CMake Error at src/libprojectM/CMakeLists.txt:12 (add_library):
  Target "projectM_static" links to target "OpenGL::GLX" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

revmischa:
Yes we can use any recent C++ standard, I don't see any issue with that. I think there are a lot of modern C++ features we could be taking advantage of

src/libprojectM/CMakeLists.txt
kblaschke: Probably a similar situation in Windows, didn't implement any specifics for those platforms yet (OpenGL in latest macOS versions is also kinda being deprecated by Apple in favor of Metal). Should compile on any recent Linux distribution though.

kblaschke:
Pushed a few additional commits, fixing some of the suggestions above.

src/libprojectM/CMakeLists.txt
kblaschke: Linking to `OpenGL::GL` instead and only adding `OpenGL::GLX` to the list if it's available (and in turn, SOIL2 requires it). Should resolve this configuration issue on MacOS (if the OpenGL.framework is available in the Xcode SDK).

presets/CMakeLists.txt
kblaschke: Changed in commit cc785a6, now only installing these presets if ENABLE_TESTING is set to a true'ish value.

CMakeLists.txt
kblaschke: Fixed in commit aa34bd5

revmischa:
Ok it cmade this time, compiling fails unable to find glm
```
/Users/cyber/dev/projectM/src/libprojectM/MilkdropPresetFactory/../Renderer/Renderable.hpp:7:10: fatal error: 'glm/mat4x4.hpp' file not found
#include <glm/mat4x4.hpp>
         ^~~~~~~~~~~~~~~~
```

```
-- Could NOT find GLM (missing: GLM_INCLUDE_FILE GLM_INCLUDE_DIR)
-- GLM library not found, using bundled version.
-- Looking for aligned_alloc
-- Looking for aligned_alloc - found
-- Looking for C++ include dlfcn.h
-- Looking for C++ include dlfcn.h - found
-- Looking for C++ include fts.h
-- Looking for C++ include fts.h - found
-- Looking for C++ include GL/gl.h
-- Looking for C++ include GL/gl.h - not found
-- Looking for C++ include inttypes.h
-- Looking for C++ include inttypes.h - found
-- Looking for C++ include memory.h
-- Looking for C++ include memory.h - found
-- Looking for posix_memalign
-- Looking for posix_memalign - found
-- Looking for enum value PTHREAD_PRIO_INHERIT
-- Looking for enum value PTHREAD_PRIO_INHERIT - found
-- Looking for C++ include stdint.h
-- Looking for C++ include stdint.h - found
-- Looking for C++ include stdlib.h
-- Looking for C++ include stdlib.h - found
-- Looking for C++ include strings.h
-- Looking for C++ include strings.h - found
-- Looking for C++ include string.h
-- Looking for C++ include string.h - found
-- Looking for C++ include sys/stat.h
-- Looking for C++ include sys/stat.h - found
-- Looking for C++ include sys/types.h
-- Looking for C++ include sys/types.h - found
-- Looking for C++ include unistd.h
-- Looking for C++ include unistd.h - found
-- Looking for C++ include windows.h
-- Looking for C++ include windows.h - not found
-- Looking for 23 include files assert.h, ..., tgmath.h
-- Looking for 23 include files assert.h, ..., tgmath.h - found
--
-- projectM v3.1.13
--     =====
--
--     prefix:                  /usr/local
--     libdir:                  lib
--     includedir:              include
--     bindir:                  bin
--     libvisual plugin dir:
--
--     compiler:                /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
--     cflags:
--     cxxflags:
--     ldflags:
--
--     DATADIR_PATH:            share/projectM
--     - - -
--
--     Applications:
--     =====
--
--     libprojectM:             ON
--     Presets:                 ON
--     Native presets:          OFF
--     Tests:                   OFF
--     Threading:               ON
--     SDL2:                    ON
--     SDL2 version:            2.0.15
--     Pulseaudio UI:           OFF
--     JACK UI:                 OFF
--     OpenGLES:                OFF
--     Emscripten:              OFF
--     LLVM JIT:                OFF
--     libvisual plug-in:       OFF
--     Use system GLM:          ON
```

kblaschke:
Found another typo regarding LLVM linkage that creeped in with the second-to-last commit, so amended the respective commits to contain the LLVM and GLM fixes.

revmischa:
```
-- Could NOT find GLM (missing: GLM_INCLUDE_FILE GLM_INCLUDE_DIR)
-- GLM library not found, using bundled version.
--
-- projectM v3.1.13
--     =====
--
--     prefix:                  /usr/local
--     libdir:                  lib
--     includedir:              include
--     bindir:                  bin
--     libvisual plugin dir:
--
--     compiler:                /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
--     cflags:
--     cxxflags:
--     ldflags:
--
--     DATADIR_PATH:            share/projectM
--     - - -
--
--     Applications:
--     =====
--
--     libprojectM:             ON
--     Presets:                 ON
--     Native presets:          OFF
--     Tests:                   OFF
--     Threading:               ON
--     SDL2:                    ON
--     SDL2 version:            2.0.15
--     Pulseaudio UI:           OFF
--     JACK UI:                 OFF
--     OpenGLES:                OFF
--     Emscripten:              OFF
--     LLVM JIT:                OFF
--     libvisual plug-in:       OFF
--     Use system GLM:          OFF
```

But still
```
/Users/cyber/dev/projectM/src/libprojectM/Renderer/Renderable.hpp:7:10: fatal error: 'glm/mat4x4.hpp' file not found
```

kblaschke:
Need to move the bundled GLEW files into a "GL" subdir, this is what makes the build fail. Change was required as 3rd party GLEW headers need to be included with `#include <GL/glew.h>`.

src/libprojectM/omptl/omptl_numeric_extentions_par.h
```diff
@@ -123,7 +123,7 @@ struct _TransformAccumulate< ::std::input_iterator_tag >
 template <class Iterator, class T, class UnaryFunction, class BinaryFunction>
 T transform_accumulate(Iterator first, Iterator last, const T init,
 			UnaryFunction unary_op, BinaryFunction binary_op,
-			const unsigned P = omp_get_max_threads())
```
kblaschke, line 126: GCC only likes one template instance having a default value, not the overrides.

BUILDING.md
```diff
@@ -1,56 +1,294 @@
...
-|-----------------------|----------------------------------------------------------------------------------------|-----------------------|
-| `--enable-sdl`        | `libsdl2-dev`                                                                          | `projectMSDL`         |
-| `--enable-pulseaudio` | `qt5-default` `qtdeclarative5-dev` `libpulse-dev` `libqt5opengl5-dev`                  | `projectM-pulseaudio` |
-| `--enable-jack`       | `libjack2-dev`OR`libjack1-dev`; `qt5-default` `qtdeclarative5-dev` `libqt5opengl5-dev` | `projectM-jack`       |
+* A working build toolchain.
+* **OpenGL**: 3D graphics library. Used to render the visualizations.
+* **GLES3**: OpenGL libraries for embedded systems, version 3. Required to build projectM on mobile devices, Raspberry
+  Pi and Emscripten.
+* [**glm**](https://github.com/g-truc/glm):  OpenGL Mathematics library. Optional, will use a bundled version with
+  autotools or if not installed.
```
revmischa, line 15: with autotools? not cmake?

BUILDING.md
```diff
@@ -1,56 +1,294 @@
...
+  pre-created Visual Studio solutions.
+* [**GLEW**](http://glew.sourceforge.net/): The OpenGL Extension Wrangler Library. Only required if using CMake to
+  configure the build, the pre-created solutions use a bundled copy of GLEW.
+
+## Building on Linux and macOS
+
+### Installing dependencies
+
+- Linux distributions will have packages available for most (if not all) required libraries. The package names and
+  commands to install them vary widely between distributions (and even versions of the same distribution). Please refer
```
revmischa, line 58: We had the list of debian packages above, would be nice to keep that since most people use that or know how to "translate" them to their weird pervert package managers

BUILDING.md
kblaschke: Yes, the bundled version is always used with autotools as far as I can tell. CMake will prefer the system headers if found and fall back to the bundled headers if they aren't there. It's just a header-only library and no new versions since a long time, so there's no real difference using either.

BUILDING.md
kblaschke: Would then add a list of suggested packages for some distros, Homebrew and vcpkg, with the latter two also being an option on Linux as well.

revmischa: I think your documentation is swell and terrific and informative, just that 90% of users are just going to want to know "wtf do I type in to make projectM run" so there should be some really short quickstart section

CMakeLists.txt
```diff
@@ -0,0 +1,193 @@
...
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+
+set(PROJECTM_BIN_DIR "bin" CACHE STRING "Executable installation directory, relative to the install prefix.")
+set(PROJECTM_DATADIR_PATH "share/projectM" CACHE STRING "Default (absolute) path to the projectM data files (presets etc.)")
+set(PROJECTM_LIB_DIR "lib" CACHE STRING "Library installation directory, relative to the install prefix.")
+set(PROJECTM_INCLUDE_DIR "include" CACHE STRING "Header installation directory, relative to the install prefix.")
+
+# Feature options, including dependencies.
+option(ENABLE_PRESETS "Build and install bundled presets" ON)
```
Be-ing, line 34:
```suggestion
option(PRESETS "Build and install bundled presets" ON)
```

This `ENABLE_` prefix for options is unconventional.

CMakeLists.txt
```diff
@@ -0,0 +1,193 @@
+cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
```
Be-ing, line 1: Is such a recent version of CMake really needed?

src/NativePresets/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
+if(NOT ENABLE_PRESETS OR NOT ENABLE_NATIVE_PRESETS)
+    return()
+endif()
+
+# Using a macro to reduce repetition.
+macro(add_native_preset name)
+    if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
+        add_library(${name}_static STATIC ${name}.cpp)
```
Be-ing, line 8: Why are you forcing static linking instead of letting the user set [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)?

src/NativePresets/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+    
+        install(TARGETS ${name}_static
+                LIBRARY DESTINATION "${PROJECTM_DATADIR_PATH}/presets"
+                RUNTIME DESTINATION "${PROJECTM_DATADIR_PATH}/presets"
+                ARCHIVE DESTINATION "${PROJECTM_DATADIR_PATH}/presets"
+                COMPONENT Presets
+                )
+    endif()
+
+    add_library(${name}_shared SHARED ${name}.cpp)
```
Be-ing, line 28: This is unconventional to have separate targets for static and shared libraries. Please use one target and let the user choose how to link with BUILD_SHARED_LIBS.

src/NativePresets/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+    target_compile_definitions(${name}_shared
+            PRIVATE
+            ${name}_EXPORTS
+            )
+
+    set_target_properties(${name}_shared PROPERTIES
+            OUTPUT_NAME ${name}
+            )
+
+    # On Windows, libprojectM is only built statically.
```
Be-ing, line 39: [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) can help with this.

Be-ing:
Thanks for working on this! I'd love to get libprojectM packaged in vcpkg after this is merged and released.

Could you add a GitHub Actions workflow file to test this on CI for Linux, Windows, and macOS? Take a look at the GH Actions workflows I wrote for [libkeyfinder](https://github.com/mixxxdj/libkeyfinder/blob/main/.github/workflows/build-and-test.yml) and [PortAudio](https://github.com/Be-ing/portaudio/blob/cmake_rewrite/.github/workflows/cmake.yml) for examples.

src/NativePresets/CMakeLists.txt
kblaschke: Totally agree.

Reason for this is: that is just the first step in adding CMake scripts, so the initial goal was to recreate the same build output/file layout as the autoconf-based builds, which actually builds both libraries (even the native presets) and don't has an option to control that.

This makes it easier for now to find any issues left from porting the build. After that, it's just a few small changes to use these CMake-standard schemes, and will definitely be done as in most use cases, only the shared libraries will be required.

src/NativePresets/CMakeLists.txt
kblaschke: Kind of a brute-force solution though. As of now, the library has no really well-defined export interface, the installed headers are not exporting all functionality the library could supply and having all symbols exported will both increase the symbol table size and make it harder to later clean up the API or change any code if people start using all the internal classes.

This is a bit out of scope of simply adding a new build system and will require some work on the codebase. I kept it the way the original build files (the VS solution in this case) handled the build.

src/NativePresets/CMakeLists.txt
Be-ing: Okay. That seems like a good plan to me. :)

src/NativePresets/CMakeLists.txt
Be-ing: If you want to put in the work to use Microsoft's nonstandard macros, sure go ahead. Personally I like CMake because it requires little work to support operating systems I don't use myself.

src/NativePresets/CMakeLists.txt
kblaschke: As stated in my other comment, the code itself is currently not ready to support proper DLL building on Windows as it's missing dllexport attributes and proper interface classes.

If you have interest in building a proper external API that hides implementation details but exports all required interfaces in a well-declared way, we can do this together with Mischa on Discord. Every input on that counts.

src/NativePresets/CMakeLists.txt
kblaschke: I'd use CMake's [export header generator](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) module that creates a header with wrapper macros tailored to the compiler being used. For example this will use `declspec _dllexport` on MSVC and `visibility(visible)` and setting default symbol visibility to hidden on UNIX compilers like GCC and Clang. Really neat.

CMakeLists.txt
kblaschke: Checked the reason, it's because I was using the [CheckLinkerFlag](https://cmake.org/cmake/help/v3.19/module/CheckLinkerFlag.html) and [CheckCompilerFlag](https://cmake.org/cmake/help/v3.19/module/CheckCompilerFlag.html) modules, which have been introduced in CMake 3.18 and 3.19, respectively.

I can resort to the older CheckCXXCompilerFlag and CheckCXXSourceCompiles modules with a bit more boilerplate code around them.

The oldest CMake version that can be supported without sacrificing more functionality - specifically the `add_link_options` command - is 3.14, released in February 2019.

As it's quite easy to install any recent CMake version and older versions have other bugs and issues,

kblaschke:
Using GitHub Actions would probably a good idea, but this is really up to @revmischa to decide, as you only get 2000 Action minutes per month for free and then have to pay for it.

revmischa:
I wholeheartedly endorse switching to GHA!!

Be-ing:
> Using GitHub Actions would probably be a good idea, but this is really up to @revmischa to decide, as you only get 2000 Action minutes per month for free and then have to pay for it.

That is only for private repositories. Public repositories have unlimited run time. https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

That said, setting up compiler caching is a good idea. ccache works with GCC and Clang. AFAIK [sccache](https://github.com/mozilla/sccache) is the only currently maintained compiler cache that works with MSVC.

CMakeLists.txt
Be-ing: Ubuntu 20.04 [has cmake 3.16](https://packages.ubuntu.com/focal/cmake) so I think that would be a good target.

Be-ing:
@kblaschke you seem to know what you're doing here with CMake. I have a [pending PR](https://github.com/PortAudio/portaudio/pull/461) rewriting PortAudio's CMake. If you could review it, I'd appreciate that :)

revmischa:
> @kblaschke you seem to know what you're doing here with CMake. I have a [pending PR](https://github.com/PortAudio/portaudio/pull/461) rewriting PortAudio's CMake build system. If you could review it, I'd appreciate that :)

This is a totally separate topic but I think we need to integrate PortAudio with ProjectM so we can get more control over low-latency audio capture and input monitoring.

kblaschke:
If there are no big issues left with this initial set of CMake files, I'd say it's ready for merging. This will make the new build system available for more users to test, giving us more data about more fixes and build problems.

PR #488, opened 2021-05-02
Made it possible to do decimal duration times between transitions.

hibengler:
…Duration

to be Double precision numbers instead of /as well as integers.
This will allow more complex transitions, such as transitions on measure
detect or 4 beat detects for example.
Forsing this to generte youtube videos, it would be good for the transitions to happen precisely when measures or stances in the music change.


revmischa (approved):
Sounds cool to me although I think a float would have sufficed :)

hibengler:
I thought so too, but the internals in TimeKeeper are all double, so it 
is consistent.

PR #502, opened 2021-06-09
Fixed a few memory management issues with presets.

kblaschke:
Mainly two problems:
- MilkdropPresetFactory deleted a private reference-typed member in MilkdropPreset. Changed this into a plain pointer which is now checked and properly cleaned up.
- In the library interface class, projectM, the active preset pointers were not deleted before the factors, which lead to a use-after-free when the presets tried to dereference the factory pointer.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,65 +1,66 @@
----
-Language:        Cpp
-# BasedOnStyle:  LLVM
-AccessModifierOffset: -2
-AlignAfterOpenBracket: false
-AlignEscapedNewlinesLeft: true
-AlignOperands:   true
-AlignTrailingComments: true
-AllowAllParametersOfDeclarationOnNextLine: true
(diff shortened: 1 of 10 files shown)


revmischa (commented):
Thanks for cleaning up the files but probably better to do that in a separate commit

kblaschke:
> Thanks for cleaning up the files but probably better to do that in a separate commit

You're right, did that - new first commit only contains clang-format changes. Makes the fix clearly visible now.

kblaschke:
Just one general thing, please also review large code format changes thoroughly as this is sometimes used to introduce malicious code unnoticed. An easy way to do this is applying clang-format to the parent commit (using the new .clang-format file) and compare the results against the first commit. They should be identical.

revmischa (approved):

revmischa:
Want to merge?

PR #503, opened 2021-06-09
Add GitHub Action to build with CMake on Windows, Linux and macOS.

kblaschke:
Just a workflow file plus some CMake scripts for each step that do the actual work.

The workflow builds projectM on all three major desktop platforms using the public GitHub runners and packages the results. Currently, the archives will not be unpack-and-run, as the structure is currently fixed to the Linux install layout. It's mainly there to check for build failures on one or more platforms using the CMake scripts.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -1,0 +1,37 @@
+name: CMake Build
+
+on: [ push, pull_request ]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ windows-latest, ubuntu-latest, macos-latest ]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Prerequisites
+        run: cmake -P ${{ github.workspace }}/.github/workflows/install_prerequisites.cmake
+
+      - name: Configure CMake
+        run: cmake -P ${{ github.workspace }}/.github/workflows/configure.cmake
+
+      - name: Build
+        run: cmake -P ${{ github.workspace }}/.github/workflows/build.cmake
+
+      - name: Package
+        run: cmake -P ${{ github.workspace }}/.github/workflows/package.cmake
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.os }}
+          path: package/projectM-*
--- a/.github/workflows/install_prerequisites.cmake
+++ b/.github/workflows/install_prerequisites.cmake
@@ -1,0 +1,47 @@
+message(STATUS "Using host CMake version: ${CMAKE_VERSION}")
+
+if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
+    # On Windows, using vcpkg to install and build is the best practice.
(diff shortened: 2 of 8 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,37 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ windows-latest, ubuntu-latest, macos-latest ]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Prerequisites
+        run: cmake -P ${{ github.workspace }}/.github/workflows/install_prerequisites.cmake
```
revmischa, line 22: Can't you simply do
```suggestion
        run: cmake -P .github/workflows/install_prerequisites.cmake
```
?

.github/workflows/build_cmake.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: cmake -P ${{ github.workspace }}/.github/workflows/build.cmake
+
+      - name: Package
+        run: cmake -P ${{ github.workspace }}/.github/workflows/package.cmake
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.os }}
+          path: package/projectM-*
```
revmischa, line 37: Will this use a lot of space?

.github/workflows/install_prerequisites.cmake
```diff
@@ -0,0 +1,47 @@
+message(STATUS "Using host CMake version: ${CMAKE_VERSION}")
+
+if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Windows")
```
revmischa, line 3: This stuff would probably be easier to do in the GHA workflow file

revmischa (approved):
Awesome thanks
Would be happy to retire travis-ci soon
I think it's easier to write these scripts in the GHA action files rather than in CMake

.github/workflows/build_cmake.yml
kblaschke: Probably works, though my experience with CI runners is that you should never, ever rely on the working dir being set to something specific, as some runners will not reset it after the previous action. It's simply more safe to always use absolute paths.

.github/workflows/install_prerequisites.cmake
kblaschke: You can only run scripts via the YAML file AFAIK, but not really do different things in the run command if using a matrix configuration. With scripts, you'd need to use bash on Linux/Mac and PowerShell for Windows. CMake is platform-agnostic and has a shell/script mode I'm using here. It's different from the standard configure/build usage as it only runs a CMake script without configuring any build system etc. [See the CMake docs](https://cmake.org/cmake/help/latest/manual/cmake.1.html#run-a-script) for more on that topic.

Another solution would be having one workflow per platform. In this case might not be so different as there's not much commonality in the configuration/install calls due to platform specifics.

.github/workflows/build_cmake.yml
kblaschke: Just under 30 MB per run currently.

AFAIK GitHub Action minutes and storage are [free for public repos](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions) anyways and old builds should be deleted after some time, depending on the project configuration.

kblaschke:
If the GLESv3 support is working properly, I'll add an Android NDK action to ensure that it also builds and links. The GitHub runners have the required tools already installed.

revmischa:
Cool, this is terrific thanks

PR #510, opened 2021-07-02
Small updates to check if files exist and better handle edge cases/vulnerabilities.

rachpchan:
Note: Line 121 if( hModule == NULLPTR ) should actually stay as NULL, not NULLPTR. Ignore the comment.

Hi, I'm contributing to this open-source project as part of a class assignment. I'm a beginning programmer and pursuing a concentration in cybersecurity, and made small changes to better check if objects actually exist.

Mostly, the if statements would just check that a variable exists, but I thought that this could lead to some potential exploits in the future. For the most part, I thought it would be safer to whitelist very specific inputs and conditions (which is why I decided to change the if statements to check for NULLPTR to see if they exist or not), than to just do if(err) or such... as those conditions can be a bit ambiguous. 

Hope these changes make sense, and provide a bit of help in handling vulnerabilities and error cases. Please review and merge/reject my pull request accordingly. Thanks!

### DIFF
--- a/msvc/dirent.h
+++ b/msvc/dirent.h
@@ -365,7 +365,8 @@ _wopendir(
 
     /* Allocate new _WDIR structure */
     dirp = (_WDIR*) malloc (sizeof (struct _WDIR));
-    if (!dirp) {
+    if (dirp == NULLPTR) { // Probably better to check that dirp exists by using dirp == NULLPTR to avoid vulnerability issues
+                           // as !dirp does/will not any whitelisted inputs this way.
         return NULL;
     }
 
@@ -431,7 +432,7 @@ _wopendir(
     *p = '\0';
 
     /* Open directory stream and retrieve the first entry */
-    if (!dirent_first (dirp)) {
+    if (dirent_first (dirp) == NULLPTR) { // See above.
         goto exit_closedir;
     }
 
@@ -482,7 +483,7 @@ _wreaddir_r(
 
     /* Read next directory entry */
     datap = dirent_next (dirp);
-    if (datap) {
+    if (datap != NULLPTR) { // See above.
         size_t n;
         DWORD attr;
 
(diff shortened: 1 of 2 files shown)


revmischa:
Thanks for choosing to contribute! 
Can you please give an example of how the code could be exploited before your change?
What is the functional difference between `!foo` and `foo != NULLPTR`? Are they not functional equivalent?

kblaschke:
Since you're stating that you are a beginner, and you seemingly don't have a lot of experience with C/C++, let me explain a few things which might not be obvious if you're mostly used to scripting languages like JavaScript.

In C/C++. using `if(!pointerVar)` doesn't check for existence of a variable, but for the actual contents being zero or nonzero. It's a strongly typed language which means a variable _must_ be declared to a single specific type. Pointers may point to object instances, but internally they are only unsigned integers with a size matching the architecture (32 or 64 bits). In C, invalid or non-assigned pointers are set to zero, because memory address zero is always considered invalid.

The `malloc` C library function will return zero if it can't allocate the requested memory for some reason. Any other non-zero return value means the allocation was successful. Also note that in this case, the variable (`dirp`) does _not_ contain any valid object - at this point, it only has a memory area reserved, with random contents. In C, you need to initialize the memory by either filling it with zeroes (works only in simple cases) or once writing each member of the structure that you want to put there. C++ has the `new` keyword for classes and structs that also initializes the memory after allocation.

The negation operation simply flips all bits in the value to the opoosite. The `if` statement interprets everything that is non-zero (e.g. at least one bit in the value being 1) as `true`, and only values which have all bits set to zero as `false`.

This means that using `if(!pointerVar)` to check if a variable contains a (possibly valid) pointer is totally fine, and even the most readable way of doing so. If a pointer variable is set to an invalid memory address, there is _no way_ of checking it, so programmers need to be very careful to keep pointer addresses valid.

As a closing note, using `NULLPTR` in uppercase letters is totally non-standard. You either use literal `0`, the C-style `NULL` macro in C code, or `nullptr` in lowercase in modern C++ code.

I'll close this pull request without merging it because the changes aren't required to improve security or fix any real issue. Looking through the functions, all checks are good and there's no real way a malicious user could exploit something here.

If you think closing this PR is wrong and there is a real issue, please reopen the PR and provide us with a working exploit that shows any existing security issue with the code which is fixed by your changes.

PR #512, opened 2021-07-08
Windows fix for SDL DPI scaling

milkdropper:
I recall there was a bug for this and we found a trick by right clicking the .exe properties and changing a setting.

I purchase a high resolution monitor and experienced the problem myself for the first time. Windows by default will offer scaling to make apps easier to read at large resolutions.
image

So projectM will virtually scale. 200% means your resolution is cut in half.
image

This minor fix makes projectM ignore scaling so you get true resolution:
image

### DIFF
--- a/msvc/projectMSDL.vcxproj
+++ b/msvc/projectMSDL.vcxproj
@@ -94,6 +94,9 @@
     <ProjectReference>
       <LinkLibraryDependencies>false</LinkLibraryDependencies>
     </ProjectReference>
+    <Manifest>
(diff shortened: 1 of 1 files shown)


kblaschke:
The .vxproj file will soon be removed in favor of the CMake build system. You could add the CMake target property `VS_DPI_AWARE` with a value of `PerMonitor` to the projectMSDL executable target, but remember to check the CMake version to be at least 3.16, earlier versions may complain about an unknown property.

kblaschke:
Also, we can ignore the macOS build failure. Homebrew sometimes fails to install a package.

revmischa:
Thanks! Had to do something on macOS as well

Issue #429, opened 2020-11-01
Steam ProjectM issue


orien3243:
So ProjectM on steam makes projectM really easy to install even on Linux. The problem is it won't sync up these visuals to audio playing on my computer or even via a mic. I left a comment in the steam community page and its gone unanswered since August. So I figured I would post it here just in case. 

Pulse audio also has this nice feature in where I can also adjust the volume of running applications such as chrome or my audio player and set the audio volume there. However pulse audio does not even see the projectm application from steam. (This is odd because the compiled version of projectm on my *buntu installation shows up pulse audio's mixer)

How to reproduce the issue?

0. use Ubuntu (or any variant such as Kubuntu or Lubuntu in my case with pulseaudio)

1. install steam and projectm

2. launch projectm

3. "no input devices found" 

4. press Ctrl+I 

5. Crash

Terminal output

`joedoe47@greatfox2:~/.local/share/Steam/steamapps/common/projectM$ ./projectM 
INFO: GL_VERSION: 3.2.0 NVIDIA 435.21
INFO: GL_SHADING_LANGUAGE_VERSION: 1.50 NVIDIA via Cg compiler
INFO: GL_VENDOR: NVIDIA Corporation
INFO: Using data directory: /usr/local/share/projectM
INFO: created ~/.projectM/config.inp successfully
[projectM] config file: /home/joedoe47/.projectM/config.inp
INFO: Using config from /home/joedoe47/.projectM/config.inp
CRITICAL: No audio capture devices found
fps: 29

joedoe47@greatfox2:~/.local/share/Steam/steamapps/common/projectM$ ./projectM --help
INFO: GL_VERSION: 3.2.0 NVIDIA 435.21
INFO: GL_SHADING_LANGUAGE_VERSION: 1.50 NVIDIA via Cg compiler
INFO: GL_VENDOR: NVIDIA Corporation
INFO: Using data directory: /usr/local/share/projectM
INFO: created ~/.projectM/config.inp successfully
[projectM] config file: /home/joedoe47/.projectM/config.inp
INFO: Using config from /home/joedoe47/.projectM/config.inp
CRITICAL: No audio capture devices found
fps: 29
INFO: Displaying preset: Idiot - MOTIVATION!.milk
INFO: Opened audio capture device index=0 devId=2: (null)
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
`

Hopefully I am just missing something basic here.

[...]

Issue #30, opened 2018-02-13
Audio input device selection


revmischa:
In projectM-sdl there needs to be a way to select which input device you want to use.

tahirdon:
Hey, am a new guy, never made any PR, can you assist me.

tahirdon:
Or provide me direction. Thanks

revmischa:
Sure... you can come chat in #projectM on freenode IRC if you want
There is some code already for selecting the input device, it needs to be improved. 
Read the SDL2 docs for audio input capture

```
SDL_AudioDeviceID projectMSDL::selectAudioInput(int _count) {
    // ask the user which capture device to use
    // printf("Please select which audio input to use:\n");
    printf("Detected devices:\n");
    for (int i = 0; i < _count; i++) {
        printf("  %i: 🎤%s\n", i, SDL_GetAudioDeviceName(i, true));
    }

    return 0;
}

int projectMSDL::openAudioInput() {
    // get audio driver name (static)
    const char* driver_name = SDL_GetCurrentAudioDriver();
    SDL_Log("Using audio driver: %s\n", driver_name);

    // get audio input device
    unsigned int i, count2 = SDL_GetNumAudioDevices(true);  // capture, please
    if (count2 == 0) {
        SDL_LogCritical(SDL_LOG_CATEGORY_APPLICATION, "No audio capture devices found");
        SDL_Quit();
    }
    for (i = 0; i < count2; i++) {
        SDL_Log("Found audio capture device %d: %s", i, SDL_GetAudioDeviceName(i, true));
    }

    SDL_AudioDeviceID selectedAudioDevice = 0;  // device to open
    if (count2 > 1) {
        // need to choose which input device to use
	selectedAudioDevice = selectAudioInput(count2);
	if (selectedAudioDevice > count2) {
            SDL_LogCritical(SDL_LOG_CATEGORY_APPLICATION, "No audio input device specified.");
            SDL_Quit();
        }
    }

    // params for audio input
    SDL_AudioSpec want, have;

    // requested format
    SDL_zero(want);
    want.freq = 48000;
    want.format = AUDIO_F32;  // float
    want.channels = 2;
    want.samples = 512;
    want.callback = projectMSDL::audioInputCallbackF32;
    want.userdata = this;

    audioDeviceID = SDL_OpenAudioDevice(SDL_GetAudioDeviceName(selectedAudioDevice, true), true, &want, &have, 0);

    if (audioDeviceID == 0) {
[...]

PR #515, opened 2021-07-28
Add SDL2 hint to use monitor/virtual capture devices which are available starting with SDL 2.0.16.

kblaschke:
Along with Ryan's fix (big thanks!) becoming available in the SDL2 2.0.16 release, adding the newly added hint will finally make "monitor" devices work in SDL's PulseAudio driver.

Closes issues #429 and #30.


revmischa (approved):
Dope. Thanks @kblaschke and @icculus !! 🙏🏻

labkey-matthewb:
Nice

PR #518, opened 2021-08-04
Fixed several build/install errors for out-of-source config's.

lnslbrty:
I'm loving to have several different build config's around, usually only possible if building out-of-source.

P.S.:
Thank's for keeping projectM up-to-date.
The world would be a little bit worse without this great piece of software.
Started using Milkdrop with WinAmp when I was a 11 year old good-for-nothing.
Two decades later, I'm still a good-for-nothing, but now a projectM worshipper.

### DIFF
--- a/src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
+++ b/src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
@@ -74,7 +74,7 @@ std::string read_config();
 #include <string.h>
 #include <errno.h>
 #include <sys/stat.h>
-#include <src/libprojectM/TestRunner.hpp>
+#include <TestRunner.hpp>
 
 //#include <pulsecore/gccmacro.h>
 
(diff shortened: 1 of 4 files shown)


revmischa:
We have cmake now maybe you want to use that instead?

lnslbrty:
So the Autotools system exists only for compatibility?

revmischa:
> So the Autotools system exists only for compatibility?

It will probably be removed once the new cmake setup is a bit more tested and stable. For now there are two. We can remove autotools at any time

src/projectM-pulseaudio/qprojectM-pulseaudio.cpp
```diff
@@ -74,7 +74,7 @@ std::string read_config();
 #include <string.h>
 #include <errno.h>
 #include <sys/stat.h>
-#include <src/libprojectM/TestRunner.hpp>
+#include <TestRunner.hpp>
```
kblaschke, line 77: Having test code directly inside a class isn't really clean IMO, but having the include dir fixed to this is at least a good step in the right direction :)

kblaschke:
Doesn't hurt to fix small issues with the old automake scripts, but for any new source builds it's recommended to use CMake. The more people use it, the more feedback we get and can improve the builds up to the point we can remove the autotools files, Xcode projects and Visual Studio solutions and only keep a single build system for all platforms.

kblaschke (approved):

Issue #519, opened 2021-08-10
Mouse pointer visible on full screen SDL frontend


donny838:
**Describe the bug**
When switching the projectM SDL frontend into fullscreen mode, the mouse pointer stays visible. There is no way to hide it.

**Desktop (please complete the following information):**
 - Application: Steam, SDL
 - OS: macOS, Linux and Windows
 - Version: latest master

donny838:
There’s no way to hide mouse when in full screen on Mac. Quite frustrating and would love to see a way to hide mouse when full screen!

kblaschke:
As the SDL frontend application (which is the one used on macOS and published on Steam) has no interactive UI that requires the mouse, we'll simply add a call to `SDL_ShowCursor(SDL_DISABLE);` in fullscreen mode and `SDL_ShowCursor(SDL_ENABLE);` when switching back to windowed mode. That should take care of the issue on all platforms.

kblaschke:
Testing on macOS showed that it actually works fine, but there might be some confusion regarding the "fullscreen" mode, as there are two different ways to get there:
- Using the SDL-based fullscreen mode by pressing Command+f
- Using the macOS "borderless window" fullscreen mode by clicking on the green title bar icon

The latter one isn't a real fullscreen mode, it's just resizing the window to the screen size and removing the decorations. From SDL's point of view, there is no way to distinguish "maximized" and "native fullscreen" states. We could only check if the window is filling the whole screen by comparing the sizes, and if so, hide the mouse pointer.

Another way of solving it would be adding a hotkey to toggle the mouse pointer in any mode, regardless of fulscreen or windows display.

revmischa:
Can't we hide the cursor when they hit cmd+f?

kblaschke:
Hiding the cursor on toggling fullscreen via the hotkey was already in the code, it just won't trigger if the user clicks on the green, macOS-native fullscreen button in the title bar. Took me a while to guess that @donny838 probably means this. There's no flag or dedicated event to detect it, and even SDL can't unset this fullscreen mode using the toggle function.

So hiding the cursor if the window fills the current screen should be a good workaround. Even if the window still has borders or the user has multiple monitors, it will only be hidden when it's inside the window area, not affecting anything outside.

kblaschke:
the fix is now in master. Please test, if the issue persists reopen the ticket.

Issue #366, opened 2020-06-04
Windows - full screen command - sets a lower resolution than native


internetaddict40:
Hello, I am using the latest ProjectM v3.1.3
My OS is Windows 10 Enterprise, AMD video card
My display is a 4k TV, native resolution set as 3840x2160, max is 4096x2160 for some reason, I have display scaling at 175%

When I hit Ctrl+F in ProjectM, the resolution is set to 2560x1440

Is it possible to make ProjectM set the resolution that the display currently had just before, rather than look at the monitor's spec and get confused by hi-dpi monitors like this?

milkdropper:
I am going to assume you are using projectmSDL. I had a friend report the same issue when stretching monitors. It looks like the SDL library is not aware of scaling and it messes up the math, and I could not come up with a solution for stretching. I just asked him to stop scaling when launching projectmSDL.

I can look into if we can at least fix full screen.

milkdropper:
It looks like it is still an open issue. Can you bump this thread? https://discourse.libsdl.org/t/wip-windows-high-dpi-support/22446/7

internetaddict40:
Thank you, I appreciate it.

It does indeed appear to follow the "fake" resolution that Windows states to non-scaling programs, as at 200% scaling it uses 1920x1080, at 300% it uses 1280x720

Edit: Found a workaround by disabling program scaling and fullscreen optimisations in the Properties - Compatibility dialog. It now fullscreens at 3840x2160

Edit2: Got it, the SDL runtime doesn't have scaling. I will bump that thread with what I found

![Capture](https://user-images.githubusercontent.com/13089641/83754932-7440a700-a675-11ea-9546-a066868e259d.PNG)

milkdropper:
Nice one - this fixes the stretching issue as well! Thanks for sharing the workaround.

kblaschke:
There are issues with fullscreen mode on other platforms as well, mainly boiling down to SDL2 using exclusive fullscreen mode with a change of resolution and maybe even refresh frequency.

IMO we might be better off using `SDL_WINDOW_FULLSCREEN_DESKTOP` instead, which basically maximizes the window on the current display and removes the border. On macOS, this equals using the native "fullscreen" toggle in the title bar, or the "fullscreen" mode most window managers on Linux provide. SDL2 honors those OS specialties, and it might also fix the current issue on Windows.

revmischa:
Sure we can try it

PR #520, opened 2021-08-10
Use SDL2's fake fullscreen mode instead of the exclusive one.

kblaschke:
Also don't maximize the window before toggling fullscreen and hide the mouse pointer if the window size equals the desktop size to also account for the macOS native fullscreen toggle.

Fixes issues #519 and potentially also #366.

### DIFF
--- a/src/projectM-sdl/pmSDL.cpp
+++ b/src/projectM-sdl/pmSDL.cpp
@@ -63,19 +63,6 @@ void projectMSDL::setHelpText(const std::string& helpText)
     projectm_set_help_text(_projectM, helpText.c_str());
 }
 
-void projectMSDL::maximize()
-{
-    SDL_DisplayMode dm;
-    if (SDL_GetDesktopDisplayMode(0, &dm) != 0)
-    {
-        SDL_Log("SDL_GetDesktopDisplayMode failed: %s", SDL_GetError());
-        return;
-    }
-
-    SDL_SetWindowSize(win, dm.w, dm.h);
-    resize(dm.w, dm.h);
-}
-
 /* Stretch projectM across multiple monitors */
 void projectMSDL::stretchMonitors()
 {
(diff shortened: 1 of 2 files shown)


src/projectM-sdl/pmSDL.cpp
```diff
@@ -281,6 +269,13 @@ void projectMSDL::addFakePCM() {
 void projectMSDL::resize(unsigned int width_, unsigned int height_) {
     width = width_;
     height = height_;
+
+		// Hide cursor if window size equals desktop size
+	  SDL_DisplayMode dm;
+	  if (SDL_GetDesktopDisplayMode(0, &dm) == 0) {
```
revmischa, line 275: I think this just checks if the call didn't fail

kblaschke: That's right. But only if the call is successful we can use the `dm` structure's members, which means if the call fails, we simply don't compare the size and carry on as before, only resizing the window.

kblaschke:
Rebased commit on latest master and PGP-signed the commit.

src/projectM-sdl/pmSDL.cpp
revmischa: I don't understand why it would be expected to fail and the docs are not enlightening on this matter https://wiki.libsdl.org/SDL_GetDesktopDisplayMode
And we aren't using the `dm` structure's members.
So I still don't understand the logic here

kblaschke: Only way it can fail is if SDL_video is uninitialized or the current display index is invalid:
https://github.com/libsdl-org/SDL/blob/402b86f2a88b5ee3e112dbde2ec232ee5f36572f/src/video/SDL_video.c#L141

Inj this case, `dm` will be left uninitialized and accessing the members would return random values, causing the mouse pointer to stay on screen. Doesn't hurt to check the return value as suggested by the SDL docs though.

PR #521, opened 2021-08-10
Replace "#ifdef USE_GLES" with "#if USE_GLES"

kblaschke:
The old configure script either did not #define USE_GLES at all, or defined it to 1. CMake only supports #define or #undef without a value, or #define it to either 0 or 1. Using #if will make it compatible with both ways of configuring the macro.

This will fix the issues with shaders not compiling, complaining about the OpenGL version being too low or #version missing in the shader declaration.


revmischa (approved):
Oh nice I hope this fixes it on macOS

kblaschke:
> Oh nice I hope this fixes it on macOS

Yep, it does - tested it on a Big Sur iMac. Also fixes failing shader compilation on Linux.

PR #522, opened 2021-08-10
Add a dummy source file so the CMake Xcode generator creates libprojectM.a

kblaschke:
Xcode targets seem to need at least one "real" source file for static libs to be built. Added a dummy file with an empty function to satisfy this requirement.


kblaschke:
To test is on macOS, explicitly run CMake with `-GXcode` in the configuration phase. Default is the `Unix Makefiles` generator.

revmischa (approved):
strike one against cmake!

Issue #517, opened 2021-08-03
Please add Qt6 support


DAP-DarkneSS:
Qt 6.2

[   36s] + ./configure --host=i586-suse-linux-gnu --build=i586-suse-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --disable-dependency-tracking --disable-sdl --enable-qt --disable-static --disable-rpath
[   36s] uname.bin: invalid option -- 'X'
[   36s] Try 'uname.bin --help' for more information.
[   36s] /usr/bin/arch: invalid option -- 'k'
[   36s] Try '/usr/bin/arch --help' for more information.
[   36s] ./configure: line 2353: /usr/convex/getsysinfo: No such file or directory
[   36s] ./configure: line 2353: /usr/bin/hostinfo: No such file or directory
[   36s] ./configure: line 2353: /bin/machine: No such file or directory
[   36s] ./configure: line 2353: /usr/bin/oslevel: No such file or directory
[   36s] ./configure: line 2353: /bin/universe: No such file or directory
[   36s] checking for a BSD-compatible install... /usr/bin/install -c
[   36s] checking whether build environment is sane... yes
[   36s] checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
[   36s] checking for gawk... gawk
[   36s] checking whether make sets $(MAKE)... yes
[   36s] checking whether make supports nested variables... yes
[   36s] ./configure: line 3196: cygpath: command not found
[   36s] checking how to create a pax tar archive... gnutar
[   36s] checking whether to enable debugging... no
[   36s] checking whether make supports the include directive... yes (GNU style)
[   36s] checking for i586-suse-linux-gnu-gcc... no
[   36s] checking for gcc... gcc
[   36s] checking whether the C compiler works... yes
[   36s] checking for C compiler default output file name... a.out
[   36s] checking for suffix of executables... 
[   37s] checking whether we are cross compiling... no
[   37s] checking for suffix of object files... o
[   37s] checking whether we are using the GNU C compiler... yes
[   37s] checking whether gcc accepts -g... yes
[...]

PR #523, opened 2021-08-10
Updated Qt UIs and CMake build scripts to support Qt 6.

kblaschke:
Changes:
- Added new parameter QT_VERSION to let the user force a specific Qt version to be searched.
- Auto-detect Qt version if not forced, preferring Qt 6.
- Updated the projectM render widget to use QOpenGLWidget instead of the deprecated QGLWidget.
- Changed some calls and types to be compatible between Qt 5 and 6, e.g. QXmlStreamReader.
- Some small fixes pointed out by assertions in the debug Qt 6 version.

This will fix issue #517.


src/projectM-qt/qplaylistmodel.cpp
```diff
@@ -382,9 +382,12 @@ void QPlaylistModel::clear()
...
 void QPlaylistModel::clearItems()
 {
-	beginRemoveRows ( QModelIndex(), 0, rowCount()-1 );
-	m_projectM.clearPlaylist();
-	endRemoveRows();
+    if (rowCount() > 0)
+    {
+        beginRemoveRows(QModelIndex(), 0, rowCount() - 1);
+        m_projectM.clearPlaylist();
+        endRemoveRows();
```
revmischa, line 389: I dunno what's going on here but maybe we should always be calling `endRemoveRows`?

revmischa (approved):

src/projectM-qt/qplaylistmodel.cpp
kblaschke: As the model represents the playlist items, we need to call [`beginRemoveRows`](https://doc.qt.io/qt-6/qabstractitemmodel.html#beginRemoveRows) first to tell the model that we are going to change the underlying structure, then perform the operation (clearing the playlist) and call `endRemoveRows` to tell teh model that we're done removing the rows. You must only call `endRemoveRows` after a matching `beginRemoveRows` call. I've only added an `if` statement checking whether there are any rows because otherwise, the last row index would be -1, which is invalid and will cause an assertion to fail (first row must be larger or equal to last row). It's not critical, Qt release builds will simply ignore it, but it might as well crash the application.

PR #525, opened 2021-08-11
Added text menu code again for Mac and Linux

kblaschke:
Also fixed a crash bug in the preset search menu.

Browsing through the search results using the up/down keys was crashing the SDL app. Fixed some index-off-by-one errors and a wrongly used parameter to check for search results.


revmischa (approved):
yikes ok

kblaschke:
> yikes ok

My "yikes" moment was seeing that the current preset index is actually 1-based. Quite unexpected.

PR #530, opened 2021-08-17
Added a C-style interface header and wrapper for projectM.

### LINKED ISSUE #529: Only provide a C-language interface in libprojectM by default.
**Is your feature request related to a problem? Please describe.**
Currently, projectM only supports a C++ interface in both static and shared libraries. This works fine as long as both projectM and applications integrating it are compiled with the same toolchain, C++ library and C++ language level, but will cause issues from dynamic linker error to segmenation faults if there are differences, as STL types are used in the interface.

In detail: STL types are, as the name suggests, template classes with (mostly) header-only implementations. Between different compilers, compiler versions and depending on the C++ language level selected during compilation, STL type implementations - specifically the memory layout - will differ. An `std::string` returned by a projectM function (for example the current preset name) would be free'd by the host application, but the destructor of this `std::string` might expect a different layout of members than the implementation used to compile projectM.

**Describe the solution you'd like**
Adding a C wrapper that exposes only C-style functions instead of the OOP interface, hiding the ProjectM class instance and only using standard types like `char*`, `int`, `float` etc. as parameters and return values.

The projectM class instance pointer will only be returned as an opaque pointer type, passed to each subsequent functions call. The wrapper functions then convert the pointer back to the proper C++ class instance to call the member functions in the specified instance.

Using the wrapper, all STL operations are done with the same implementation as the projectM code itself. For returned `char*` type values, a `free` function should be provided by the wrapper as some platforms like Windows separate heaps between applications and shared libraries, which can cause an application-side free on a library-allocated pointer to crash the program.

To make the C++ interface still useable, there are two options:

1. Only install the C++ headers if a new build option like `INSTALL_CXX_HEADERS` is specified, with an author warning emitted that this can cause problems.
2. Always install the C++ headers as usual, but add a `#pragma warning` (that can be suppressed) to the main C++ header that using this interface requires some additional care.

**Describe alternatives you've considered**
There is no real alternative, as there's no way to check STL compatibility on library load.

kblaschke:
This will make it easier to integrate projectM into C applications and also removes potential issues of STL/C++ version differences between host and library.

Changed visibility of all symbols except the C API to "hidden". This effectively prevents the C++ API from being used in the shared library.

Added an option "ENABLE_CXX_INTERFACE" to re-enable the old export behaviour, displaying a prominent author warning after configuring the build if enabled.

Fixes #529.


CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ set(PROJECTM_LIB_DIR "lib" CACHE STRING "Library installation directory, relativ
 set(PROJECTM_INCLUDE_DIR "include" CACHE STRING "Header installation directory, relative to the install prefix.")
 
 # Feature options, including dependencies.
+option(ENABLE_CXX_INTERFACE "Enable exporting all C++ symbols, not only the C API, in the shared library. Warning: This is dangerous!" OFF)
```
revmischa, line 40:
```suggestion
option(ENABLE_CXX_INTERFACE "Enable exporting all C++ symbols, not only the C API, in the shared library. Warning: This is not very portable." OFF)
```

kblaschke:
Building with default settings will now hide all symbols which are not explictily exported by `projectM.h`. This will be a breaking change for almost all current projectM integrations which do not directly use the sources (like the Clementine Player for example). Yet forcing authors to switch to the C API will certainly improve compatibility and make it far easier to use projectM in C-based applications, OBS being a prime example.

revmischa (approved):
This is amazing work thanks so much. Now the docs should be updated so downstream maintainers are aware of it and maybe someday some HTML or MD generated docs would be lovely.

kblaschke:
> This is amazing work thanks so much. Now the docs should be updated so downstream maintainers are aware of it and maybe someday some HTML or MD generated docs would be lovely.

Yes, some good documentation, including Doxygen-based API docs and Markdown-based end-user/developer manuals would be nice to have. Should be easy to build with Sphinx and we could host it on readthedocs.io.

kblaschke:
All reference UIs now build & run fine with the C-style API while using projectM as a shared library.

kblaschke:
Will merge this PR as everything (except the already broken JACK SDL app) is in a working state. Currently working on streamlining the addPCM methods, fur which I'll create another PR.

Build failures are realted to GitHub's VMs having network issues while downloading the prerequisite packages. Ignoring this for now.

Issue #379, opened 2020-07-10
Docs Needed: Keys, Config File Reference [ToDo]


revmischa:
If someone wants to update the README or otherwise this would be greatly useful as evidenced by comments in various issues here
And where to find config files is needed documenting

Keybindings can mostly be found here: https://github.com/projectM-visualizer/projectm/blob/master/src/libprojectM/KeyHandler.cpp#L115

Config file reference and paths should be nearby

revmischa:
Also would be nice: help menu, text when opening projectMSDL

milkdropper:
Can you accept https://github.com/projectM-visualizer/projectm/pull/284 so the config file has a proper home in Windows? I can then update the build process on projectMSDL on Windows so it pushes a config.inp into the package and update the documentation at the same time.

Shrey-Dua:
Hi ! 
I am a beginner and i would like to contribute , could you help me out please ?

revmischa:
@Shrey-Dua sure what do you need help with? https://docs.github.com/en/github

KishkinJ10:
I am a beginner can i work on this ? can you please guide me from where should i start?

revmischa:
Open a pull request with changes to README.md, create a new markdown file and link to it from the README.md, or set up a [wiki](https://github.com/projectM-visualizer/projectm/wiki). Your choice. Wiki might be friendliest.

NoobMaster1999:
is this issue still open?

revmischa:
Yes please just open a PR

PR #537, opened 2021-08-25
Update README.md add wiki page

NoobMaster1999:
Resolve: #379

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,3 +149,6 @@ If you maintain packages of libprojectM, we are happy to work with you! Please n
 
 ## License
 [LGPL](https://github.com/projectM-visualizer/projectm/raw/master/LICENSE.txt)
+
+## Wiki
+More information for developers is available from the [projectM Wiki](https://github.com/projectM-visualizer/projectm/wiki).


rorypierce:
Yo every1:... I'm a little rusty since having to work for so many companies that hated OpenSource. Yuck... The CI failure--what I understand thus far--... a CMake generated file *_export.h is somewhat bad practice and it seems to be failing on CI... Can I help in any way. Finally found a workaround; and this suite is sweet! Thanks.

revmischa:
> Yo every1:... I'm a little rusty since having to work for so many companies that hated OpenSource. Yuck... The CI failure--what I understand thus far--... a CMake generated file *_export.h is somewhat bad practice and it seems to be failing on CI... Can I help in any way. Finally found a workaround; and this suite is sweet! Thanks.

Sure please submit a PR!

kblaschke:
> a CMake generated file *_export.h is somewhat bad practice and it seems to be failing on CI... 

@rorypierce  Why is this actually bad practice and what would be good practice for a cross-platform, CMake-built application like projectM?

projectM is now built with CMake, so the export header will contain the Windows-specific dllimport/dllexport declspec definitions. On UNIX, these will be generated using visibility attributes or any other method supported by the compiler used.  This saves us all the work to manually write these macros. If you want to know more about how the file is created, please read the [GenerateExportHeader module documentation](https://cmake.org/cmake/help/v3.22/module/GenerateExportHeader.html) in the CMake docs, which is the module we use to create the header.

If you talk about the TravisCI builds always failing: This has nothing to do with the generated export definition header, at least that's not the only thing, just the first issue causing the build to fail. The TravisCI build still uses the manually created Visual Studio solution, which simply no longer works as it wasn't updated anymore and will be removed completely along witht he Xcode project files and the autoconf-based configure script. To properly build the project using CMake, the server would need a few adjustments like installing CMake and vcpkg, but I have no access to the machine.

PR #538, opened 2021-08-30
Fix FileScanner::scanGeneric for filesystems that dont set d_type

falkTX:
filesystems are free to return DT_UNKNOWN, which happens on XFS on my system.
Typically not a problem because on linux HAVE_FTS_H will be set which activates posix search, but in case that is not available for some reason the alternative code path was failing.
Seems a quick fix, so I just added it in.


kblaschke (approved):
Looks totally fine to me, good catch. Sorry for the late review though, this just slipped through.

PR #539, opened 2021-08-30
Fix windows includes for mingw

falkTX:
mingw kindly asks for `winsock2.h` to always be included before `windows.h`
doesn't hurt or break anything for MSVC as far as I know, so why not.

### DIFF
--- a/src/libprojectM/projectM-opengl.h
+++ b/src/libprojectM/projectM-opengl.h
@@ -31,9 +31,10 @@
 #include <EGL/eglext.h>
 #elif defined(_WIN32)
 #define GLM_FORCE_CXX03
+# include <winsock2.h>
 # include <windows.h>
-#include "GL/glew.h"
-#include "GL/wglew.h"
+# include "GL/glew.h"
+# include "GL/wglew.h"
 #else /* linux/unix/other */
 # if USE_GLES
 #  include <GLES3/gl3.h>


kblaschke:
AFAIK this warning is only shown if you actually include winsock2.h after windows.h as order is important here because of some redefines. But projectM doesn't use any socket functions or types and thus doesn't require winsock2.h, so this warning should never be shown (as it is emitted in winsock2.h).

So the question is: do we really need to include this header?

falkTX:
ah I was not aware of that actually.
I include `projectM-opengl.h` in my project before other headers, and was getting this warning, that explains why.

So on my side I would just need to include `winsock2.h` before the projectM stuff, correct?
I guess you can close this one as invalid then.

kblaschke:
That would probably fix it, yes. The projectM-opengl.h header is also more of an internal construct to select the different OpenGL libraries on each platform, and not really meant to be used by an outside project. Will most probably be removed from the public headers in the next release, leaving only the new C API header.

falkTX:
> The projectM-opengl.h header is also more of an internal construct to select the different OpenGL libraries on each platform, and not really meant to be used by an outside project. Will most probably be removed from the public headers in the next release, leaving only the new C API header.

I am aware of that, but in my case I use projectM as a submodule, compiled alongside the main project so the final binary is self-contained. The case for importing that header is so that the glew version is imported from projectM rather than regular OpenGL stuff.

Issue #543, opened 2021-09-27
libprojectm: Qt PulseAudio CMake wrong linker path


digitalcircuit:
## In short
**Describe the bug**
* `projectM-pulseaudio` build variant fails to build with CMake on Ubuntu 20.04 & 21.04
  * Specific build variant: `cmake .. -DENABLE_SDL=OFF -DENABLE_SDL_UI=OFF -DENABLE_PULSEAUDIO=ON`

*See also: [pull request #536: libprojectm: Fix static DL_LIBS for Qt PulseAudio](https://github.com/projectM-visualizer/projectm/pull/536 )*

**Screenshots**
```
[100%] Linking CXX executable projectM-pulseaudio
/usr/bin/ld: ../libprojectM/libprojectM.a(NativePresetFactory.cpp.o): undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/9/../../../x86_64-linux-gnu/libdl.so: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [src/projectM-pulseaudio/CMakeFiles/projectM-pulseaudio.dir/build.make:168: src/projectM-pulseaudio/projectM-pulseaudio] Error 1
make[1]: *** [CMakeFiles/Makefile2:673: src/projectM-pulseaudio/CMakeFiles/projectM-pulseaudio.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

![Screenshot of two VirtualBox virtual machines showing the results of following the compilation steps.  Both end in the same error message as mentioned earlier in this pull request, except Ubuntu 21.04 references '10' instead of '9' due to updates.](https://zorro.casa/sync/Hosting/Utilities/Development/projectM/pr/fix-build-static-pulseaudio/Screenshot%20-%20projectM%20CMake%20build%20test%20Ubuntu%2020.04%20and%2021.04.png#v1 )

**Desktop (please complete the following information):**
 - Application: Qt with PulseAudio
 - OS: Linux (Kubuntu 20.04, Ubuntu 20.04, Ubuntu 21.04)
 - Version: [commit hash 8446784](https://github.com/projectM-visualizer/projectm/commit/844678472e434986a6c94de02d3e3aa3deab60a6 )

**Additional context**

### Steps
1.  Install Ubuntu (20.04.3 LTS and 21.04) with "install third-party drivers" checked
[...]

PR #544, opened 2021-10-15
Fixes "libprojectm: Qt PulseAudio CMake wrong linker path" Issue #543

rorypierce:
I view my relationship with CMake as more hate than love; however, it's pretty okay 8^)... @digitalcircuit thanks for the very helpful documentation! I was super-frustrated that I couldn't get the non-SDL rendering to work (only ~25fps)--now it's supah-High! This project is fantastic!

PR #551, opened 2021-11-05
Fix native preset build issues

kblaschke:
First commit is the minimum required change to fix issue #543 and is basically overlapping with PR #544.

This PR contains an additional commit that fully disables native preset loading support if the `ENABLE_NATIVE_PRESETS` build setting if `OFF` (default value) by omitting the NativePresetFactory library.

### DIFF
--- a/src/libprojectM/PresetFactoryManager.cpp
+++ b/src/libprojectM/PresetFactoryManager.cpp
@@ -10,17 +10,16 @@
 //
 //
 #include "PresetFactoryManager.hpp"
-
-#ifndef DISABLE_MILKDROP_PRESETS
 #include "MilkdropPresetFactory/MilkdropPresetFactory.hpp"
-#endif
 
-#ifndef DISABLE_NATIVE_PRESETS
+#ifdef ENABLE_NATIVE_PRESETS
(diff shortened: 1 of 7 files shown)


kblaschke:
The Windows build failure is fixed in separate PR #552 as it's unrelated to this changeset (was introduced in #546).

src/libprojectM/PresetFactoryManager.cpp
```diff
@@ -45,13 +44,11 @@ void PresetFactoryManager::initialize(int gx, int gy) {
 	}
 	  
 	PresetFactory * factory;
-	
-	#ifndef DISABLE_MILKDROP_PRESETS
```
revmischa, line 49: why was this even here lol

revmischa (approved):

src/libprojectM/PresetFactoryManager.cpp
kblaschke: Definitely a good question. Define this macro and projectM is simply useless...

Probably someone once had the idea to make the library more modular, having multiple types of presets which could be enabled or disabled independently.

PR #553, opened 2021-11-05
Add GLES3 support to CMake builds on Linux.

kblaschke:
Two commits in this changeset:

- Added GLES3 flag to SOIL2 as it was only supporting GLES up to version 2. Basically only the included headers have changed.
- Added a patched copy of CMake's `FindOpenGL` module, which has additional components to detect GLES 1-3. I'll propose this change as an upstream patch to become part of CMake. You may want to [follow my dev branch there](https://gitlab.kitware.com/kblaschke/cmake/-/tree/feature/gles-detection) for any progress on that.

Windows support will be added later, as GLES is currently used in projectM exclusively on Linux platforms. Build configuration will fail if GLES is enabled on any other platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,11 +93,19 @@ if(ENABLE_SDL)
 endif()
 
 if(ENABLE_GLES)
-    # Might be implemented in the CMake OpenGL module.
-    # We may provide a find module in the future, and add support for CMake's own module later:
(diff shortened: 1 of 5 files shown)


kblaschke:
Last push only removed a forgotten test output line in the find module.

In the meantime, I compiled the branch on a RaspberryPi 4 with libSDL 2.0.16 and GCC 8.3, now works perfectly fine with Mesa GLES 3.1 there, including presets using warp and composite shaders.

PR #554, opened 2021-11-07
CMake: Support emscripten builds

kblaschke:
Implement proper emscripten support in the new CMake build system.

Should be merged after PR #553 as the GLES changes are required for emscripten and contained in this branch.

Changes:
- Removed the unneeded and non-standard gl3ext.h include in SOIL2 that's not present in the official GLES3 specs.
- Auto-detect emscripten via CMAKE_SYSTEM_NAME instead of using an option.
- Disable all other feature detection code if emscripten is used.
- Remove old project files and convert the sample entry point app to use the new C-based API and SDL2 only.
- Updated emscripten build instructions in the README.md.


src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -133,23 +128,15 @@ int main( int argc, char *argv[] ) {
...
-        printf("SDL init version 2\n");
-    #elif SDL_MAJOR_VERSION==1
-        screen = SDL_SetVideoMode(width, height, 32, SDL_OPENGL | SDL_DOUBLEBUF);
-        printf("SDL init version 1\n");
-        if (! screen) {
-            fprintf(stderr, "Failed to create renderer with SDL_SetVideoMode(): %s\n", SDL_GetError());
-            return PROJECTM_ERROR;
-        }
-    #endif
+    app.win = SDL_CreateWindow("SDL Fun Party Time", 50, 50, width, height, 0);
```
revmischa, line 131:
```suggestion
    app.win = SDL_CreateWindow("projectM", 50, 50, width, height, 0);
```

src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -133,23 +128,15 @@ int main( int argc, char *argv[] ) {
...
+    app.rend = SDL_CreateRenderer(app.win, 0, SDL_RENDERER_ACCELERATED);
+    if (! app.rend) {
+        fprintf(stderr, "Failed to create renderer: %s\n", SDL_GetError());
+        return 1;
+    }
+    SDL_SetWindowTitle(app.win, "SDL Fun Party Time");
```
revmischa, line 137:
```suggestion
    SDL_SetWindowTitle(app.win, "projectM");
```

src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -133,23 +128,15 @@ int main( int argc, char *argv[] ) {
...
+    printf("SDL init version 2\n");
```
revmischa, line 138:
```suggestion
```

revmischa (approved):
super exciting!

src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -5,224 +5,275 @@
...
+    if (!app.audioInputDevicesCount)
+    {
+        fprintf(stderr, "No audio input capture devices detected. Faking audio using random data.\n");
+        return false;
     }
 
-    printf("count: %d\n", count);
-    for (i = 0; i < count; ++i) {
-        printf("Audio device %d: %s\n", i, SDL_GetAudioDeviceName(i, 0));
+    printf("count: %d\n", app.audioInputDevicesCount);
```
revmischa, line 62:
```suggestion
```

src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -5,224 +5,275 @@
...
+
+    // Start with first device
+    app.audioInputDevice = SDL_OpenAudioDevice(SDL_GetAudioDeviceName(0, true), true, &want, &have, 0);
+    app.audioChannelsCount = have.channels;
+
+    if (have.channels == 1 || have.channels == 2)
+    {
+        SDL_PauseAudioDevice(app.audioInputDevice, false);
+        printf("Audio device specs: Channels=%d, Samplerate=%d, Format=%d\n", have.channels, have.freq, have.format);
+    }
```
revmischa, line 85: maybe else explode?

src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -5,224 +5,275 @@
...
+            }
             break;
         case SDL_QUIT:
             app.done = true;
             break;
     }
 
+    if (app.audioInputDevicesCount == 0)
+    {
+        short pcm_data[2][512];
```
revmischa, line 171: Refactor into a function

revmischa (approved):
exciting! 
next steps I think: 
* bundle presets/textures with it so it can load them
* embed it inside a Create-React-App (w/ typescript) app

src/projectM-emscripten/projectM_SDL_emscripten.cpp
```diff
@@ -5,224 +5,275 @@
...
+    app.settings.window_width = width;
+    app.settings.window_height = height;
+    app.settings.soft_cut_duration = 3; // seconds
+    app.settings.preset_duration = 5; // seconds
+    app.settings.beat_sensitivity = 0.8;
+    app.settings.aspect_correction = 1;
+    app.settings.easter_egg = 0; // ???
+    app.settings.shuffle_enabled = 1;
+    app.settings.soft_cut_ratings_enabled = 1; // ???
+    app.settings.preset_url = projectm_alloc_string(8);
```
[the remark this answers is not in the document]
kblaschke: That's not too hard. We can just package the whole presets dir, and use the standard filesystem calls in projectM (using the preset path setting) to load them just as in desktop versions.

https://emscripten.org/docs/porting/files/packaging_files.html

Added a commit that packages the presets by adding a simple linker option in CMake. Works out of the box.

src/projectM-emscripten/projectM_SDL_emscripten.cpp
kblaschke: Count should always be 1 in browser environments, so yeah, this output is rather unnecessary. No matter how many mics are connected, I always got a single "System Audio Device" in the enumeration, even if the user hasn't allowed mic access. If no access is allowed, the audio data provided by SDL is just zeroes.

src/projectM-emscripten/projectM_SDL_emscripten.cpp
kblaschke: Probably a case that never happens, so we're just not enabling the audio callback. Could alternatively fall back to the random audio generator to there's some movement at least.

In Chrome, Safari and Firefox at least, SDL always give 2-channel, 32-bit floating point audio with two channels, even if no input device is even connected. Looks more like a generic device that feeds from the Browser's HTML5 audio API.

src/projectM-emscripten/projectM_SDL_emscripten.cpp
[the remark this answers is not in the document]
kblaschke: BTW, textures in the presets dir are of course also packaged automatically, e.g. the "tux flower" preset displays the tux image.

PR #556, opened 2021-11-15
Add missing "round" intrinsic in HLSLParser

kblaschke:
Just a one-liner.
Adds the missing "round" intrinsic in the HLSL transpiler, which is used in a few presets ("EVET - The Rift" for example) and can be used 1:1 in GLSL.


revmischa (approved):
sicc

kblaschke:
Can't explain why this basic function was missing. Probably not the only one, but if I encounter any other missing intrinsic, I'll see if I can add it too. Some functions like `alle()` can't be easily converted though as they have no equivalent in GLSL.

PR #558, opened 2021-11-16
Add DPI awareness flag to the SDL app on Windows.

kblaschke:
This was fixed for the manually created VS solution in commit 3171507, but not yet ported to CMake.

### DIFF
--- a/src/projectM-sdl/CMakeLists.txt
+++ b/src/projectM-sdl/CMakeLists.txt
@@ -28,6 +28,20 @@ target_link_libraries(projectMSDL
         ${CMAKE_DL_LIBS}
         )
 
+if(MSVC)
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.16")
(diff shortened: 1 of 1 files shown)


kblaschke:
It's basically exactly the same thing that was merged in PR #512 for the pre-made VS solutions, adding the CMake flag used to set this value in the generated solution.

It'll tell Windows that the executable is "DPI aware", which means that Windows doesn't actively scale up the window, but leaves it to the application to manage font/UI sizes etc. according to the monitor's DPI. In projectM's case, that's easy: we just render at the native resolution, and that's it. Windows automatically scaling our window means lower-than-intended resolution on HighDPI displaying with scaling set to >100%, e.g. at 200%, projectM will only render at half the native window size.

kblaschke:
I'll just merge this change, successfully tested it on my Windows 10 box.

With this change, projectM renders at full native window size with scaling set to >100%. Without, it uses the reduced resolution, creating a blurry image.

PR #561, opened 2021-11-23
Improve overlay text rendering performance and a few Renderer cleanups.

kblaschke:
For each line of text, glText was fully reinitialized, which meant that the font texture, vertex and fragment shaders were recompiled and uploaded to the GPU, then destroyed again. This created a huge FPS drop, especially in GLES environments.

Also moved the text rendering into a separate class and cleaned up the initialization of the Renderer class, moving all default values to the header and deleting the default constructor to avoid unwanted effects.

### DIFF
--- a/src/libprojectM/Renderer/CMakeLists.txt
+++ b/src/libprojectM/Renderer/CMakeLists.txt
@@ -6,31 +6,33 @@ add_library(Renderer OBJECT
         BeatDetect.hpp
         Filters.cpp
         Filters.hpp
+        MenuText.cpp
+        MenuText.h
         MilkdropWaveform.cpp
         MilkdropWaveform.hpp
+        PerPixelMesh.cpp
+        PerPixelMesh.hpp
         PerlinNoise.cpp
         PerlinNoise.hpp
         PerlinNoiseWithAlpha.cpp
         PerlinNoiseWithAlpha.hpp
-        PerPixelMesh.cpp
-        PerPixelMesh.hpp
-        PipelineContext.cpp
-        PipelineContext.hpp
         Pipeline.cpp
         Pipeline.hpp
-        Renderable.cpp
-        Renderable.hpp
-        Renderer.cpp
-        Renderer.hpp
+        PipelineContext.cpp
+        PipelineContext.hpp
         RenderItemDistanceMetric.cpp
         RenderItemDistanceMetric.hpp
         RenderItemMatcher.cpp
         RenderItemMatcher.hpp
         RenderItemMergeFunction.hpp
+        Renderable.cpp
+        Renderable.hpp
+        Renderer.cpp
+        Renderer.hpp
         Shader.cpp
+        Shader.hpp
         ShaderEngine.cpp
         ShaderEngine.hpp
-        Shader.hpp
         StaticGlShaders.cpp
         Texture.cpp
         Texture.hpp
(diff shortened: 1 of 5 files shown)


src/libprojectM/Renderer/Renderer.hpp
```diff
@@ -192,20 +188,10 @@ class Renderer
 private:
 
   PerPixelMesh mesh;
-  BeatDetect *beatDetect;
-  TextureManager *textureManager;
-  Pipeline* currentPipe;
-  TimeKeeper *timeKeeperFPS;
-  TimeKeeper *timeKeeperToast;
```
kblaschke, line 199: These TimeKeeper pointers were never used or set.

src/libprojectM/Renderer/CMakeLists.txt
```diff
@@ -6,31 +6,33 @@ add_library(Renderer OBJECT
         BeatDetect.hpp
         Filters.cpp
         Filters.hpp
+        MenuText.cpp
+        MenuText.h
```
kblaschke, line 10: These two lines added. The rest was sorting the lines alphabetically.

src/libprojectM/Renderer/MenuText.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    _viewportWidth = viewportWidth;
+}
+
+void MenuText::CleanUp()
+{
+    gltTerminate();
+}
+
+void MenuText::DrawBegin() const
```
kblaschke, line 36: Moved these calls outside Draw() as binding/unbinding shaders and textures also adds some serious overhead when drawing the preset menu.

src/libprojectM/Renderer/MenuText.cpp
```diff
@@ -0,0 +1,163 @@
...
+        if (textLine.find('\n') != std::string::npos)
+        {
+           textLine.pop_back();
+           textLine.pop_back();
+           textLine.pop_back();
+           textLine += "...";
+        }
+    }
+
+    // redraw without transparency
```
kblaschke, line 102: Not sure why it was drawn transparent Probably someone thought it's required to get the actual text width, but that is simply calculated from text & scale.

src/libprojectM/Renderer/Renderer.cpp
```diff
@@ -988,22 +841,29 @@ void Renderer::draw_stats()
...
+	m_menuText.Draw(this->fps(), 30, 20, 2.5);
+    m_menuText.DrawEnd();
 #endif /** USE_TEXT_MENU */
 }
 
 void Renderer::draw_toast()
 {
 #ifdef USE_TEXT_MENU
-	drawText(this->toastMessage().c_str(), (vw/2), (vh/2), 2.5, GLT_CENTER, GLT_CENTER);
+    m_menuText.DrawBegin();
```
revmischa, line 863: It could probably be a macro or wrapper that takes a lambda

revmischa (approved):

src/libprojectM/Renderer/Renderer.cpp
kblaschke: For a single line, sure. Or just a boolean parameter `singleLine` that calls begin/end automatically if true.

As discussed on Discord, this whole text rendering should prbably be outsourced into a separate component that injects drawables similar to the rest of the rendering pipeline, so the `Renderer` class doesn't need to care about this stuff. Saving that for later.

PR #564, opened 2021-11-24
Removed autoconf/automake build system files. Please use CMake instead.

kblaschke:
Hopefully got all autotools-related files killed.

If you find other remnants that were only used by autoconf or automake, please put up a note here.


revmischa (approved):
Hell yea

PR #565, opened 2021-11-24
Removed Visual Studio project files. Please use CMake instead.

kblaschke:
One exception: EyeTune still has its own solution, as it's a UWP app and not yet ported to CMake.

Also removed the AppVeyor build config as it was broken for a long time. Updated documentation so it no longer states devs can use the pre-made solution files.


revmischa (approved):

revmischa:
Wondering why appveyor ran here

kblaschke:
> Wondering why appveyor ran here

Let's see if it stops if we remove the config file.

PR #569, opened 2021-11-25
Replaced Perlin noise textures with the original Milkdrop noise textures.

kblaschke:
Also added two minimal presets that simply display the noise_lq and noisevol_lq textures. Load the presets in an old version and compare the output to the changed code in this PR to see what's different.


presets/tests/260-compshader-noise_lq.milk
```diff
@@ -0,0 +1,13 @@
+[preset00]
```
labkey-matthewb, line 1: So happy to see new test presets!

kblaschke: Oh, does it really require this line in Milkdrop? The projectM parser just ignores it. :D

These presets were the easiest way to visualize the noise textures, before and after. I've also made a version that cyclces through the four individual color channels of the 2D noise texture plus RGB combined to make it clear the old implementation simply lacked a lot of random data. I can also add this if it's useful.

```
[preset00]
per_frame_1000=// spectrum vs pcm

fDecay=0
warp=0.000000
wave_a=0

comp_1=`shader_body
comp_2=`{
comp_3=`float t = time % 5;
comp_4=`if(t == 0) { ret = float3(tex2D(sampler_fw_noise_lq, uv).x); }
comp_5=`if(t == 1) { ret = float3(tex2D(sampler_fw_noise_lq, uv).y); }
comp_6=`if(t == 2) { ret = float3(tex2D(sampler_fw_noise_lq, uv).z); }
comp_7=`if(t == 3) { ret = float3(tex2D(sampler_fw_noise_lq, uv).w); }
comp_8=`if(t == 4) { ret = tex2D(sampler_fw_noise_lq, uv); }
comp_9=`}
```

src/libprojectM/Renderer/MilkdropNoise.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+void MilkdropNoise::generate2D(int size, int zoomFactor, uint32_t** textureData)
+{
+    *textureData = new uint32_t[size * size]();
+
+    // write to the bits...
+    auto dst = *textureData;
+    auto RANGE = (zoomFactor > 1) ? 216 : 256;
+    srand(time(nullptr));
```
revmischa, line 34: don't think we need to srand every time do we?

revmischa (approved):

src/libprojectM/Renderer/MilkdropNoise.cpp
kblaschke: Probably not. Milkdrop does it though, just with a different RNG from the Windows API.

Maybe I should better convert this to std::random instead.

kblaschke: Done. Decided to still create a PRNG in each function to skip the need to either declare two additional class members or a parameter to pass on the generator into the noise functions. Distribution and random seed should be better now (not that it is highly relevant for a noise texture).

PR #575, opened 2021-12-13
SDL: Allow using the system default capture device

chatziko:
SDL allows to use the "system default" capture device by passing `device == NULL` to [`SDL_OpenAudioDevice`](https://wiki.libsdl.org/SDL_OpenAudioDevice).
This is a well-behaved default option that has several advantages:

- It is much more likely that the user expects to use the device configured as system default, rather than the one that happens to be at index 0 (the one currently used by `projectMSDL`).

- When the default device changes, SDL __automatically moves to the new device__. I tried this with `pulsemixer`, I change the default capture device when `projectMSDL` is running (with this PR) and I immediately see `projectMSDL` seamlessly switching to the new one!

- The system default can be changed via a variety of methods (GUI tools, command line scripts, etc), it is more user friendly than having to use `Ctrl-I`. Such changes also persist across restarts of `projectMSDL`, chages via `Ctrl-I` do not.

- It allows to use __monitor__ devices even on older SDL versions (without `SDL_HINT_AUDIO_INCLUDE_MONITORS`), by simply setting the monitor device as default. Quite useful, since most systems still have old SDL versions with no `SDL_HINT_AUDIO_INCLUDE_MONITORS` support.

This PR adds a `System default capture device` to the list of devices selectable by `Ctrl-I`. So `projectMSDL` starts with this default, but the user can select a specific device via `Ctrl-I`, as usual.

I hope you will consider such a PR, I think it makes `projectMSDL` easier to use.

PS. Thanks for this truly amazing tool!

### DIFF
--- a/src/projectM-sdl/audioCapture.cpp
+++ b/src/projectM-sdl/audioCapture.cpp
@@ -17,7 +17,9 @@ int projectMSDL::initAudioInput() {
     want.callback = projectMSDL::audioInputCallbackF32;
     want.userdata = this;
 
-    audioDeviceID = SDL_OpenAudioDevice(SDL_GetAudioDeviceName(selectedAudioDevice, true), true, &want, &have, 0);
+    // index -1 means "system deafult", which is used if we pass deviceName == NULL
+    const char *deviceName = selectedAudioDevice == -1 ? NULL : SDL_GetAudioDeviceName(selectedAudioDevice, true);
+    audioDeviceID = SDL_OpenAudioDevice(deviceName, true, &want, &have, 0);
 
     if (audioDeviceID == 0) {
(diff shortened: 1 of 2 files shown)


revmischa (approved):
Awesome thx!

kblaschke:
Good idea! Should make it easier for most users to specify a default device for now.

Proper audio device selection in the standalone app is really something that needs a major overhaul though. At least, it should remember the last selected device (if the user changed it) and use it on next startuip if it is available. Then there is WASAPI, which supports loopback devices, but these can't be used via SDL's WASAPI audio driver. As long as you're only playing audio or require capturing from a microphone, all is well, but recording desktop audio is pure horror. On macOS, it doesn't work at all due to OS limitations.

PR #584, opened 2022-02-13
Removed museum directory as it's now in its own repo.

kblaschke:
New repo: https://github.com/projectM-visualizer/Museum


kblaschke:
Will create a new PR with all outsourced code removed at once.

PR #585, opened 2022-02-14
Removed ancient, never finished libvisual/ALSA standalone application.

kblaschke:
Status and purpose are unknown, if someone really wants to continue working on it, it's still in the Git history and can be outsourced into a separate repo.


kblaschke:
Will create a new PR with all outsourced code removed at once.

PR #587, opened 2022-02-18
Some CMake build fixes, mostly affecting Windows builds.

kblaschke:
- Got rid of the WORD_BITS usage in omptl_tools.h, which is a macro that is only defined on UNIX platforms, but not MSVC.
- Removed export(PACKAGE) call, as it's been disabled/deprecated since CMake 3.15 and puts a half-baked config into the registry. Better install and use CMAKE_PREFIX_PATH.
- Removed set_and_check commands in the package config file, added missing include for the target definition file.
- Added "lib" prefix to the static library file on Windows as it will otherwise have the exact same name (projectM.lib) as the shared library exports file.
- Enable shared library building on Windows by default as on all other platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,7 @@ endif()
 
 # Feature options, including dependencies.
 option(ENABLE_STATIC_LIB "Build and install libprojectM as a static library" ON)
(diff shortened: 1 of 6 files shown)


revmischa (commented):
I've nothing to add here

PR #595, opened 2022-03-16
Refactor PCM code (chunk 1)

ErikPrantare:
First chunk of refactoring for PCM, more to come.

Please tell if you would prefer larger or smaller PRs for these kinds of
changes.

### DIFF
--- a/src/libprojectM/PCM.cpp
+++ b/src/libprojectM/PCM.cpp
@@ -86,7 +86,7 @@ public:
 };
 
 
-PCM::PCM() : start(0), newsamples(0)
+PCM::PCM()
 {
     leveler = new AutoLevel();
 
--- a/src/libprojectM/PCM.cpp
+++ b/src/libprojectM/PCM.cpp
@@ -210,6 +210,29 @@ void PCM::addPCM8_512(const unsigned char PCMdata[2][512])
 }
 
 
+template<size_t aSize, size_t ipSize, size_t wSize>
+void rdft(
+    int const isgn,
+    std::array<double, aSize>& a,
+    std::array<int, ipSize>& ip,
+    std::array<double, wSize>& w)
+{
+    // per rdft() documentation
+    //    length of ip >= 2+sqrt(n/2) and length of w == n/2
+    //    n: length of a, power of 2, n >= 2
+    // see fftsg.cpp
+    static_assert((ipSize - 2) * (ipSize - 2) >= aSize,
+                  "rdft invariant not preserved: length of ip >= 2+sqrt(n/2)");
+    static_assert(2 * wSize == aSize,
+                  "rdft invariant not preserved: length of w == n/2");
+    static_assert(aSize >= 2,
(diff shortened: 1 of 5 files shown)


ErikPrantare:
Seems like many of the changes from #594 ended up showing here too, changes are from [5512e96: Make values constexpr where possible](https://github.com/projectM-visualizer/projectm/pull/595/commits/5512e965af8868e892f3b99777decd9e6e5772a0) onwards.

kblaschke:
> Seems like many of the changes from #594 ended up showing here too

Sure, as the PR shows all differences between your branch and the target, so it includes all commits from your other PR, as these commits are part of this branch, too. These will vanish though as soon as the other PR is merged. Also don't forget to rebase this branch again onto the other one  in #594 after the squash, so there won't be duplicated commits.

ErikPrantare:
Will do, and will make sure to organize my branches better in the future.

kblaschke:
Can always get messy with multiple PRs open, sometimes there's simply no "good" way. Never be afraid to rebase and force-push, as long as the branch isn't merged, you can change everything at any time.

ErikPrantare:
I think I've gotten the history correct now

src/libprojectM/PCM.cpp
```diff
@@ -125,7 +125,7 @@ void PCM::addPCMfloat(const float* PCMdata, size_t samples)
         size_t j=(i+start)%maxsamples;
         a=pcmL[j] = PCMdata[i];
         pcmR[j] = PCMdata[i];
-        sum += fabs(a);
+        sum += std::abs(a);
```
kblaschke, line 85: Suggest placing a `using std::abs;` at the top of the implementation file to make the calls a bit less verbose (same for `std::max`).

ErikPrantare: I'm not a fan of file-wide using declarations when it comes to libraries, but if you think it'd be more readable maybe they can be put locally? Always `std::abs` would be consistent, if we have exceptions to that I think we should indicate that as locally as possible. Personally my brain just filters out the `std::`, like lipsers filter out all the parentheses when they code :^)

kblaschke: Can just keep it that way. Would be more of an issue with a class in a nested namespace, like the Poco library often has, but in this case `using namespace` is a good solution if these classes are heavily used.

src/libprojectM/PCM.cpp
```diff
@@ -24,380 +24,280 @@
...
+                  "rdft invariant not preserved: n is power of two");
+    rdft(aSize, isgn, a.data(), ip.data(), w.data());
 }
 
 
 // puts sound data requested at provided pointer
 //
 // samples is number of PCM samples to return
-// smoothing is the smoothing coefficient
+// smoothing does nothing
```
kblaschke, line 210: Exactly, this should be removed.

kblaschke (approved):

PR #596, opened 2022-03-17
Add .clang-tidy

ErikPrantare:
Add .clang-tidy.

This simply contains my personal preferences.
I like m_ as a prefix for private members,
it can make code a lot clearer ime to quickly see which variables are members.

The specific casing otherwise I don't have much opinion on,
just that we need to be more consistent.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,34 @@
+Checks: '
+-*,
+
+readability-*,
+-readability-named-parameter,
+-readability-magic-numbers,
+-readability-make-member-function-const,
+
+cppcoreguidelines-*,
+-cppcoreguidelines-owning-memory,
+-cppcoreguidelines-pro-type-union-access,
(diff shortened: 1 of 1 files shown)


ErikPrantare:
One other thing I've also started doing in general is treating acronyms as one word. So depending on the casing, a function could be named `send_http` or `sendHttp`, rather than `sendHTTP`. It is, imo, much easier to read correctly than the alternative, and avoids some ambiguities. Especially with acronyms after eachother.

kblaschke:
Purists would say the `m` is clutter, but your point of getting into trouble with reserved names in libraries starting with underscores is probably a valid thing. Regarding the casing, I think lower camel case for local and member variables and using Pascal case for functions is a great way to distinguish between those two. Also most C macros use upper snake case, they also have distinct casing.

I'm fine with the Microsoft'ish `m_` prefix.

PR #598, opened 2022-03-19
Some additional clang-tidy casing clarifications.

kblaschke:
Would only keep lower snake case for C-style functions, as it's basically the standard. Class names & member functions using Pascal/upper camel case, members lower camel case. Macros upper snake case. Also using `m_` prefix for protected members as well, but not public members.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -22,11 +22,41 @@ misc-*'
 
 CheckOptions:
   - key: readability-identifier-naming.VariableCase
-    value: lower_case
+    value: camelBack
+  - key: readability-identifier-naming.LocalVariableCase
+    value: camelBack
+  - key: readability-identifier-naming.ParameterCase
+    value: camelBack
   - key: readability-identifier-naming.ClassCase
-    value: lower_case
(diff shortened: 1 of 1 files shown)


ErikPrantare:
I think maybe it's a good idea to let functions and methods have the same casing (both are functions after all). And also variables and functions, as that promotes the functional view of functions as variables. Especially when working with lambdas, assigning a lambda to a local variable is not different from defining a function with the name of the variable. I think camelCase would be a good choice.

If there is a way to flag for clang-tidy to ignore casing for certain functions, we could still use snake_case for the outwards facing API. I think also there is an option to exclude functions matching a regex from case checking, so we could exclude functions starting with `projectm_`/`projectM`. Personally never got it to work, but may have been fixed by now.

revmischa:
My opinion is camelCase or snake_case is fine for C, but it should be consistent. We use SDL and OpenGL, maybe if they use the same scheme we can adopt that?

ErikPrantare:
I think this looks good

revmischa (approved):

PR #600, opened 2022-03-24
Refactor Pcm (chunk 2)

ErikPrantare:
There is a bit of a mix between doubles and floats,
e.g. m_freqL vs m_spectrumL.
Is there some fundamental reason we wouldn't want to use doubles across the
board?

This depends on #595

### DIFF
--- a/src/libprojectM/PCM.hpp
+++ b/src/libprojectM/PCM.hpp
@@ -36,7 +36,7 @@
 
 // FFT_LENGTH is number of magnitude values available from getSpectrum().
 // Internally this is generated using 2xFFT_LENGTH samples per channel.
-size_t constexpr FFT_LENGTH = 512;
+size_t constexpr fftLength = 512;
 class Test;
 
 enum CHANNEL
@@ -47,33 +47,33 @@ enum CHANNEL
     CHANNEL_1 = 1
 };
 
-class PCM
+class Pcm
 {
 public:
     /* maximum number of sound samples that are actually stored. */
     static constexpr size_t maxSamples = 2048;
 
-    void addPCMfloat(const float* PCMdata, size_t samples);
-    void addPCMfloat_2ch(const float* PCMdata, size_t count);
-    void addPCM16(const int16_t[2][512]);
-    void addPCM16Data(const int16_t* pcm_data, size_t samples);
-    void addPCM8(const uint8_t[2][1024]);
-    void addPCM8_512(const uint8_t[2][512]);
+    void AddPcmFloat(const float* pcmData, size_t samples);
+    void AddPcmFloat2Ch(const float* pcmData, size_t count);
+    void AddPcm16(const int16_t pcmData[2][512]);
+    void AddPcm16Data(const int16_t* pcmData, size_t samples);
+    void AddPcm8(const uint8_t pcmData[2][1024]);
+    void AddPcm8(const uint8_t pcmData[2][512]);
 
     /**
      * PCM data
      * smoothing does nothing
      * The returned data will 'wrap' if more than maxsamples are requested.
      */
-    void getPCM(float* data, CHANNEL channel, size_t samples, float smoothing);
+    void GetPcm(float* data, CHANNEL channel, size_t samples, float smoothing);
 
     /** Spectrum data
      * smoothing does nothing
      * The returned data will be zero padded if more than FFT_LENGTH values are requested
      */
-    void getSpectrum(float* data, CHANNEL channel, size_t samples, float smoothing);
+    void GetSpectrum(float* data, CHANNEL channel, size_t samples, float smoothing);
 
-    static Test* test();
+    static Test* MakeTest();
 
 private:
     // mem-usage:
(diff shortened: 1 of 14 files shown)


ErikPrantare:
Might remove the macro in the future, only causes grief.

kblaschke:
> Might remove the macro in the future, only causes grief.

In my parser rewrite branch, I've added a proper GTest-based unit test project in /tests, into which we should move all the test code and make it a bit more fine-grained as it is now. Might leave it as-is now since this affects a few other classes as well, could refactor this in a separate branch in one go.

Having PCM data as floats is key, as all calculations down the road are done with floating-point numbers. Doing the conversion only once if samples are provided in an int format saves us a bunch of calculation time.

src/libprojectM/PCM.hpp
```diff
@@ -76,13 +76,16 @@ class Pcm
     static auto MakeTest() -> Test*;
 
 private:
-    template<int signalAmplitude, int signalOffset, class SampleType>
-    void AddPcm(
-        const SampleType* pcmData,
+    // CPP20: Could use a struct for the first 5 params to clarify on call site
+    // together with designated initializers
```
kblaschke, line 78: As noted above, stride/offset should not be a compile-time argument, as it would not allow for passing 6-channel audio buffers to projectM, requiring apps to first reorganize their memory, which is not necessary as we copy the values into new buffers anyway.

kblaschke:
Part 1 is merged, so you can also rebase the branch on latest master to get rid of those 29 commits from the first batch.

src/libprojectM/PCM.hpp
revmischa: I don't think anyone should be passing 6-channel audio buffers into projectM. There's no good reason to do that. We should accept at most two channels.

kblaschke: As discussed on Discord, almost all audio APIs use interleaved streams, so only allowing two-channel data to be passed in would mean to reorganize the data stream layout before passing it to projectM to [LR][LR]... format, just for projectM to copy each field into a separate array. By allowing any number of channels and specifying the (possibly optional) channel offsets in each frame, integrating apps don't need this additional step, and projectM can simply pick the proper values from the buffer and ignore the rest, without any additional cost for projectM _or_ the integrating app.

That said, most if not all audio APIs have a fixed channel order, which, for 8 channels, is: front left, front right, center, bass, rear left, rear right, side left, side right. Might be a bit different for unconventional formats like 2.1, but front L/R are always the first two and in this exact order. So we could provide two functions, one with default channel offsets (0 for left, 1 for right) and thus less parameters, and another one with user-defined offsets.

If applications need to downmix audio from 5.1 to stereo, for example, authors can still do that if required. But the API should make it seasy to pass the exact data the audio API provides as a pointer to projectM, and let the library pick the correct bytes from the buffer and perform any data type conversions like uint8->float.

kblaschke:
> Is there some fundamental reason we wouldn't want to use doubles across the board?

As of today, where most if not all platforms potentially running projectM use 64-bit architectures, I'd say it will not make any difference at all. Internally, processors will perform all calculations with 64 bit registers anyway, so in my opinion, doubles would be fine inside the audio buffer class, but only internally.

From a simple precision perspective, I think floats already have enough precision and value range for all our use cases, so using doubles will simply increase memory usage a tiny bit and require more (potentially slow and ugly) downcasts to floats, but not give any real advantage. In a high-quality DSP app, I'd say "sure, make it as precise as possible", but since we're just roughly visualizing the waveforms, we do not require a precision to the tenth decimal or real values in the billions while still retaining a good decimal precision.

ErikPrantare:
That's fair. Is there a reason to use doubles elsewhere? I guess maybe it's simply because `rdft` expects doubles.

src/libprojectM/PCM.hpp
ErikPrantare: I think such an addition might be good to make its own task. I can see the value in it, but it feels a bit outside the scope of this one. The current API should be good for most uses as you said, and in a while we'll perhaps too be able to snoop how people make use of the library. At that point we'd also get some better data on where we could take API.

kblaschke: Sure, no need to fix everything in this PR, just make it MVP.

mbellew: It's awesome that this is getting some love.  I've messed with these files, but wasn't brave enough to mess with the public APIs.  My one comment is that initial caps for method names seems very strange to me.  I've never worked a project that did that in C or Java, but maybe it's a common usage elsewhere...

Since I have messes with these files, here are some thoughts that may not have been shared:

I don't see any benefit to using doubles for PCM data.  32bit float is already very high precision for sound.  I've been annoyed by the conversion to double for rdft in the past.  I solved that in a different project (microprocessor) by just search/replace double->float.  That worked great.

One problem we haven't solved is that sometimes the audio data comes in too fast and we drop a lot of PCM data on the floor.  I think this is the cause of some complaints of "lag" when in fact we are actually way ahead of the "play head" rather than behind.  To fix that we'd need to use a much larger buffer and make use of the APIs like pa_stream_get_timing_info() (or equivalent in other libraries).  It's not a that much memory, but we need to buffer 0.5s of sound double the buffer size for no benefit seems kinda wasteful.

ErikPrantare:
@mbellew 
> My one comment is that initial caps for method names seems very strange to me. I've never worked a project that did that in C or Java, but maybe it's a common usage elsewhere...

I agree, I have seen `camelCase` most commonly for methods. I'm just following the rules laid out in `.clang-tidy` that was just added to the project 😁

>I don't see any benefit to using doubles for PCM data. 32bit float is already very high precision for sound. I've been annoyed by the conversion to double for rdft in the past. I solved that in a different project (microprocessor) by just search/replace double->float. That worked great.

This sounds like a better solution if we don't need the extra precision.

>One problem we haven't solved is that sometimes the audio data comes in too fast and we drop a lot of PCM data on the floor.

Like, if more than 2k samples (max amount of samples, ~1/20s) are entered in the same frame?

src/libprojectM/PCM.hpp
kblaschke: I've seen issues with audio data buffering. If new data arrives in larger batches and the outside program just slams in the whole buffer, projectM throws away many samples due to the internal 2k ring buffer. This can result in multiple frames rendering the same wave data, beat detection not working properly and other weirdnesses.

I can't really think of a general solution for that problem we could implement in the library as the use case might be quite different. Audio players integrating projectM might provide samples way in advance to the library, which is also something Milkdrop has to take care of: see pluginshell.cpp, it asks Winamp for the actual delay to set the buffer size, then use a set number of samples per rendered frame to match the samplerate and player delay to use the exact samples currently playing. Then for standalone apps using recording devices you'll want a buffer that is more or less immediate, and fill it with the most current data before rendering the next frame to keep the visuals in sync as best as possible.

I think a configurable internal buffer size might be a good thing, so the external app can decide how many samples it should fill in advance, which can be calculated using sample rate times the expected delay.

mbellew: Thanks for clarifying my comments Kai.  Of course, as currently implemented the buffering problem is on the host side.  I had pondered adding a way to a) buffer more and b) add a way for the app to communicate position/latency information to PCM.  However, requiring the app to do isn't unreasonable.  BTW, the scenario that drove me nuts until I figured this out was Spotify on Linux which would cause the SDL demo app to send 0.5s of data at a time.

ErikPrantare:
Rebased on top of master

kblaschke (approved):

kblaschke:
Looks great so far from my side. Any further API changes and improvements are TBD and we can sort these out while improving the library C++ "entry point".

If there's no changes to be added, I'd merge this set of commits.

ErikPrantare:
No more changes on my side

PR #602, opened 2022-03-27
Parser fixes to support multiple equations/assignments per line.

kblaschke:
Just a few preliminary changes to fix the issue with multiple equations in a single line in shape and wave equations, e.g. "x=1; y=2;".

This PR is only here to make a large number of otherwise broken presets render correctly, as I'm already working on a parser rewrite, as there are so many missing features in it which aren't easy to integrate in the current parser, e.g. having `megabuf(expr) = value;` or `loop(num, expr; expr; ...)`, while the parser currently only supports variables on the left-hand side of a simple equal sign.

### DIFF
--- a/src/libprojectM/MilkdropPresetFactory/Parser.cpp
+++ b/src/libprojectM/MilkdropPresetFactory/Parser.cpp
@@ -490,26 +490,34 @@ int Parser::parse_line(std::istream &  fs, MilkdropPreset * preset)
     /* Per frame equation case */
     if (!strncmp(eqn_string, PER_FRAME_STRING, PER_FRAME_STRING_LENGTH))
     {
-	tokenWrapAroundEnabled = true;
-      /* Sometimes per frame equations are implicitly defined without the
-      per_frame_ prefix. This informs the parser that one could follow */
-      line_mode = PER_FRAME_LINE_MODE;
+        tokenWrapAroundEnabled = true;
+        /* Sometimes per frame equations are implicitly defined without the
+        per_frame_ prefix. This informs the parser that one could follow */
+        line_mode = PER_FRAME_LINE_MODE;
 
-      //if (PARSE_DEBUG) printf("parse_line: per frame equation found...(LINE %d)\n", line_count);
+        //if (PARSE_DEBUG) printf("parse_line: per frame equation found...(LINE %d)\n", line_count);
 
-      /* Parse the per frame equation */
-      if ((per_frame_eqn = parse_per_frame_eqn(fs, ++per_frame_eqn_count, preset)) == NULL)
-      {
-        if (PARSE_DEBUG) printf("parse_line: per frame equation parsing failed (LINE %d)\n", line_count);
-        tokenWrapAroundEnabled = false;
-        return PROJECTM_PARSE_ERROR;
-      }
+        /* Parse the per frame equation */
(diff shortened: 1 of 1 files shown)


src/libprojectM/MilkdropPresetFactory/Parser.cpp
```diff
@@ -1355,63 +1363,93 @@ int Parser::parse_float(std::istream &  fs, float * float_ptr)
...
+    {
+        return NULL;
+    }
+    if (PARSE_DEBUG)
+    {
+        std::cerr << "parse_per_frame_eqn: parameter \"" << param->name << "\" retrieved (LINE" << line_count << ")"
+                  << std::endl;
+    }
+    /* Make sure parameter is writable */
+    if (param->flags & P_FLAG_READONLY)
```
revmischa, line 1402: This `if` is a bit different from the one below

src/libprojectM/MilkdropPresetFactory/Parser.hpp
```diff
@@ -29,7 +29,7 @@
 #ifndef _PARSER_H
 #define _PARSER_H
 //#define PARSE_DEBUG 2
-#define PARSE_DEBUG 0
+#define PARSE_DEBUG 2
```
revmischa, line 32: want to commit this?

revmischa (approved):
Looks good I guess? hard to tell lol

mbellew:
Consider adding a simple test preset?  These are handy for testing new parser too.

src/libprojectM/MilkdropPresetFactory/Parser.cpp
kblaschke: Hum... yeah, I did not try to understand this code at all. This function was just reformatted and I replaced a few `NULL`s with `nullptr`. This line was line 1380 previously, it just moved down a bit.

I think you can see here why it's just better to rewrite this file reader from scratch than trying to "fix" it. Reimplementation will first read all the lines into a map of key/value pairs, then use a key prefix to retrieve init values or code blocks, more or less the same as Milkdrop does it. The existing parser here is just way too complex.

src/libprojectM/MilkdropPresetFactory/Parser.hpp
kblaschke: Ah, good catch! This should rather be done as a CMake flag, and/or only be enabled in debug builds.

kblaschke:
> Consider adding a simple test preset? These are handy for testing new parser too.

Definitely. [New parser [will be fully unit tested](https://github.com/kblaschke/projectm/blob/9fdebb194692d7c7011f401efcae53bd37d57f72/tests/FileParserTest.cpp), with plenty of presets and equation code to test each feature and function. For the old parser, its simply not worth adding any test cases. For the equation part, I'd personally trust unit tests more than running presets and check the visuals because you don't have any comparison to know how it _should_ look like.

mbellew:
> 

Totally agree.  I would have suggested unit test but i don't know where the unit test app migrated to.

kblaschke:
> Totally agree. I would have suggested unit test but i don't know where the unit test app migrated to.

It's still there, in the top-level "tests" dir, but it'll be replaced with a proper GTest-based framework. Will move all existing "TesstRunner"-based test cases there where it makes sense to keep them.

kblaschke:
I'll just merge this one, as it's a quick, temporary hack to get a few more things working until the new parser is ready.

PR #603, opened 2022-03-27
A truckload of preset rendering fixes.

kblaschke:
As the title says, this set of commits fixes quite a number of rendering issues, mostly shape and wave problems, but also some shader-related stuff. Here's a rundown on what's included and some things that are not.

Note: The timer and parser fix commits are in separate PRs, so ignore these.

What is fixed:
- Shaders: Generate better noise textures for use in shaders, as the previous textures were missing a lot of entropy. The new algorithm is now the same as the one Milkdrop uses.
- Shaders: Fixed texture wrap settings for predefined textures such as noise and blur.
- Shapes and waveforms: Replaced thick line drawing code previously using OpenGL's line width setting with Milkdrop's 1px width, 2x2 shifting logic. Enables anti-aliased lines and creates a crispier result, matching what Milkdrop renders. This also applies for single points, which are now rendered as 2x2 dots in thick mode, which wasn't possible before.
- Waveforms: Fixed an issue that internally, waveform modes were starting with 1, while preset sytax starts with 0, effectively shifting every mode one to the right and rendering only static memory garbage for wavemode 0 (e.g. looking like a non-moving shape).
- Waveforms: Rewrote the whole waveform math to match Milkdrop's code.
- Waveforms: smoothing was delegated to the PCM class, but it didn't work and produced wrong results. Replaced it with Milkdrop's smoothing logic by duplicating the waveform vertices and interpolating between them depending on the smoothness factor. This also adds more drawn points to point-only waveforms, making them look more dense.
- Shapes: Instanced shapes and per-instance code were not supported at all. Added the `num_inst` parameter and per-instance code execution to draw all configured instances.
- Shapes: Rewrote the shape vertex math, simplified the code a bit and fixed aspect ratio correction to some degree. The rewrite also includes proper texture coordinate calculation for textured shapes. Removed the unsupported `image_url` parameter which was probably meant to display a specific texture on the shape.

What remains:
- Shader translation issues. There are still some shaders either running super slowly or producing broken output. This will require in-depth analysis and fixing of the HLSLParser subproject.
- Aspect correction is still broken in some places, e.g. visible in shape rendering. In Milkdrop, aspect correction is always Y-major, so everything should scale up/down with Y resolution, and just stay fixed in the center if X resolution is changed.
- Shape positioning is still a bit off, which can be seen prominently in `amandio c - living boxes 7` for example.
- All equation-related issues. There's a bunch of missing features, basically everything introduced in `ns-eel2` like megabuf and loop, plus a number of additional operators like `+=` and `<`. See my [preset parser rewrite branch](https://github.com/kblaschke/projectm/tree/rewrite-preset-parser) for progress on that front.

I hope I've got everything listed, but there might be some additional, smaller fixes in the commits. All in all, this is a _huge_ step forward to make projectM compatible with Milkdrop 2.x and improve the quality of the rendered visuals by orders of magnitude. Just compare it side-by-side to previous versions, e.g. on Steam, and the "Cream of the Crop" preset pack to see the difference.


src/libprojectM/Renderer/MilkdropWaveform.cpp
```diff
@@ -4,393 +4,613 @@
...
-        else
-            glDrawArrays(GL_LINE_STRIP, 0, samples);
+        // If thick outline is used, draw the shape four times with slight offsets
+        // (top left, top right, bottom right, bottom left).
+        for (auto iteration = 0; iteration < iterations; iteration++)
+        {
+            switch (iteration)
+            {
+                case 0:
+                default:
```
revmischa, line 99: It looks so very wrong to put the `default` at the top...
Do we expect it to always have four iterations? maybe it should blow up if more?

src/libprojectM/Renderer/MilkdropWaveform.cpp
```diff
@@ -4,393 +4,613 @@
...
+    delete[] wave1Vertices;
+    wave1Vertices = nullptr;
 
-            samples = 512;
+    delete[] wave2Vertices;
+    wave2Vertices = nullptr;
 
-			float inv_nverts_minus_one = 1.0f/(float)(samples);
+    // NOTE: Buffer size is always 512 samples, waveform points 480 or less since some waveforms use a positive offset!
+    float pcmDataL[512]{ 0.0f };
```
revmischa, line 248: Can we use a constant for the 512 please?

src/libprojectM/Renderer/MilkdropWaveform.cpp
```diff
@@ -4,393 +4,613 @@
...
+                        case 3:
+                            if (edgeY[i] < -1.1f)
+                            {
+                                t = (-1.1f - edgeY[1 - i]) / (edgeY[i] - edgeY[1 - i]);
+                                clip = true;
+                            }
+                            break;
+
+                        default:
+                            break;
```
revmischa, line 507: idk shouldn't we blow up here?

src/libprojectM/Renderer/MilkdropWaveform.cpp
```diff
@@ -4,393 +4,613 @@
...
+            float angle2 = atan2f(dy, dx);
+            float perpetualDX = cosf(angle2 + PI * 0.5f);
+            float perpetualDY = sinf(angle2 + PI * 0.5f);
+
+            if (mode == MilkdropWaveformMode::Line)
+            {
+                for (int i = 0; i < samples; i++)
+                {
+                    wave1Vertices[i].x =
+                        edgeX[0] + dx * static_cast<float>(i) + perpetualDX * 0.25f * pcmDataL[i + sampleOffset];
```
revmischa, line 532: Just curious why we only use pcm left data here?

revmischa (approved):

src/libprojectM/MilkdropPresetFactory/CustomWave.hpp
```diff
@@ -18,109 +18,99 @@
...
+#include "MilkdropPresetFactory/Param.hpp"
+#include "MilkdropPresetFactory/PerFrameEqn.hpp"
 
 #include <map>
+#include <vector>
 
 class CustomWave : public Waveform
 {
 public:
+     CustomWave() = delete;
```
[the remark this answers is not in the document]
kblaschke: Deletes the default constructor, otherwise the compiler might generate a default one. It's a C++11 feature:

https://en.cppreference.com/w/cpp/language/function#Deleted_functions

src/libprojectM/MilkdropPresetFactory/Parser.hpp
```diff
@@ -29,7 +29,7 @@
 #ifndef _PARSER_H
 #define _PARSER_H
 //#define PARSE_DEBUG 2
-#define PARSE_DEBUG 0
+#define PARSE_DEBUG 2
```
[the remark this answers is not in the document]
kblaschke: Yeah, it's from the old parser fix commit, that's still in here. Will rebase the branch on master after merging the other to PRs and drop these commits in this branch before doing so.

src/libprojectM/MilkdropPresetFactory/CustomWave.cpp
```diff
@@ -43,469 +43,165 @@
...
+        if (CUSTOM_WAVE_DEBUG)
+        {
+            printf("add_per_point_eqn: failed to allocate a new parameter!\n");
+        }
+        return PROJECTM_FAILURE;
+    }
 
-  per_point_eqn_tree.push_back(per_point_eqn);
+    /* Get largest index in the tree */
+    per_point_eqn_tree.push_back(new PerPointEqn(static_cast<int>(per_point_eqn_tree.size()), param, gen_expr));
```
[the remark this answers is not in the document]
kblaschke: This indeed makes no sense at all. Should use the comment from line 474 in the old version.

src/libprojectM/Renderer/MilkdropWaveform.cpp
kblaschke: The `iterations` value is either 1 or 4, this is hard-coded a few lines above. I'll change it to a const int to make it even clearer, and just drop the default, as the compiler will know about the possible values and not emit a note or warning.

src/libprojectM/Renderer/MilkdropWaveform.cpp
kblaschke: Sure thing will add a constexpr at the top of the function and also turn this into `std::array<float,n>`.

src/libprojectM/Renderer/MilkdropWaveform.cpp
kblaschke: Well, no. But remove the default, as it's not needed. Compiler will most probably unroll the loop in optimized builds.

src/libprojectM/Renderer/MilkdropWaveform.cpp
kblaschke: It's the "single line" waveform mode. Milkdrop also only uses the left channel here, so I just did the same for maximum compatibility.

Original line of code is here, note the `fL[i + sample_offset]`, which is the equivalent of `pcmData[]`: https://github.com/projectM-visualizer/milkdrop2/blob/f05b0d811a87a17c4624170c26c93bac39b05bde/src/vis_milk2/milkdropfs.cpp#L3206

kblaschke:
Fixed the above commetns and also remove the two commits from the previous RPs to cause less confusion.

PR #606, opened 2022-03-31
Cleanup projectM class

kblaschke:
Cleaned up and reformatted the main projectM class/header. Not done yet, just putting it up as draft until I'm finished.

Many changes to make the class more manageable for future refactorings and further cleanups:
- Renamed class from projectM to ProjectM
- Swapped pthread API with STL's concurrency library, so it will now also work on Windows. STL is even less verbose, making the code easier to read.
- Enabled threading for all platforms including emscripten.
- Reorganized the header into function groups, used Doxygen comments to emphasize.
- Added Doxygen comments to most members.
- Fixed names and casing of all members.
- Removed the font file settings.
- Removed unused code and members.
- Made all member variables private. Any access required by future features can be implemented via getters.
- Use trailing return types for all members.
- Replaced all raw pointers with std::unique_ptr for automatic memory management. Down the call hierarchy, there are still a lot of raw pointers stored. Might be better to use std::shared_ptr for some.
- Renamed many functions and parameters to be more consistent and speaking.


ErikPrantare:
Is there a reason for putting all the changes in a single commit? Personally I find it easier to review many small individual changes, rather than one big compound one.

kblaschke:
> Is there a reason for putting all the changes in a single commit? Personally I find it easier to review many small individual changes, rather than one big compound one.

It's really hard to work through so many changes one by one, as each commit would have almost the full header changed, and mus of the source file.
If you find this too much, I can of course revert and start over again, then try to change only a few lines and a single thing each time. Will take a few more days or even a fortnight then though to redo the changes.

ErikPrantare:
No need to start over again, just wondered for the future. When I make changes, I usually focus on one thing at a time, e.g. "Give identifiers correct casing", "Use smart pointers instead of raw pointers", "Use trailing return types", et.c.. And then after each such chunk I commit. Sometimes it's not perfect (I forgot something in some earlier commit, et.c.), but I think it makes review a bit simpler.

PR #607, opened 2022-04-01
Replaced pthread API with STL's concurrency library.

kblaschke:
Multithreading will now also work on Windows while the STL variant is even less verbose, making the code easier to read. Thus enabled threading for all platforms, including emscripten.

### DIFF
--- a/src/libprojectM/projectM.cpp
+++ b/src/libprojectM/projectM.cpp
@@ -329,7 +329,7 @@ Pipeline* projectM::renderFrameOnlyPass1(
     Pipeline* pPipeline) /*pPipeline is a pointer to a Pipeline for use in pass 2. returns the pointer if it was used, else returns NULL */
 {
 #if USE_THREADS
-    std::lock_guard<std::mutex> guard(m_presetSwitchMutex);
+    std::lock_guard<std::recursive_mutex> guard(m_presetSwitchMutex);
 #endif
 
 #ifdef DEBUG
@@ -1029,7 +1029,7 @@ std::unique_ptr<Preset> projectM::switchToCurrentPreset() {
   std::unique_ptr<Preset> new_preset;
 
 #if USE_THREADS
-    std::lock_guard<std::mutex> guard(m_presetSwitchMutex);
+    std::lock_guard<std::recursive_mutex> guard(m_presetSwitchMutex);
 #endif
 
   try {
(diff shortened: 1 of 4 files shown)


src/libprojectM/projectM.hpp
```diff
@@ -63,6 +63,9 @@
 #include "event.h"
 #include "fatal.h"
 #include <vector>
+#include <mutex>
+#include <thread>
+#include "BackgroundWorker.h"
```
kblaschke, line 68: Will reorder that mess in the projectM class refactoring.

src/libprojectM/projectM.cpp
```diff
@@ -565,20 +537,9 @@ void projectM::projectM_init ( int gx, int gy, int fps, int texsize, int width,
...
-
-    worker_sync.reset();
-    if (pthread_create(&thread, NULL, thread_callback, this) != 0)
-    {
-
-        std::cerr << "[projectM] failed to allocate a thread! try building with option USE_THREADS turned off" << std::endl;;
-        exit(EXIT_FAILURE);
-    }
+    m_workerSync.Reset();
+    m_workerThread = std::thread(&projectM::ThreadWorker, this);
```
kblaschke, line 542: We may want to catch `std::system_error` in case the thread could not be started. Calling `exit()` in a library is bad habit, so having a generic exception handler in the C wrapper's `projectm_create*()` functions might be a better option, returning nullptr if construction fails. Then again, throwing in the constructor may cause memory leaks...

src/libprojectM/projectM.hpp
```diff
@@ -407,7 +407,7 @@ class projectM
 #if USE_THREADS
     void ThreadWorker();
 
-    mutable std::mutex m_presetSwitchMutex; //!< Mutex for locking preset switching while rendering and vice versa.
+    mutable std::recursive_mutex m_presetSwitchMutex; //!< Mutex for locking preset switching while rendering and vice versa.
```
kblaschke, line 410: That'd otherwise cause a deadlock, as both functions are also called in the same stack and thread. I think this lock is just there to prevent/postpone preset switching while rendering a frame at the same time.

src/libprojectM/projectM.cpp
ErikPrantare: Throwing in the constructor calls the destructors, so if we use `unique_ptr` for resource management leaking shouldn't be an issue if we throw in the constructor.

PR #609, opened 2022-04-04
Reformatted the main projectM class.

kblaschke:
As requested, will redo the projectM class refactoring. Since it makes sense to start work on a clean, formatted codebase, this PR only contains code formatting changes done by clang-format.

I can't see any way to break down formatting into smaller parts, so to check if there are other changes, reviewers could take the previous commit, apply clang-format yourself to the files using the `.clang-format` file in the project root and then compare with this PR.


revmischa (approved):
Cool maybe do it to all files?

PR #610, opened 2022-04-04
Refactor RenderItemDistanceMetric

ErikPrantare:
Done my best to distill this one to just be the minimal amount of code that
does the same thing as before.
Will have to clean up RenderItemMatcher and RenderItemMergeFunction too,
they seem pretty connected to this one.

Hopefully I didn't mess with the logic, it was quite convoluted 😅


ErikPrantare:
This may be completely unnecessary, I just noticed that these classes aren't even used. Closing this for now.

PR #611, opened 2022-04-05
Remove unused RenderItem* files

ErikPrantare:
Was refactoring these when I noticed that they weren't even used anywhere.


kblaschke:
We might want to wait merging this until my `ProjectM` class cleanups are merged, as this way around, it'll be easier to rebase.
I can throw out any references to the deleted classes already during my refactor, so only deleting the unused files remains and there won't even be any conflicts.

ErikPrantare:
Sure, sounds good

kblaschke (approved):

PR #612, opened 2022-04-06
Refactor Common.hpp

ErikPrantare:
After having refactored this file a fair bit,
I get the feeling that it doesn't really pull its own weight.
However, I don't have enough knowledge of the codebase yet to do much about it,
so it'll live for another day :^)

Some of the platform specific macros removed I think weren't used,
but haven't tested building for those platforms.

Also added some comments for improvements that can be done if we ever bump the
C++ version.


src/libprojectM/projectM.cpp
```diff
@@ -827,7 +826,7 @@ void projectM::populatePresetMenu()
             int presetIndex = getSearchIndex(presetName);
             for (unsigned int i = 0; i < getPlaylistSize(); i++)
             { // loop over all presets
-                if (caseInsensitiveSubstringFind(getPresetName(i), renderer->searchText()) != std::string::npos)
+                if (CaseInsensitiveSubstringFind(getPresetName(i), renderer->searchText()) != std::string::npos)
```
kblaschke, line 829: Just to note: this already fails with Unicode/UTF-8 preset names (which aren't even displayed correctly), and ideally shouldn't be in the core library at all. Moving it to a helper library and making it multi-byte-aware would be a good idea later on...

src/libprojectM/Common.hpp
```diff
@@ -26,215 +26,107 @@
...
-#ifdef _MSC_VER
-#define strcasecmp(s, t) _strcmpi(s, t)
-#endif
-
-#if defined(_MSC_VER ) && !defined(EYETUNE_WINRT)
-	#pragma warning( disable : 4244 4305 4996; once : 4018 )
-	#define WIN32_LEAN_AND_MEAN
-	#define NOMINMAX
-	#include <windows.h>
-	typedef unsigned int uint;
```
kblaschke, line 44: Yeah, should've ideally used library-predefined types like uint32_t instead, if it was still being used.

src/libprojectM/Common.hpp
```diff
@@ -26,218 +26,112 @@
...
-#endif
-
```
kblaschke, line 46: Hard-coded paths with usernames... probably someone forgot to remove some test code here.

kblaschke (approved):

PR #613, opened 2022-04-09
More BeatDetect cleanup (and one bugfix)

ErikPrantare:
This started as just simply fixing a small bug from an earlier PR.
Then I got bored and tried to make the logic a bit clearer,
in addition to some cleanup to satisfy clang-tidy
and also improve const-correctness.

I still don't really see the point of the `*Att` variables.
They simply look like a lowpass filter on a given band intensity value,
which already is lowpass filtered.
If `Att` stands for attenuated, I think this usage is a bit
[misleading](https://en.wikipedia.org/wiki/Attenuator_(electronics)).
I'm not sure if these are really necessary.
In what context are they used?

Please tell if something here made the logic less clear than it previously was
:)

### DIFF
--- a/src/libprojectM/MilkdropPresetFactory/PresetFrameIO.cpp
+++ b/src/libprojectM/MilkdropPresetFactory/PresetFrameIO.cpp
@@ -24,9 +24,9 @@ void PresetInputs::update(const BeatDetect & music, const PipelineContext & cont
     this->bass = music.bass * music.beatSensitivity;
     this->mid = music.mid * music.beatSensitivity;
     this->treb = music.treb * music.beatSensitivity;
-    this->bass_att = music.bass_att * music.beatSensitivity;
-    this->mid_att = music.mid_att * music.beatSensitivity;
-    this->treb_att = music.treb_att * music.beatSensitivity;
+    this->bass_att = music.bassAtt * music.beatSensitivity;
+    this->mid_att = music.midAtt * music.beatSensitivity;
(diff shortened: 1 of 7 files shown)


kblaschke (approved):

labkey-matthewb:
I believe the Att values are exposed directly to the presets by the bass_att, mid_att, treb_att values.  I _think_ these are generally used by presets to see if the value is question is rising quickly (as opposed to just being high).

kblaschke:
> I believe the Att values are exposed directly to the presets by the bass_att, mid_att, treb_att values. I _think_ these are generally used by presets to see if the value is question is rising quickly (as opposed to just being high). @kblaschke

The `*_att` values are attenuated, e.g. dampened/smoothed values of the acutal `bass`/`mid`/`treb` values, changing less and not as strong over time. From what I've seen in presets is they're mostly used for smoother color changes, while the non-attenuated values are mostly used for "jumping" type animations like direction changes or immediate/strong color cycling.

Milkdrop's original code is here for reference, it's also not terribly sophisticated:
https://github.com/projectM-visualizer/milkdrop2/blob/master/src/vis_milk2/plugin.cpp#L9539

ErikPrantare:
@kblaschke
attenuated simply means it is multiplied by a constant less than one, yielding a weaker signal. Smoothing the signal is applying a lowpass filter (because only the low frequencies pass through, the frequencies that change slowly), so I think *_lowpass would be less misleading

PR #619, opened 2022-04-25
Cleanup projectM class (first set of changes)

kblaschke:
Going to merge the first set of code cleanups, which is basically renaming the class, methods and member variables to their proper names.

Other changes include:
- Removed some dead code and members
- Made some internally used methods private
- Made sure all function parameters are named
- Use trailing return types for all non-void functions
- Removed FPS limiting code, as this should only be done by the embedding application. Could otherwise cause unwanted lags if the app renders more than just projectM in the loop.


src/libprojectM/ProjectM.cpp
```diff
@@ -586,16 +543,16 @@ int ProjectM::initPresetTools(int gx, int gy)
 
     std::string url = (m_flags & FLAG_DISABLE_PLAYLIST_LOAD) ? std::string() : settings().presetURL;
 
-    if ((m_presetLoader = new PresetLoader(gx, gy, url)) == 0)
+    if ((m_presetLoader = std::make_unique<PresetLoader>(gx, gy, url)) == 0)
```
ErikPrantare: On failure to allocate, `make_unique` (and `new` too) will throw `std::bad_alloc`, not return `nullptr`: https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

src/libprojectM/ProjectM.cpp
```diff
@@ -586,16 +543,16 @@ int ProjectM::initPresetTools(int gx, int gy)
...
     {
         m_presetLoader = 0;
         std::cerr << "[projectM] error allocating preset loader" << std::endl;
         return PROJECTM_FAILURE;
     }
 
-    if ((m_presetChooser = new PresetChooser(*m_presetLoader, settings().softCutRatingsEnabled)) == 0)
+    if ((m_presetChooser = std::make_unique<PresetChooser>(*m_presetLoader, settings().softCutRatingsEnabled)) == 0)
```
ErikPrantare: Same here

src/libprojectM/ProjectM.cpp
```diff
@@ -607,7 +564,7 @@ int ProjectM::initPresetTools(int gx, int gy)
     // Start the iterator
     if (!m_presetPos)
```
ErikPrantare, line 545: Is init called more than once? If not, is there any reason for this check? If yes, maybe it shouldn't be called init...

src/libprojectM/ProjectM.cpp
```diff
@@ -656,39 +613,11 @@ void ProjectM::destroyPresetTools()
...
-        delete m_matcher;
-        m_matcher = NULL;
-    }
-
-    if (m_merger)
-    {
-        delete m_merger;
-        m_merger = NULL;
-    }
+    m_presetPos.reset();
```
ErikPrantare, line 595: Makes more sense to just have all resource handling logic in the ctor and dtor, then we don't have to think about resetting the member pointers.

src/libprojectM/ProjectM.hpp
```diff
@@ -343,7 +343,7 @@ class ProjectM
     };
 
 
-    inline Pcm* pcm()
+    inline Pcm& pcm()
```
ErikPrantare: `inline` should be removed from all member functions in the codebase.

src/libprojectM/ProjectM.cpp
```diff
@@ -541,7 +541,11 @@ int ProjectM::initPresetTools(int gx, int gy)
     /* Set the seed to the current time in seconds */
     srand(time(NULL));
 
-    std::string url = (m_flags & FLAG_DISABLE_PLAYLIST_LOAD) ? std::string() : settings().presetURL;
+    std::string url;
+    if ((m_flags & Flags::DisablePlaylistLoad) == Flags::DisablePlaylistLoad)
```
ErikPrantare, line 521: Could just use `!= 0`.

src/libprojectM/ProjectM.hpp
```diff
@@ -91,8 +91,15 @@ typedef enum
...
-    static const int FLAG_NONE = 0;
-    static const int FLAG_DISABLE_PLAYLIST_LOAD = 1 << 0;
+    /*
+     * Behaviour flags for the projectM instance. Currently, it's only used to prevent automatically filling
+     * the preset playlist by traversing the preset path for files.
+     */
+    enum Flags
+    {
+        None = 0, //!< No special flags.
+        DisablePlaylistLoad = 1 << 0 //!< Prevent automatic playlist loading on startup.
```
ErikPrantare, line 101: Only one flag? Perhaps we should just use a boolean instead.

src/libprojectM/ProjectM.cpp
```diff
@@ -279,14 +279,14 @@ void ProjectM::RenderFrame()
...
-    RenderFrameOnlyPass2(comboPipeline, 0, 0, 0);
+    RenderFrameOnlyPass2(comboPipeline, 0, 0);
 
     ProjectM::RenderFrameEndOnSeparatePasses(comboPipeline);
 }
 
 
 auto ProjectM::RenderFrameOnlyPass1(
-    Pipeline* pPipeline) -> Pipeline* /*pPipeline is a pointer to a Pipeline for use in pass 2. returns the pointer if it was used, else returns NULL */
+    Pipeline* pipeline) -> Pipeline* /*pPipeline is a pointer to a Pipeline for use in pass 2. returns the pointer if it was used, else returns NULL */
```
ErikPrantare, line 279: doc comment out-of-date

src/libprojectM/ProjectM.cpp
```diff
@@ -463,7 +461,7 @@ void ProjectM::Reset()
     ResetEngine();
 }
 
-void ProjectM::Initialize(int gx, int gy, int fps, int texsize, int width, int height)
+void ProjectM::Initialize(int meshResolutionX, int meshResolutionY, int targetFps, int width, int height)
```
ErikPrantare, line 444: We should be on the lookout for the letters x, y and z (and also rgba). Most of the time, these can be replaced by `glm::ivec2`, `glm::vec3` or `glm::vec4`.

src/libprojectM/ProjectM.cpp
```diff
@@ -585,6 +534,7 @@ void ProjectM::projectM_init(int gx, int gy, int fps, int texsize, int width, in
     m_timeKeeper->StartPreset();
     assert(pcm());
 
+    // ToDo: Calculate the real FPS instead
```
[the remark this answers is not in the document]
kblaschke: CLion [can display a list with all "todo" items](https://user-images.githubusercontent.com/5038276/165306668-4216a1d2-d012-4d73-8304-a90aeb3405c5.png) in the source, so that's actually how I track things I still need to do. It's currently set to a static value via configuration.

src/libprojectM/ProjectM.cpp
kblaschke: Would definitely make the parameter list easier to read, yes.

src/libprojectM/ProjectM.hpp
kblaschke: Totally agree. Probably was meant to get more behavior flags at some point, but since it didn't, it's now definitely overengineered.

src/libprojectM/ProjectM.cpp
kblaschke: Or just `!` after changing it to a boolean.

src/libprojectM/ProjectM.hpp
kblaschke: You're right. Doesn't make a lot of sense nowadays, as compilers tend to just ignore it anyway.

src/libprojectM/ProjectM.cpp
kblaschke: Ideally, yes. Most of this stuff should probably be cleaned up in other ways as well, but that's definitely something for later.

src/libprojectM/ProjectM.cpp
kblaschke: Unsure, probably was meant to in the past, don't know. Implementation cleanup is something we should certainly do later, now I'm just focusing on the header.

src/libprojectM/ProjectM.cpp
kblaschke: Correct. I think I've replaced the logic in most, but not all places. Also comparing with `0` instead of using `nullptr` is not a great thing.

src/libprojectM/ProjectM.cpp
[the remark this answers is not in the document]
kblaschke: There's a new Projects beta running now, need to look into it what it can provide. The previous version was just able to pin issues and PRs plus "notes" to the board, but didn't allow for subtasks and other cards linked to certain work items like branches or issues.

kblaschke:
Rebased on latest master.

src/libprojectM/ProjectM.cpp
kblaschke: Yes, certainly no need to check. Will remove it alongside other dead code in this method.

src/libprojectM/ProjectM.cpp
kblaschke: Looked at it again, there's no need to do that manually, so I've just removed the whole function and let the `unique_ptr` dtor take care of it automatically.

src/libprojectM/ProjectM.cpp
kblaschke: Will fix that at a later point, as it also affects the external API. Will for now try to keep the changes closely to ProjectM.hpp/.cpp.

src/libprojectM/ProjectM.cpp
kblaschke: I removed the parameters, at least in the `ProjectM` class. They're just settings vars.

PR #621, opened 2022-05-16
Fix OpenMP export for static linking

kblaschke:
Use CMake's FindOpenMP module and link to the target if found.

Also export the target for linking the static libprojectM library in downstream projects. This is not required for the shared lib, and in this case, missing OpenMP compiler/linker support will not fail the build.


kblaschke:
Need to fix OpenMP detection on macOS, as the SDK includes the headers, but AppleClang wasn't built with OpenMP support (would require to install libomp via Homebrew to make it work).

kblaschke:
Just missed linking OpenMP in the shared lib. Works now.

PR #630, opened 2022-07-08
A few fixes for emscripten and GLES support

kblaschke:
Also fixed a typo in a function call when threading was disabled.

Emscripten builds will now no longer use OpenMP or threads, as the WebAssembly platform doesn't really support multithreading anyways.

Also had to change a few GL constants introduces in recent refactorings which aren't defined in OpenGL ES headers. Line smoothing is not supported by GLES, so the `glDisable` calls are now wrapped in preprocessor `#ìf` blocks.

### DIFF
--- a/src/libprojectM/Renderer/Shape.cpp
+++ b/src/libprojectM/Renderer/Shape.cpp
@@ -100,7 +100,7 @@ void Shape::Draw(RenderContext& context)
         }
         else
         {
-            auto texture = context.textureManager->getTexture(image, GL_CLAMP_TO_BORDER, GL_LINEAR);
+            auto texture = context.textureManager->getTexture(image, GL_CLAMP_TO_EDGE, GL_LINEAR);
             if (texture.first)
             {
                 glBindTexture(GL_TEXTURE_2D, texture.first->texID);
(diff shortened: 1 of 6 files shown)


src/libprojectM/Renderer/Shape.cpp
```diff
@@ -177,7 +177,9 @@ void Shape::Draw(RenderContext& context)
                            glm::value_ptr(context.mat_ortho));
         glVertexAttrib4f(1, border_r, border_g, border_b, border_a * masterAlpha);
         glLineWidth(1);
+#if USE_GLES == 0
```
revmischa, line 180: I thought that we used `#ifndef` other places.. should double-check that

revmischa (approved):

src/libprojectM/Renderer/Shape.cpp
kblaschke: Yeah, we did, and it caused issues as [it's a 0/1 define](https://github.com/projectM-visualizer/projectm/blob/460a2dcd840c7651bcf22f877793178983841d20/config.h.cmake.in#L67), so the `#ifdef` was always true'ish. Some other defines are simple #define/#undef, but not this one.

PR #637, opened 2022-08-18
Fixed some warnings

JohannesKauffmann:
- Use #pragma warning on MSVC only; again this fixes some warnings in client applications
- Fix -Wshadow warnings
- De-inline event function stubs, this fixes some unused parameter warnings, now both when building the library and using the library

The only change which could potentially break ABI (?) is the de-inlining of the event stubs? I wasn't sure what was the best way to solve these warnings. I noticed that, in the past, the names of these parameters have been commented out in ProjectM.hpp, but that was then reverted for some reason?... But then again with a new major release ABI breakage is excepted (4.0?), so let me know what to do about that.

### DIFF
--- a/src/libprojectM/ProjectM.hpp
+++ b/src/libprojectM/ProjectM.hpp
@@ -278,14 +278,14 @@ public:
     auto ShuffleEnabled() const -> bool;
 
     /// Occurs when active preset has switched. Switched to index is returned
-    virtual void PresetSwitchedEvent(bool hardCut, size_t index) const {};
+    virtual void PresetSwitchedEvent(bool hardCut, size_t index) const;
 
-    virtual void ShuffleEnabledValueChanged(bool enabled) const {};
+    virtual void ShuffleEnabledValueChanged(bool enabled) const;
 
-    virtual void PresetSwitchFailedEvent(bool hardCut, unsigned int index, const std::string& message) const {};
+    virtual void PresetSwitchFailedEvent(bool hardCut, unsigned int index, const std::string& message) const;
 
     /// Occurs whenever preset rating has changed via ChangePresetRating() method
-    virtual void PresetRatingChanged(unsigned int index, int rating, PresetRatingType ratingType) const {};
+    virtual void PresetRatingChanged(unsigned int index, int rating, PresetRatingType ratingType) const;
 
     auto Pcm() -> class Pcm&;
 
--- a/src/libprojectM/ProjectM.hpp
+++ b/src/libprojectM/ProjectM.hpp
@@ -27,12 +27,15 @@
 #include "fatal.h"
 
 #ifdef WIN32
+
+#ifdef _MSC_VER
 // libs required for win32
 #pragma comment(lib, "psapi.lib")
 #pragma comment(lib, "kernel32.lib")
 
 #pragma warning(disable : 4244)
 #pragma warning(disable : 4305)
+#endif /* _MSC_VER */
 
 #include <windows.h>
 #else
--- a/src/libprojectM/ProjectM.cpp
+++ b/src/libprojectM/ProjectM.cpp
@@ -907,6 +907,14 @@ auto ProjectM::ShuffleEnabled() const -> bool
     return m_settings.shuffleEnabled;
 }
 
+void ProjectM::PresetSwitchedEvent(bool hardCut, size_t index) const {}
+
+void ProjectM::ShuffleEnabledValueChanged(bool enabled) const {}
+
+void ProjectM::PresetSwitchFailedEvent(bool hardCut, unsigned int index, const std::string& message) const {}
+
+void ProjectM::PresetRatingChanged(unsigned int index, int rating, PresetRatingType ratingType) const {}
+
 void ProjectM::ChangePresetRating(unsigned int index, int rating, const PresetRatingType ratingType)
 {
     m_presetLoader->setRating(index, rating, ratingType);
--- a/src/libprojectM/ProjectM.cpp
+++ b/src/libprojectM/ProjectM.cpp
@@ -875,7 +875,6 @@ void ProjectM::SelectPresetPosition(unsigned int index)
 
 auto ProjectM::SelectedPresetIndex(unsigned int& index) const -> bool
 {
-
     if (*m_presetPos == m_presetChooser->end())
     {
         return false;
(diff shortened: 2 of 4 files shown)


JohannesKauffmann:
Another option would be to make these event functions pure virtual, to force clients to implement these. It would rid us of the warnings as well.

revmischa:
Cool, thanks!
I think pure virtual would be a little more annoying for people trying to use the library so I like your solution

revmischa (approved):

kblaschke:
> I think pure virtual would be a little more annoying for people trying to use the library so I like your solution

Second that, not everyone needs the callbacks. An empty default implementation is best.

src/libprojectM/ProjectM.hpp
```diff
@@ -27,12 +27,15 @@
 #include "fatal.h"
 
 #ifdef WIN32
+
+#ifdef _MSC_VER
 // libs required for win32
 #pragma comment(lib, "psapi.lib")
```
kblaschke, line 33: Not sure these pragmas are required anymore. CMake will probably add the libraries.

src/libprojectM/ProjectM.hpp
```diff
@@ -27,12 +27,15 @@
...
 #pragma comment(lib, "kernel32.lib")
 
 #pragma warning(disable : 4244)
```
kblaschke, line 36: This is the typical "potential data loss" warning when casting to lower precision types. We should simply use `static_cast<>` to silence the warning - or better fix the actual conversions.

src/libprojectM/ProjectM.hpp
```diff
@@ -27,12 +27,15 @@
...
 #pragma warning(disable : 4305)
```
kblaschke, line 37: Same here, but truncation of data, e.g. converting double to float.

JohannesKauffmann:
Rebased on latest master.

src/libprojectM/ProjectM.hpp
JohannesKauffmann: When trying to see what those warnings exactly were, I had some trouble. Initially, I wasn't getting any. But then I saw that CMake checks if `/W4` can be enabled, which it decides it can't for some reason. Then I configured with `-DCMAKE_CXX_FLAGS="/W4"`, and I got swarmed with warnings, much more than I know what to do about.

I will note, however, that some of these were about not using `/EHsc` despite "using" C++ exceptions (by virtue of using the Microsoft C++ standard library which uses exceptions). When I asked CMake to add `/EHsc` (in `features.cmake`), it decided again that it wouldn't do it and I had to configure with `-DCMAKE_CXX_FLAGS="/EHsc"`.

src/libprojectM/ProjectM.hpp
JohannesKauffmann: I think `kernel32.lib` can be removed, but removing `psapi.lib` results in:
```
dlfcn.c.obj : error LNK2019: unresolved external symbol EnumProcessModules referenced in function dlopen
```

That said, I think it would be better to link to `psapi.lib` explicitly only for those who need it. But `target_link_libraries` doesn't work for `OBJECT` targets yet [1], so maybe the iterim solution would be to move this #pragma to the vendored `dlfcn.h`.

[1]. https://gitlab.kitware.com/cmake/cmake/-/issues/18090

src/libprojectM/ProjectM.hpp
JohannesKauffmann: With the comments above in mind, say the #pragma's are removed from ProjectM.hpp. Then, what's left is an #include of `windows.h` in a public header, without `WIN32_LEAN_AND_MEAN` and friends defined! The `windows.h` header is really heavy and pulls in a lot of stuff, most notably a definition of `max` which breaks the standard library `numeric_limits<T>::max`.

I checked the `ProjectM.hpp` header quickly, and I think the `windows.h` header can safely be removed, and ditto for `unistd.h` on Unix(like). Then this whole thing can just be dropped. Do you have any thoughts about that?

JohannesKauffmann:
A more general comment: I had a lot of trouble getting this library to compile on Windows with MSVC. As it stands, it still won't link GLEW, and I had to edit a lot of files because of `WIN32` vs `_WIN32` macro's (the former not being defined on MSVC). I'll make a pull request shortly.

src/libprojectM/ProjectM.hpp
kblaschke: Any OS-specific headers we can drop, especially windows.h, would be a huge win. I think most of it is just legacy code, someone added it for some reason and it simply was left there.

Also, the lib pragmas should not be required anymore. If the libraries still must be linked specifically, this can be done via CMake, which would also be a better solution as the dependencies will be passed downstream if an application links the static library.

src/libprojectM/ProjectM.hpp
kblaschke: Could add this to the static/shared library targets, which will also pass the library to downstream users of the CMake package. They'll have a similar issue when linking a static libprojectM.lib into their applications.

src/libprojectM/ProjectM.hpp
JohannesKauffmann: After a lot of back and forth, I was not able to come up with a CMake construct to link `psapi`. So if we want to remove the pragma's in this PR all at once, I think I might just move them to `dirent.c` for the time being, unless you can come up with a solution. Unfortunately, that means that the MinGW build stays broken (GCC ignores pragma comment lib).

src/libprojectM/ProjectM.hpp
kblaschke: I think the slash in the compiler option causes CMake to bail out as it's not an allowed character in CMake variable names (`enable_cflags_if_supported` uses the flag 1:1 as part of the variable name). I'll fix that by replacing slashes with underscores in the variable name.

src/libprojectM/ProjectM.hpp
kblaschke: From what I can see is that `psapi.lib` is only required for the `dlopen()` wrapper, which is used in the "Native Preset Factory". The native presets are long defunct and, with the restructured project, do no longer compile.

So my take on that would be removing the whole native preset stuff (which never really worked properly) and get rid of the whole issue. That said, just keep it as-is (with the `#pragma comment`), and I'll swing the axe to remove the broken code.

PR #641, opened 2022-09-11
MSVC fixes

JohannesKauffmann:
Found when further investigating review issues from #637.

Also some minor CMake adjustments. Mostly replacing the broken `WIN32` macro with `_WIN32`.


src/libprojectM/libprojectM_shared.cmake
```diff
@@ -71,7 +71,10 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
             )
 endif()
 
-enable_target_linker_flags_if_supported(projectM_shared PRIVATE --no-undefined)
+# --no-undefined doesn't make sense for Windows as all symbols are always required to be defined
+if (NOT WIN32)
```
kblaschke, line 75: Better use `MSVC` instead of `WIN32`, as MinGW also sets WIN32 to `true` while using GCC, where the flag still makes sense.

kblaschke:
Apart from the small CMake thing, all looks great!

src/libprojectM/libprojectM_shared.cmake
JohannesKauffmann: I make that change based on threads [here](https://github.com/mesonbuild/meson/issues/3670) and [here](https://github.com/mesonbuild/meson/pull/4230). Apparently, it has no effect for Windows binaries, as there can be no undefined symbols in DLLs when building.

MinGW GCC silently swallows the flag but does nothing with it, and according to the issue thread, the LLVM linker warns about ignoring the flag. Unfortunately I wasn't able to test that particular behaviour, not with MSVC LLD or MinGW LLD.

That being said, I can change it if you want.

JohannesKauffmann:
Update: enabling `/W4` and `/EHsc` works fine, it just wasn't happening on my machine because I also passed an option to CMake which `cl.exe` chose to ignore and print a diagnostic about. That's why the compile test thought that the option wasn't supported. I'll update the commit message accordingly.

JohannesKauffmann:
As a further note of clarification, it seems like the CMake on my system adds `WIN32` and `_WINDOWS` to both `cflags` and `cxxflags` ... but only when no user-supplied variables (`-DCMAKE_C_FLAGS=...` or CXX equivalent) are given on the command line. Otherwise, the user-supplied flags take precedence and no default flags are added. Because I had given a custom CXXFLAG, my build wasn't working the first around for me, and the `#if WIN32` check did nothing on my system.

For `cxxflags`, CMake additionally adds `/GR` (rtti) and `/EHsc` by default when no user options are given (and without this PR). I still think it would be better to add `/EHsc` ourselves when supported, like done in this PR, in case the user supplies any options. This will ensure a correct build. I also don't think we need rtti, so there's no need to add that.

Should I update the commit message(s) to reflect this behaviour?

kblaschke:
> As a further note of clarification, it seems like the CMake on my system adds `WIN32` and `_WINDOWS` to both `cflags` and `cxxflags` ... but only when no user-supplied variables (`-DCMAKE_C_FLAGS=...` or CXX equivalent) are given on the command line. Otherwise, the user-supplied flags take precedence and no default flags are added. Because I had given a custom CXXFLAG, my build wasn't working the first around for me, and the `#if WIN32` check did nothing on my system.

Correct, as it's a cache variable, CMake uses the default syntax and provides `WIN32;_WINDOWS` as a default value if not already set.

Nevermind the flag detection stuff, I botched it, but now it's fixed. Rebase on current master and just add the required flags in features.cmake, it'll work as intended now.

JohannesKauffmann:
Rebased on master

Edited:
- Also enable RTTI expliticly on MSVC, as we use `dynamic_cast`.
- Updated commit messages to explain the CMake default `WIN32` vs the standard `_WIN32` define.

src/libprojectM/libprojectM_shared.cmake
JohannesKauffmann: Ping, is this alright?

kblaschke: Sure, as long as it compiles, it's fine!

kblaschke (approved):

PR #644, opened 2022-09-14
CMake: set ENABLE_SHARED_LINKING to ON when ENABLE_SHARED_LIB is enabled

JohannesKauffmann:
This fixes the following CMake error when configuring with just -DENABLE_STATIC_LIB=OFF:

CMake Error at src/libprojectM/CMakeLists.txt:130 (add_library):
  add_library cannot create ALIAS target "projectM::libprojectM" because
  target "projectM_static" does not already exist.

In this case, static linking is disabled, thus the "projectM_static" CMake target doesn't exist. Previously, ENABLE_SHARED_LINKING was always false, and because of that, "projectM::libprojectM" was added as as an alias to "projectM_static", which didn't exist.

Now, ENABLE_SHARED_LINKING is set to true only when ENABLE_SHARED_LIB is enabled, otherwise it will be false.


kblaschke:
Yeah, probably a better default that way around!

PR #649, opened 2022-09-15
Entirely removed native presets.

kblaschke:
The native presets did not build anymore due to refactorings, cause other build issues on Windows (e.g. require linking to psapi.lib) and would need a complete makeover anyway.

Similar functionality might later be added via some scripting interface with a JIT compiler, without the need to compile native code which makes these presets non-portable.


JohannesKauffmann:
It would seem that this is also a good opportunity to remove the `#pragma comment(lib` thingies from ProjectM.hpp.

kblaschke:
Huh, I was sure I did that, but somehow the file wasn't committed. Thanks for catching that!

PR #650, opened 2022-09-15
Fix more warnings

JohannesKauffmann:
- Add missing override
- Use size_t where applicable
- Use correct forward declaration
- Removed unused private member
- Fixed some tabs/spaces as a drive-by
- Use ++i instead of i++ in for-loops


kblaschke:
Don't invest too much effort in the equation parser classes, this whole thing is on the kill list and in for a full rewrite.

kblaschke (approved):

PR #652, opened 2022-09-20
Removed all text-related code from the core lib.

kblaschke:
This removes:
- "Toast" messages
- The help text
- Statistics
- Preset name and (unimplemented) song title display
- Preset selection list and search menu

Some of the above functionality might later be added as an optional library, in a separate repository and with proper UTF-8 support.

Also note this commit will break the rewritten SDL2 frontend for now, but that's fine. I'll fix that after the API changes are finished.


revmischa (approved):
Might make sense to remove key handling too someday

kblaschke:
> Might make sense to remove key handling too someday

Yeah, already have a working branch for that. Need to first clean up the playlist management stuff too, then the keyboard handler is next on the short list.

Issue #656, opened 2022-10-21
Debug image is written on each start of projectM


kblaschke:
**Describe the bug**
In most cases when projectM is started (or the library is reinitialized), a debug main texture image is written to the current user's home directory.

**Platforms affected**
All.

**Additional context**
Reason is an [uninizialized member in the Renderer class](https://github.com/projectM-visualizer/projectm/blob/ad2203f7bbd36690f93dcc7b562fd1c04668fc22/src/libprojectM/Renderer/Renderer.hpp#L105). It should be default-initialized to `false`.

OfficialIncubo:
Silly me, it can't show the project logo on Windows. It only shows the M with Headphones only.

OfficialIncubo:
Silly me, it can't show the project logo on Windows. It only shows the M with Headphones only.

kblaschke:
> Silly me, it can't show the project logo on Windows. It only shows the M with Headphones only.

You mean the "idle" preset? Yes, that's correct. The "project" text was removed a good while ago from the preset, probably was too noisy, dunno. The texture is still there though.

PR #657, opened 2022-10-23
Control debug frame saving via API

kblaschke:
Fixed a bug with an uninitialized member, removed the "d" hotkey and added an API function which can be used to write debug images if required. The filename and path can't be controlled, the file is written to the current working directory with a fixed filename containing a timestamp.

For proper render debugging I suggest using [RenderDoc](https://renderdoc.org/) anyway, as this reveals all GL draw calls and state changes.

Fixes issue #656

### DIFF
--- a/src/libprojectM/Renderer/Renderer.hpp
+++ b/src/libprojectM/Renderer/Renderer.hpp
@@ -102,7 +102,7 @@ public:
     bool correction{ true };
 
     bool noSwitch{ false };
-    bool writeNextFrameToFile;
+    bool writeNextFrameToFile{ false };
 
     milliseconds lastTimeFPS{ nowMilliseconds() };
     milliseconds currentTimeFPS{ nowMilliseconds() };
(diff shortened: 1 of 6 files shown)


src/libprojectM/Renderer/Renderer.hpp
```diff
@@ -102,7 +102,7 @@ class Renderer
     bool correction{ true };
 
     bool noSwitch{ false };
-    bool writeNextFrameToFile;
+    bool writeNextFrameToFile{ false };
```
revmischa, line 105: what's the purpose of the false? just curious

revmischa (approved):
was the 'd' hotkey documented anywhere?

src/libprojectM/Renderer/Renderer.hpp
kblaschke: It initializes the member value with `false`, which will prevent an image being dumped on every library init. Being uninialized before, it (mostly) evaluated to a true'ish value (using whatever 32-bit value was present at the memory position on the heap), which caused undefined behaviour - e.g. the image being dumped without triggering.

kblaschke:
> was the 'd' hotkey documented anywhere?

No, I don't think so. It's also nothing I'd really tie to a key, as it's a developer tool. IMO even dumping the image isn't really necessary as one can simply use RenderDoc and see everything projectM does - every single draw call and state change - while rendering a frame. That's way more helpful than a simple image.

PR #658, opened 2022-10-25
Setter and Getter for beat sensitivity

kockie69:
This will add the possibility to change the beat sensitivity by calling projectm_set_beat_sensitivity instead of using the key handler.

### DIFF
--- a/src/libprojectM/ProjectM.cpp
+++ b/src/libprojectM/ProjectM.cpp
@@ -829,6 +829,11 @@ void ProjectM::SetBeatSensitivity(float sensitivity)
     m_beatDetect->beatSensitivity = sensitivity;
 }
 
+auto ProjectM::GetBeatSensitivity() const -> float
+{
+    return m_beatDetect->beatSensitivity;
+}
+
 auto ProjectM::SoftCutDuration() const -> double
 {
     return m_settings.softCutDuration;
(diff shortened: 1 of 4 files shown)


kblaschke:
Looks great so far, thanks!

Just two small requests:
- Please edit the commit message (and PR title) to state what each does. "First version" isn't really necessary. Something like "Add setter for beat sensitivity value to API".
- When adding a setter, please also add an accompanying getter function (`projectm_get_beat_sensitivity`) so applications can also get the current value easily, e.g. for incrementing/decrementing and removing the necessity to redundantly store the value on the application side. Just add another commit as the aboce with the getter method.

kblaschke (changes requested):
Just two small requests:
- Please edit the commit message (and PR title) to state what each does. "First version" isn't really necessary. Something like "Add setter for beat sensitivity value to API".
- When adding a setter, please also add an accompanying getter function (`projectm_get_beat_sensitivity`) so applications can also get the current value easily, e.g. for incrementing/decrementing and removing the necessity to redundantly store the value on the application side. Just add another commit as the above with the getter method.

kockie69:
I have changed the title and have added a getter

kblaschke:
> I have changed the title and have added a getter

Could you also update the commit message of your first commit, please? Thanks!

PR #660, opened 2022-10-27
Add get and set for beat sensitivity value to API

kockie69:


kblaschke (changes requested):
If the declaration in projectM.h is fixed, the PR is ready to merge!

kockie69:
I have not rebased but pushed to my branch.

kblaschke:
> I have not rebased but pushed to my branch. Some help on the rebase appreciated.

The getter function still has a void return type and thus currently doesn't do anything.

You can amend the last commit with more changes, and/or edit the commit message using `git commit --amend`. Rebasing and editing commits further down in the history is a bit more complicated, but worth getting into it.

I suggest using a Git UI, e.g. TortoiseGit on Windows, which makes it a bit easier.

PR #672, opened 2022-12-04
Split out playlist management

kblaschke:
Another large chunk of work mostly done. I've removed all playlist- and rating-related functionality from the core library and implemented a new, optional library that app authors can use if they don't wish to implement playlist management on their own.

Noteable changes:

- All playlist-related functionality removed from libprojectM. The core library now only requests a preset change via a callback once per preset.
- Added support for loading presets from a data buffer instead of a file. This will enable applications to load presets from any source without the need for physical files.
- Added early error handling capabilities and callbacks to the core library.
- The new library doesn't support ratings. They were never really of any use, and even Milkdrop doesn't do anything with the rating value except for displaying it in the preset editor.
- The playlist can recursively scan directories for presets. Adding filenames directly is also possible.
- There's now a function to retrieve the current playlist in one array, e.g. for saving it to disk.
- A new feature is the filter list. By using a syntax similar to `.gitignore` glob expressions, files can now be filtered by filename or path. Both include and exclude filters can be added in order to create more complex filters.
- All playback control is  now done via API calls. No more key codes to look up.
- Playlist editing features have been improved. Applications can insert or remove items in the playlist at any position. The playlist can also filter duplicate filenames.
- The playlist manager keeps a history of the last 1000 presets played, no matter if shuffle is used or not. As long as the playlist isn't changed, there's a new "play_last" method to go back in history instead of the playlist via "play_previous".
- projectM can now give feedback on presets that failed to load. The playlist manager will retry the next presets in the playlists for a limited, configurable number of times if that happens. If all retries fail, the application can set a callback to receive this information.

The work is not done yet. There's still some missing parts regarding the playlist library:

- Add shared library build plus required targets, export header etc.
- Proper error handling and feedback on more cases.
- When retrying a failed preset load, the playlist manager only skips forward or randomly, but should keep the direction requested by the application (previous/last in history) and probably not retry if  a specific index was played.
- Find issues with different use cases for the playlist and add the missing functionality.


revmischa (commented):
wow, with tests! super

src/libprojectM/MilkdropPresetFactory/MilkdropPresetFactory.cpp
```diff
@@ -229,7 +229,7 @@ MilkdropPresetFactory::LoadPresetFromFile(const std::string& filename)
     {
         return IdlePresets::allocate(this, presetOutputs);
     }
-    else if (protocol == "" or protocol == "file")
```
kblaschke: Never code if you're too tired. :D

kblaschke:
> wow, with tests! super

New code is way easier to test. Will definitely add more tests as I work on other parts as well, helps a lot.

PR #676, opened 2022-12-11
Update Workflow

Blaquewithaq:
Builds Debug, Runs Test, Builds Release:
- Linux
- Emscripten
- MacOS
- Windows

Push_Release:
- Activates based on pushed tag with following syntax: "*-release"
- Waits for all builds to pass
- Downloads artifacts from each build
- Repackage based on OS target (allows for any custom modifications to packaging)
- Create Release Draft (allows for us to make edits before official release)

If we start using a commit convention, we can use actions to auto generate out changelogs.


kblaschke (approved):
Sorry for the delay, looks good to me, merge if you like.

PR #687, opened 2023-01-12
Only build either shared or static projectM and playlist libraries.

kblaschke:
Target name is now just libprojectM::projectM for the main library instead of libprojectM::static or libprojectM::shared.

Now using CMake's default mechanism to select the resulting library type via the BUILD_SHARED_LIBS variable. Main use case will be working with the shared library, which is the default.

Playlist library now also builds as a shared library, including proper symbol exports.

Additionally, deleted the leftover TestRunner class and a few old and rather useless test cases.

Also added LLVM find_package call and as a link dependency if the build is configured for using (experimental) LLVM code.

Reorganized tests to link either one of the object libraries for unit testing.


revmischa (approved):
Shared libs is a whole nother can of worms

Blaquewithaq (approved):

PR #690, opened 2023-01-24
General improvements from issue tracker

### LINKED ISSUE #643: API/CMake and general improvements
Some points of improvements I noticed:
- When building with `-DENABLE_CXX_INTERFACE=ON`, it seems like `ProjectM.hpp` includes `PipelineContext.hpp`, but does not install that header file, resulting in compilation failure.
- Related to this, `ProjectM.hpp` [includes `PipelineContext.hpp` directly](https://github.com/projectM-visualizer/projectm/blob/master/src/libprojectM/ProjectM.hpp#L25) without a relative path to the `Renderer` folder in which it resides. Assuming `PipelineContext.hpp` gets installed to `$CMAKE_INSTALL_PREFIX/include/libprojectM/Renderer`, either the `#include` statement should reflect it's location, or the extra include directory should be exposed in the public CMake/pkgconfig interface.
- `ProjectM.hpp` relies on a `USE_THREADS` define. This is fine when building ProjectM itself, but for libraries, this flag does not get propagated. _Side note: it maybe better to rename this to `PROJECTM_USE_THREADS` to prevent possible name clashes._
- Installing the static library installs `libprojectM.a` to the lib folder, but `-l:projectM.a` is added to the pkgconfig file (and thus linking fails). For shared builds, no pkgconfig file is written.
- Related: the debug prefix (`d` by default) is not propagated to the pkgconfig file.
- The OpenGL library usage requirements are not propagated to the pkgconfig file, leading to linker failures. The same goes for `omp_get_num_threads` and friends, whereever that comes from.
- Last but not least, does ProjectM provide a version define, to faciliate projects conditionally using either an older or newer version? Something like `PROJECTM_MAJOR`, `PROJECTM_MINOR` possibly.

kblaschke:
Refactored and added a few smaller things requested by users or just being broken:

- Fix hard SIGABRT crash if OpenGL wasn't properly initialized on projectM instance creation (caused by std::string being created from nullptr).
- Add version header and API methods to get build/run time versions of libprojectM. Using the header defines will give the version an application was built against, wile the new API function will return the actual projectM library version. For static builds, both should be the same.
- Create proper pkg-config file for libprojectM and libprojectM_playlist and add (most) required flags and libraries.
- Set default symbol visibility to hidden. This will hide all symbols except those prefixed with PROJECTM_EXPORT.
- Reduce header footprint of optional C++ API files and export C++ symbols properly. The classes `ProjectM` and `PCM` will then be marked as visible/declspec dllexport in the shared library.

Will fix #643.


revmischa (approved):
very cool

Blaquewithaq (approved):

PR #695, opened 2023-01-25
Prepare repository for 4.0 RC1

kblaschke:
A few final CMake cleanups. Updated the build isntructions accordingly and bumped the library version to `4.0.0` in preparation for creating the RC1 tag and release.


Blaquewithaq (approved):

BUILDING-cmake.md
```diff
@@ -6,22 +6,37 @@ This file contains in-depth information for building with the CMake build system
...
 mkdir build
 cd build
-cmake ..
-make -j8
+cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/path/to/install-dir
+cmake --build . --target install --config Release
 ```
 
-Output binaries will be in `build/src/...`.
+If the build succeeded, you should now have the projectM libraries and include files in the specified install dir.
```
revmischa, line 28: Maybe it'd be simpler to not tell people to set the install prefix and use the system install prefix? Wouldn't that simplify this quite a bit?

BUILDING-cmake.md
```diff
@@ -78,25 +93,35 @@ CMake has no built-in way of printing all available configuration options. You c
 top-level `CMakeLists.txt` which contains a block of `option` and `cmake_dependent_option` commands, or use one of the
 available CMake UIs which will display the options after configuring the project once.
 
-### Boolean switches
+### Important build switches
 
-The following table also gives you an overview of the available options and their defaults. All options accept a boolean
+The following table also gives you an overview of important build options and their defaults. All options accept a
+boolean
```
revmischa, line 99: Could mention how to change the switches (`ccmake .` is it?)

revmischa (approved):

BUILDING-cmake.md
kblaschke: Depends on the use case, actually. If you're using libprojectM for development, and probably want to have different variants (static/shared) etc., installing somewhere into the user directory is to be preferred. Besides that, installing into the system's default prefix requires admin/root privileges. Plus, the install mechanism doesn't care about existing files, so users may unexpectedly end up overwriting files of their system, which - on Linux - can seriously mess up the package manager's file database.

So in general, setting the prefix to something less destructible is preferable.

Even in package building and on meta distros like Gentoo, files are installed into a staging dir before being copied into their final locations.

Additionally, when using CPack, `CMAKE_INSTALL_PREFIX` is overridden to CPack's temporary install dir anyway.

BUILDING-cmake.md
kblaschke: There's also the Qt-based CMake GUI which can do it, yet I'd rather not go too deeply into CMake details in the build instructions as it's all documented on the CMake website. And depending on the OS and CMake version installed, none of those frontend may even be available. Deleting the build dir (or at least `CMakeCache.txt` inside it) and freshly reconfiguring the build is better in most cases, as fiddling around with the cache can break the build system.

Besides that, most people probably won't build libprojectM themselves, or with default settings. We can still put a link to the CMake website in a prominent place, in case someone can't find it on google.

kblaschke:
I'll merge the PR, if there are any further additions needed in the build documentation, we can still add them in the repo and Wiki later. These will also go on a future developer documentation website alongside the API docs.

PR #699, opened 2023-02-08
Fix collisions with projectM <4

hartwork:
Hello! :wave:

This addresses the concerns about file collusion between projectM >=4 and projectM <4 in a way where projectM <4 and >=4 can be installed in parallel and Linux distros will remain friends with projectM — that would be nice.

This was discussed with @kblaschke at https://github.com/projectM-visualizer/frontend-libvisual-plug-in/issues/1#issuecomment-1421081449, some ideas are his, some ideas are mine and awaiting review.

This pull request affects three key things:
- Headers locations
  - `#include <libprojectM/foobar>` becomes `#include <projectM-4.0/foobar>`
- Shared library names
  - `libprojectM.so` becomes `libprojectM-4.0.so`
  - `libprojectM_playlist.so` becomes `libprojectM-4.0-playlist.so`
- pkg-config files
  - `libprojectM.pc` becomes `projectM-4.0.pc`
  - `libprojectM-playlist.pc` becomes `projectM-4.0-playlist.pc`

Looking forward to review and feedback.

Best, Sebastian


revmischa (approved):
Thanks!
It's fine with me, I don't really know what the best practice is here. "4.0" looks odd to me, why not just "4"? easier to type and less confusing when we release 4.1 etc

BUILDING-cmake.md
```diff
@@ -263,7 +263,7 @@ documentation on how to set up the libprojectM build system for use in another p
 ### Importing libprojectM targets from an installed version
 
 This is the recommended and supported way of importing libprojectM in your project. This project installs a set of CMake
-files in `<PREFIX>/<LIBDIR>/cmake/libprojectM`, containing target definitions, version and dependency checks as well as
+files in `<PREFIX>/<LIBDIR>/cmake/projectM-4.0`, containing target definitions, version and dependency checks as well as
```
kblaschke, line 266: When renaming the package this way, be aware that this will make `find_package` potentially unusable because it doesn't really like dots in the name. this would make CMake-based integration in other projects impossible.

I'd suggest the following, which other libs also do:
- Rename the current files/package name to a usable variant, e.g. `projectM4` or `projectM4-0`.
- Keep the namespace and target names as-is, without any version decorations.
- Create a meta package that has no version number in its package name, e.g. `projectM` as it was before.
- Create additional config scripts in the meta package that use the major and minor version numbers to search the specific version's config dir.
- Perform an additional `find_package(projectM4 ...)` in the meta package to find this package.

Since CMake allows both specifying a minimum version and also EXACT versions, this would still allow using these suffixes in the package name. Qt had this for two versions, but they also replaced it recently with version-less package names, which is _way_ easier to use and maintain.

cmake/GeneratePkgConfigFiles.cmake
```diff
@@ -23,9 +23,12 @@ macro(GENERATE_PKG_CONFIG_FILES target package_name)
             endforeach()
         endif()
 
+        # Get name for "-l:<name>" linker flags
+        get_target_property(_lib_name ${target} OUTPUT_NAME)
```
kblaschke, line 27: This will not work in almost all circumstances, as OUTPUT_NAME is mostly calculated by the generator/build processor at build time and thus empty at configure time.
You'll end up with `-l:_lib_name_NOTFOUND` appearing in the .pc file. I actually tried exactly this and it failed in all my configurations. Also tried `OUTPUT_DEBUG_NAME` etc, same issue. It just doesn't work.

kblaschke (changes requested):
Apart from the two CMake-related issues mentioned, the changes are fine. the CMake issues must be fixed though, otherwise both CMake and pkg-config usage will end up broken.

revmischa:
Thinking about this more, I don't really understand the goal. Why would anyone want or need two different versions of projectM installed at a time?

kblaschke:
> Thinking about this more, I don't really understand the goal. Why would anyone want or need two different versions of projectM installed at a time? If I compile against a shared library then the version of that .so will be built into the binary

I asked the same question in the main discussion linked above. There's a possibility of backporting V4 to an older Linux distro still having V3 and applications using both. A slight possibility would be newer distros also installing both versions, but IMO installing an old, broken library isn't really a good idea.

For future versions, this may come in handy though, see how libSDL struggled with it. They also put the major version into the library base name, to get around this exact issue. Qt is another example. Or the GTK libraries.

hartwork:
> "4.0" looks odd to me, why not just "4"? easier to type and less confusing when we release 4.1 etc

@revmischa good idea, adjusted and pushed.

cmake/GeneratePkgConfigFiles.cmake
hartwork: Hi @kblaschke,

I understand what you are saying, but I cannot confirm it in practice, it works well with generator `Ninja Multi-Config` over here. For proof:

```console
# cd "$(mktemp -d)"
# git clone --depth 1 --branch fix-collisions-with-projectm-prior-to-4-0 https://github.com/hartwork/projectm
# cd projectm
# cmake -S . -B build-ninja-multi -G 'Ninja Multi-Config'
# DESTDIR="$PWD"/ROOT/ ninja -C build-ninja-multi install
# grep -h ^Libs: ROOT/usr/local/lib*/pkgconfig/*.pc
Libs: -L${libdir} -l:projectM-4d  -l/usr/lib/gcc/x86_64-pc-linux-gnu/11/libgomp.so -l/usr/lib64/libpthread.a
Libs: -L${libdir} -l:projectM-4-playlistd
```

What am I missing?

BUILDING-cmake.md
hartwork: Hi @kblaschke, I have taken the time to look at this in detail now.

> When renaming the package this way, be aware that this will make `find_package` potentially unusable because it doesn't really like dots in the name. this would make CMake-based integration in other projects impossible.

I have changed it to `projectM4` now.

I should note that folder name  `projectM-4.0` does work for `find_package(projectM [..])` technically because of the asterisk in [CMake's search patterns](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure), e.g. the one in `<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/` near the end, and e.g. Boost uses this feature consistently:
```
# ls -d1 /usr/lib64/cmake/boost*-[0-9]*.[0-9]/ | wc -l
37
```
No need for that with `projectM4` though, good.

> I'd suggest the following, which other libs also do:
> 
> * Rename the current files/package name to a usable variant, e.g. `projectM4` or `projectM4-0`.

Done. This will save use headaches with projectM 5.

I have pushed a new version now that works well with `find_package` over here. Here's one easy way to see it in action:

```console
$ cd "$(mktemp -d)"
$ git clone --depth 1 --branch fix-collisions-with-projectm-prior-to-4-0 https://github.com/hartwork/projectm
$ cd projectm
$ cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -G Ninja
$ DESTDIR="$PWD"/ROOT/ ninja -C build install

$ cat >hello.c <<HELLO_C_EOF
#include <projectM-4/core.h>
#include <projectM-4/playlist_core.h>

#include <assert.h>
#include <stdio.h>

int main()
{
    projectm_handle pm = projectm_create();
    projectm_playlist_handle playlist = projectm_playlist_create(pm);
    projectm_playlist_destroy(playlist);
    projectm_destroy(pm);
    return 0;
}
HELLO_C_EOF

$ cat >CMakeLists.txt <<CMAKELISTS_TXT_EOF
cmake_minimum_required(VERSION 3.0)

project(hello VERSION 1.0.0)

add_executable(hello hello.c)

find_package(projectM4 REQUIRED CONFIG)
find_package(projectM4Playlist REQUIRED CONFIG)

target_link_libraries(hello PUBLIC libprojectM::projectM libprojectM::playlist)
CMAKELISTS_TXT_EOF

$ cmake -DCMAKE_PREFIX_PATH="${PWD}"/ROOT/usr/local/ -S . -B build_hello
$ make -C build_hello/ VERBOSE=1
$ ./build_hello/hello ; echo $?
```
There is no OpenGL context here, so that program tests compilation and linking more than anything, but that's a detail.

What do you think?

hartwork:
I believe, this is good to go by know. Please let me know what more to adjust or fix, if any. Thank you! :beers:

cmake/GeneratePkgConfigFiles.cmake
kblaschke: If that here works correctly, we can of course keep it that way. Currently having only a Windows machine on hand, so can't test it right now.
Starting after `cd projectm` above:
```bash
# cmake -S . -B build-ninja-multi -G 'Ninja Multi-Config' -DCMAKE_INSTALL_PREFIX="$PWD"/ROOT/
# cmake --build build-ninja-multi --config Debug --parallel
# cmake --build build-ninja-multi --config Release --parallel
# cmake --install build-ninja-multi --config Debug
# cmake --install build-ninja-multi --config Release
```
Then check if both debug and release pkg-config files are installed in the prefix and both have the correct library filenames, e.g. `-l:projectM-4-playlistd` in the projectM-4-playlist-debug.pc file and `-l:projectM-4-playlist` in the projectM-4-playlist.pc file.

src/playlist/CMakeLists.txt
```diff
@@ -117,39 +118,39 @@ install(DIRECTORY api/libprojectM
...
 # pkg-config export, only supported on UNIX systems.
 if(UNIX)
     include(GeneratePkgConfigFiles)
 
-    set(PKGCONFIG_PACKAGE_NAME "libprojectM-playlist")
+    set(PKGCONFIG_PACKAGE_NAME "projectM-4-playlist")
     set(PKGCONFIG_PACKAGE_DESCRIPTION "projectM Playlist Library")
-    set(PKGCONFIG_PACKAGE_REQUIREMENTS_RELEASE "libprojectM")
+    set(PKGCONFIG_PACKAGE_REQUIREMENTS_RELEASE "projectM-4")
     set(PKGCONFIG_PACKAGE_REQUIREMENTS_DEBUG "libprojectM-debug")
```
kblaschke, line 153: This line also needs to be changed to
`    set(PKGCONFIG_PACKAGE_REQUIREMENTS_DEBUG "projectM-4-debug")`
Otherwise it will not find the package dependency.

kblaschke (changes requested):
In the Ubuntu build, which uses Ninja Multi-Config, the .pc files look good. Apart from one line that definitely needs changing, the other changes look good to me.

Blaquewithaq (approved):

src/playlist/CMakeLists.txt
hartwork: @kblaschke great catch, the `Requires:` entry was indeed broken in file `projectM-4-playlist-debug.pc`.
I have pushed a dedicated commit to fix that very issue just now. To reproduce:

### Before
```console
# cd "$(mktemp -d)"
# git clone --depth 1 --branch fix-collisions-with-projectm-prior-to-4-0 https://github.com/hartwork/projectm
# cd projectm
# cmake -S . -B build-ninja-multi -G 'Ninja Multi-Config'
# cmake --build build-ninja-multi --config Debug --parallel
# cmake --build build-ninja-multi --config Release --parallel
# DESTDIR="$PWD"/ROOT_DEBUG/ cmake --install build-ninja-multi --config Debug
# DESTDIR="$PWD"/ROOT_RELEASE/ cmake --install build-ninja-multi --config Release
# find ROOT_*/ -name \*.pc -exec grep -E '^(Libs|Requires):' {} +
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-playlist-debug.pc:Requires:  libprojectM-debug  # <-- BROKEN!
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-playlist-debug.pc:Libs: -L${libdir} -l:projectM-4-playlistd 
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-debug.pc:Requires: opengl 
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-debug.pc:Libs: -L${libdir} -l:projectM-4d  -l/usr/lib/gcc/x86_64-pc-linux-gnu/11/libgomp.so -l/usr/lib64/libpthread.a
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4-playlist.pc:Requires:  projectM-4
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4-playlist.pc:Libs: -L${libdir} -l:projectM-4-playlist 
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4.pc:Requires: opengl 
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4.pc:Libs: -L${libdir} -l:projectM-4  -l/usr/lib/gcc/x86_64-pc-linux-gnu/11/libgomp.so -l/usr/lib64/libpthread.a
```

### After
```console
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-playlist-debug.pc:Requires:  projectM-4-debug
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-playlist-debug.pc:Libs: -L${libdir} -l:projectM-4-playlistd 
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-debug.pc:Requires: opengl 
ROOT_DEBUG/usr/local/lib64/pkgconfig/projectM-4-debug.pc:Libs: -L${libdir} -l:projectM-4d  -l/usr/lib/gcc/x86_64-pc-linux-gnu/11/libgomp.so -l/usr/lib64/libpthread.a
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4-playlist.pc:Requires:  projectM-4
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4-playlist.pc:Libs: -L${libdir} -l:projectM-4-playlist 
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4.pc:Requires: opengl 
ROOT_RELEASE/usr/local/lib64/pkgconfig/projectM-4.pc:Libs: -L${libdir} -l:projectM-4  -l/usr/lib/gcc/x86_64-pc-linux-gnu/11/libgomp.so -l/usr/lib64/libpthread.a
```

cmake/GeneratePkgConfigFiles.cmake
hartwork: @kblaschke I believe that^^ is all working now (since today), more details at https://github.com/projectM-visualizer/projectm/pull/699#discussion_r1111125934 .

kblaschke (approved):

PR #703, opened 2023-02-26
Fix unused variable warnings in release mode

JohannesKauffmann:
The assert()'s are removed in release mode causing unused variable warning noise.

### DIFF
--- a/src/libprojectM/Renderer/hlslparser/src/HLSLTree.cpp
+++ b/src/libprojectM/Renderer/hlslparser/src/HLSLTree.cpp
@@ -2035,6 +2035,7 @@ struct StatementList {
                 // @@ Output function as is?
                 // @@ We have to flatten function arguments! This is tricky, need to handle input/output arguments.
                 assert(!NeedsFlattening(functionCall->argument));
+                (void)functionCall;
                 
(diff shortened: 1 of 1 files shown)


kblaschke (changes requested):
The two .hpp files will be removed soon, as I'm currently integrating the new preset expression parser and those files are part of the old one, so I don't see a reason to change those files (the change will potentially just create rebasing issues).

HLSLParser is a third-party library, so any warnings there shouldn't really bother us and we may rather silence them all via CMake instead of changing the code.

JohannesKauffmann:
Ah, I didn't know those two headers were going to go away soon. I'll remove those changes.

The warnings in the third party library seem to be caused by changes made by projectM to the fork (b03fb81711e3e3bd769a143ae10547f8eb2b01b8), so it seems only logical to fix them here in the fork as well. While it could be done with CMake, surely that's more complex than simple oneliner patch. You're going to end up fiddling with flags that work with one compiler but not another, etc. So for those two reasons I'd prefer the change proposed here.

kblaschke (approved):

kblaschke:
No warnings are the best warnings! Merging.

JohannesKauffmann:
Thanks!

PR #704, opened 2023-03-01
Remove obsolete defines from the buildsystem

JohannesKauffmann:
None of these defines are used anywhere in the code, so they should be safe to delete.

### DIFF
--- a/src/libprojectM/CMakeLists.txt
+++ b/src/libprojectM/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_compile_definitions(
         DATADIR_PATH="${PROJECTM_DATADIR_PATH}"
-        GL_SILENCE_DEPRECATION
+        $<IF:$<PLATFORM_ID:Darwin>,GL_SILENCE_DEPRECATION,>
 )
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
--- a/src/libprojectM/CMakeLists.txt
+++ b/src/libprojectM/CMakeLists.txt
@@ -1,19 +1,15 @@
 add_compile_definitions(
         DATADIR_PATH="${PROJECTM_DATADIR_PATH}"
         $<IF:$<PLATFORM_ID:Darwin>,GL_SILENCE_DEPRECATION,>
+        $<IF:$<PLATFORM_ID:Windows>,NOMINMAX,>
+        $<IF:$<PLATFORM_ID:Windows>,WIN32_LEAN_AND_MEAN,>
+        $<IF:$<PLATFORM_ID:Windows>,STBI_NO_DDS,>
 )
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     # dirent.h support
     set(MSVC_EXTRA_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/msvc")
     include_directories("${MSVC_EXTRA_INCLUDE_DIR}")
-
-    # Additional preprocessor definitions for Windows builds
-    add_compile_definitions(
-            NOMINMAX
-            WIN32_LEAN_AND_MEAN
-            STBI_NO_DDS
-    )
 endif()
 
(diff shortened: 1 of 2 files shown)


src/libprojectM/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 add_compile_definitions(
         DATADIR_PATH="${PROJECTM_DATADIR_PATH}"
-        GL_SILENCE_DEPRECATION
```
kblaschke, line 3: This is used on macOS to silence a _huge_ number of "OpenGL is deprecated, use Metal instead!" compiler warnings (one for each OpenGL function used). Please keep it.

src/libprojectM/CMakeLists.txt
```diff
@@ -12,11 +11,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     add_compile_definitions(
             NOMINMAX
             WIN32_LEAN_AND_MEAN
-            USE_FTGL
```
kblaschke, line 15: This define is used in emscripten builds to use ftgl.h in the OpenGL iincludes. Please keep it, otherwise it'll break stuff.

src/libprojectM/CMakeLists.txt
```diff
@@ -12,11 +11,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
...
-            USE_NATIVE_GLEW
```
kblaschke, line 16: This flag is required on Windows for GLEW to work properly in some configurations, please keep it to not break Windows builds.

kblaschke (changes requested):
The other define are acutally unused and can be removed.

JohannesKauffmann:
Thanks for the review, I indeed didn't know those defines were used that way.

Would it be okay to scope those defines to macOS and Emscripten? The macOS one is used unconditionally, and it seems the Emscripten define is only used for Windows builds?

kblaschke:
Huh, right, now I see. So then just remove USE_FTGL.

For `GL_SILENCE_DEPRECATION`, use a generator expression to limit it to Darwin:
```cmake
add_compile_definitions(
        DATADIR_PATH="${PROJECTM_DATADIR_PATH}"
        $<IF:$<PLATFORM_ID:Darwin>,GL_SILENCE_DEPRECATION>
        )
```

src/libprojectM/CMakeLists.txt
JohannesKauffmann: Done with generator expression.

src/libprojectM/CMakeLists.txt
JohannesKauffmann: Removed as per comments below.

JohannesKauffmann:
Hi, sorry for getting back so late on this.

I've rebased on master, and also included a minor refactor with the Windows-only defines switched to generator expressions.

src/libprojectM/CMakeLists.txt
```diff
@@ -1,23 +1,15 @@
 add_compile_definitions(
         DATADIR_PATH="${PROJECTM_DATADIR_PATH}"
-        GL_SILENCE_DEPRECATION
+        $<IF:$<PLATFORM_ID:Darwin>,GL_SILENCE_DEPRECATION,>
+        $<IF:$<PLATFORM_ID:Windows>,NOMINMAX,>
+        $<IF:$<PLATFORM_ID:Windows>,WIN32_LEAN_AND_MEAN,>
+        $<IF:$<PLATFORM_ID:Windows>,STBI_NO_DDS,>
```
kblaschke, line 6: Only as a note to self, I was just wondering why DDS support was disabled since Milkdrop explicitly supports it. We can keep it disabled for now, but I'll look into that at a later point and find out what the reasons were.

kblaschke (approved):
Looks really good now, will merge.

Blaquewithaq (approved):

Issue #708, opened 2023-03-08
Beat detection values (bass, mid, treb + _att variants) are too high.


kblaschke:
While working on integrating the new preset parser and rewriting the preset rendering process, I noticed some presets were behaving highly erratic or were completely broken (e.g. "shifter - robotopia"). I've tracked down these issues to the beat values being way too high.

The Milkdrop preset authoring guide isn't very specific about the actual value range (>0), but says that the values are expected to be in the range from 0.7 to 1.3 in general and around 1.0 if the levels don't change much.

Looking at [Milkdrop's audio analyzing code](https://github.com/projectM-visualizer/milkdrop2/blob/master/src/vis_milk2/plugin.cpp#L9539), these values aren't absolute sums/averages of the current frequency spectrum intensity, but rather slighly or more smoothed relative values to the intensities of previous frames.

@ErikPrantare - you've refactored the beat detection code a while ago. Would you mind looking into this again and implement it in a more Milkdrop'esque way to keep the value ranges in the expected range?

PR #716, opened 2023-04-26
[DRAFT] Integrate new preset parser and rewrite/refactor preset rendering.

kblaschke:
Will add a list of change later.


kblaschke:
Emscripten version builds again, a crash was fixed and cleanups are done so far.
I'd say the branch is in a good enough state to be merged to master, even if there's some work left to get it into a releasable state.

Please test extensively, review the changes and give your opinions.

src/libprojectM/MilkdropPreset/BlurTexture.cpp
```diff
@@ -0,0 +1,363 @@
...
+                                 staticShaders->GetBlur2FragmentShader());
+
+    m_blurFramebuffer.CreateColorAttachment(0, 0);
+
+    // Initialize Blur VAO/VBO with a single fullscreen quad.
+    static constexpr std::array<float, 16> pointsBlur{
+        -1.0, -1.0, 0.0, 0.0,
+        1.0, -1.0, 1.0, 0.0,
+        -1.0, 1.0, 0.0, 1.0,
+        1.0, 1.0, 1.0, 1.0};
```
revmischa, line 28: I had ChatGPT explain this to me

> This Vertex Buffer Object (VBO) represents a square (technically, two triangles) that cover the entire screen in Normalized Device Coordinates (NDC), with texture coordinates associated with each vertex.
> 
> Here's a breakdown of the values:
> 
> The first two values in each row (-1.0, -1.0, 1.0, -1.0, -1.0, 1.0, 1.0, 1.0) represent the x and y coordinates of four vertices in a 2D space. This is a square that covers the entire viewport in NDC, where (-1, -1) is the bottom left corner and (1, 1) is the top right corner.
> The third and fourth values in each row (0.0, 0.0, 1.0, 0.0, 0.0, 1.0, 1.0, 1.0) represent the texture coordinates (u and v) associated with each vertex. They range from (0, 0) at the bottom left to (1, 1) at the top right of the texture.
> The drawing order of the vertices forms two triangles that together cover the entire viewport. The first triangle is formed by the vertices at (-1,-1), (1,-1), and (-1,1), and the second triangle is formed by the vertices at (-1,1), (1,-1), and (1,1).

src/libprojectM/MilkdropPreset/MilkdropPreset.cpp
```diff
@@ -0,0 +1,304 @@
...
+{
+    Load(presetData);
+}
+
+void MilkdropPreset::Initialize(const RenderContext& renderContext)
+{
+    assert(renderContext.textureManager);
+    m_state.renderContext = renderContext;
+
+    // Initialize variables and code now we have a proper render state.
```
revmischa, line 60:
```suggestion
    // Initialize variables and code now that we have a proper render state.
```

src/libprojectM/MilkdropPreset/MotionVectors.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "PerFrameContext.hpp"
+#include "PresetState.hpp"
+
+#include "Renderer/RenderItem.hpp"
+
+/**
+ * @brief Draws a flexible motion vector field.
+ *
+ * This is broken right now, as it only renders a relatively static 1px point grid, with no apparent motion trails.
```
revmischa, line 11: is this still true?

revmischa (approved):
great stuff! thank you

src/libprojectM/Audio/BeatDetect.cpp
```diff
@@ -64,6 +60,28 @@ auto BeatDetect::GetPCMScale() const noexcept -> float
 }
 
 
+auto BeatDetect::GetFrameAudioData() const -> FrameAudioData
```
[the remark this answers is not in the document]
kblaschke: Yes, it's an object that's - in theory - copied. The compiler optimizes such return values as implicit moves, so when the code actually runs, no copy of created, but the object instance created in the function is moved to the caller's context without creating another object.

The audio data is only requested once per frame, and changes every frame, so using shared pointers here would just add overhead. The presets will get a const reference of this object, so ideally, only this single instance exists during a render call.

src/libprojectM/MilkdropPreset/BlurTexture.cpp
kblaschke: That's a very sophisticated explanation of "what is a quad?" 😅

Yeah, the blur textures are rendered as a simple quad, drawn with a specific gaussian blur shader. The quad also doesn't always fill the whole screen (or framebuffer, in this case), only in the first iteration. Subsequent blur steps will use only half or a quarter of the texture to create even more blurry images.

revmischa: yeah I just didn't get the texture coords part of it

src/libprojectM/MilkdropPreset/MotionVectors.hpp
kblaschke: No, not anymore. It's now properly implemented, I perform it purely on the GPU using a second color attachment and a vertex shader. Some presets like "Radar" and many others now actually look as they should.

JohannesKauffmann:
> First of all, big sorry for creating such a huge PR.

Forgive me if this may come across as bikeshedding, but some CMake parts of bf344b6b49adfcd3b5488759edd9622bfe9f96cd (the `PROJECTM_STATIC_DEFINE` parts) could be extracted to make this PR a bit smaller :)

kblaschke:
Replaced the color masking with attaching/detaching the U/V texture for use with motion vectors, as GLES only supports per-buffer color masking with v3.2.

Blaquewithaq (approved):

kblaschke:
> > First of all, big sorry for creating such a huge PR.
> 
> Forgive me if this may come across as bikeshedding, but some CMake parts of [bf344b6](https://github.com/projectM-visualizer/projectm/commit/bf344b6b49adfcd3b5488759edd9622bfe9f96cd) (the `PROJECTM_STATIC_DEFINE` parts) could be extracted to make this PR a bit smaller :)

What exactly are you proposing there? Can't remember that I've done anything with the linking stuff, besides additing the evaluation library. Also unsure how this would make the PR substantially smaller, as the overwhelming part was rewriting code in the Renderer and MilkdropPreset[Factory] targets.

If we can make the static linking part easier, that's definitely a good thing to do. I'd so that separately though, as this PR already changes enough things, or better said, too many to even properly review.

kblaschke:
I'd personally say this changeset is as good as it gets without adding even more stuff to it, so I'd like anyone to pick up the branch, build and test it for stability. I've not seen any serious issues besides a few remaining rendering problems noted in the description, but these will be addressed later. Overall, it's a _huge_ improvement over the current master, visually and performance-wise.

kblaschke:
FYI, last push was a rebase on latest master to prepare for merging.

JohannesKauffmann:
> What exactly are you proposing there?

The commit `A few Windows build fixes` from this PR changed both CMakeLists.txt such that `PROJECTM_STATIC_DEFINE` is defined for a wider range/scope of targets. My proposal was to move these changes specifically to a separate commit/PR, as these CMakeLists.txt changes didn't seem related to the new preset parser/rendering rewrite, but would apply more globally. It'd make this PR a bit smaller (not much smaller admittedly).

Not sure it is worth changing now, however. PR looks good either way.

kblaschke:
Ah, I see. Yes, I had to do this because otherwise the class definitions in the PCM class headers were different between the compile units, which the MSVC compiler doesn't like (different declspec attributes), causing linker errors. The MilkdropPreset target uses the PCM class to retrieve the frame audio data, and I didn't also want to refactor this class structure as it needs more attention later due to the broken/high spectrum and beat detection values. Could've done this in an earlier PR, but it's really just a small fix for a typo and adding missing defines to other targets. Proper cleanup will be done in issue #708.

kblaschke:
Merging now, as the current branch state is stable enough for most uses. Emscripten might still have some issues left on top of the other things on the to-do list, so @Blaquewithaq please create an issue if there are any problems running on Emscripten, I'll tackle these in a later fix.

PR #735, opened 2023-10-06
Fix filenames of static libraries.

kblaschke:
These were missing the "-4" suffix, and the static playlist library also used underscores instead of dashes.


Blaquewithaq (approved):

revmischa (approved):
i'm still not totally clear on why we have this -4 bit

kblaschke:
So we can install different major versions at the same time. Since not all operating systems support SO versions, specifically Windows, this may cause conflicts in some (rare) circumstances. Besides that, it's not really relevant in any way, but having the -4 in there doesn't hurt.

PR #737, opened 2023-10-14
Constrain/clamp per-pixel mesh size to [8,300] and even values.

kblaschke:
The per-mixel mesh size needs to be a mutiple of two to avoid "bars" being rendered in each axis if the resolution is an odd number.

Also now clamping the value range from 8 to 300 (inclusive) on each axis, with 300 being well above anyone would need (Milkdrop's default of 64x48 is totally sufficient even in 4K).

The third commit just fixes a few compiler warnings by making the casts explicit.

### DIFF
--- a/src/api/include/projectM-4/parameters.h
+++ b/src/api/include/projectM-4/parameters.h
@@ -159,7 +159,8 @@ PROJECTM_EXPORT double projectm_get_preset_duration(projectm_handle instance);
 
 /**
  * @brief Sets the per-pixel equation mesh size in units.
- * @note This will currently remove any active presets and reload the default "idle" preset.
+ * Will internally be clamped to [8,300] in each axis. If any dimension is set to an odd value, it will be incremented by 1
+ * so only multiples of two are used.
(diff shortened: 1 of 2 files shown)


Blaquewithaq (approved):

labkey-matthewb:
I recall that the mesh size might get rounded to a multiple of 4 internally to enable SSE optimization.

kblaschke:
> I recall that the mesh size might get rounded to a multiple of 4 internally to enable SSE optimization.

The SSE code has been removed, as the U/V coordinate calculations are now done in the warp mesh vertex shader, which is even better than using SSE (and portable to ARM as well).

labkey-matthewb:
> > I recall that the mesh size might get rounded to a multiple of 4 internally to enable SSE optimization.
> 
> The SSE code has been removed, as the U/V coordinate calculations are now done in the warp mesh vertex shader, which is even better than using SSE (and portable to ARM as well).
> 
> But this explains why it wasn't an issue in the 3.x versions.

Very cool!

PR #739, opened 2023-11-01
Add vcpkg.json

insel-maz:
Use vcpkg in (the recommended) `manifest mode`.

This allows the library to be build by running the following commands in the `Developer PowerShell for VS 2022`:

```PowerShell
cd projectm
cmake . -B build -DCMAKE_BUILD_TYPE=Release "-DCMAKE_INSTALL_PREFIX=$PWD/dist" "-DCMAKE_TOOLCHAIN_FILE=$Env:VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake"
cmake --build build --target install --config Release
```

### DIFF
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg-configuration.schema.json",
+  "default-registry": {
+    "kind": "git",
+    "baseline": "638b1588be3a265a9c7ad5b212cef72a1cad336a",
+    "repository": "https://github.com/microsoft/vcpkg"
+  }
+}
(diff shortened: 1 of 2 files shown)


kblaschke:
Good thing to have!
Yould you mind changing the GitHub Actions Windows workflow accordingly? I think we can then omit the "install dependencies" phase, right?

kblaschke (changes requested):
The GitHub Actions workflow for the Windows builds need to be stripped of the "install dependencies" pahse, as this is not required in manifest mode.

Would also be nice if you'd update the build instructions for Windows accordingly.

Blaquewithaq:
I'll finish this PR off sometime today

kblaschke:
Done in PR #744, so closing this one. Still thanks for the great idea!

PR #741, opened 2023-11-07
Add shader-based smooth transitions between presets

kblaschke:
## New Preset Transitions

This changeset implements a new transition method using fragment shaders to smoothly blend from one preset to another. Currently, there are six different, randomly chosen transitions, most having some varation to make the transitions less repetitive:

1. A simple crossfade blend.
2. A sweeping effect with a random direction and slightly differing width of the blend edge smoothing.
3. A circular inside-out or outside-in blend, again with a random edge width.
4. A radial blur/zoom effect.
5. A sweeping "warp" effect, either horizontally or vertically with a random direction.
6. A plasma dissolve effect with a randomized pattern and scale.

More shaders can be added quite easily. Documentation to write the shaders [is available as a Gist here](https://gist.github.com/kblaschke/a53952cbaf4e469a0733222f09d6dc1b), will move it to the Wiki after merging the PR. I've implemented them in a way that makes it really easy to test them on Shadertoy and use them with minimal or even no changes in projectM.

## Other Changes

While the main focus was implementing the transitions, a few other changes were done either to aid in the implementation or to improve smaller things which I've stumbled upon during the implementation and didn't find big enough to create a separate PR:

- Moved some classes from the MilkdropPreset dir into the Renderer dir and generalized them to be reusable outside of Milkdrop-specific code.
- Updated the projectm-eval submodule. This will now prevent the static lib and header from the internal target to be installed with libprojectM.
- Removed some dead code, specifically the "Renderer" class, which once was the main workhorse and is now completely gone, wrapping up the renderer rewrite.


kblaschke:
Updated projectm-eval submodule to contain the yyerror function signature fix.

kblaschke:
Added a fix for blur textures. Was using the wrong input texture, and the blur textures weren't properly flipped to align with the main texture.

Blaquewithaq (approved):

kblaschke:
The last commits should also wrap up all known rendering issues, except for some preset shaders rendering blackness due to NaN/Inf values in the calculations. These aren't easy to fix (if at all), but not affecting too many presets.

Will then merge this branch shortly.

PR #744, opened 2023-11-16
Add vcpkg configuration

Blaquewithaq:
This PR is the completed idea/work from: #739 by @insel-maz

I've also updated the documentation, GitHub action, and added GTest to the dependency list.


kblaschke (approved):
Nice! Will make it way easier for vcpkg users to build the library, and that's not only true for Windows.

PR #745, opened 2023-11-16
Improve Audio Data Processing and Spectrum Analyzer

kblaschke:
Reimplemented audio data processing to closely match Milkdrop's algorithms and resulting sample/spectrum values.

proejctM's previous implementation was very simplistic, and thus resulted in sample/beat detection values which weren't even close to what Milkdrop passes to presets. This PR changes/fixes the following audio-related issues:

- Waveform scaling is now using the same value range as Milkdrop, which is a floating-point representation of an 8-bit signed sample format. Milkdrop only gets 8-bit samples from Winamp, so the values are always whole numbers. projectM can also process floating-point samples. These are scaled to the same range (-128.0 to +127.0), but also have their decimals preserved. Waveform data scaling in the custom/default wave drawing code has been restored to Milkdrop's values.
- The spectrum analyzer is now a (modernized) copy of Milkdrop's `FFT` class. It also smoothes and equalized the spectrum a bit to remove noise. Now spectrum-based custom waveforms draw properly.
- The bass/mid/treb and attenuated beat-detection values are now calculated as relative values to the previous frame(s), as in Milkdrop. They're no longer absolute sums of the spectrum, but a difference value revolving around 1.0.
- Waveforms are now best-match aligned to the previous frame, resulting in a more stable, less jaggy waveform rendering as the wave doesn't jump randomly around as much as before.

Fixes issue #708.


src/libprojectM/Audio/MilkdropFFT.hpp
```diff
@@ -0,0 +1,159 @@
...
+     *          2) 200*2^1.928 - 200*2^(1.928*2)        or  761  - 2897  Hz
+     *          3) 200*2^(1.928*2) - 200*2^(1.928*3)    or  2897 - 11025 Hz
+     *
+     * A simple sine-wave-based envelope is convolved with the waveform
+     * data before doing the FFT, to emeliorate the bad frequency response
+     * of a square (i.e. nonexistent) filter.
+     *
+     * You might want to slightly damp (blur) the input if your signal isn't
+     * of a very high quality, to reduce high-frequency noise that would
+     * otherwise show up in the output.
```
revmischa, line 96: is there any audio we expect people to feed in that will not be of "a very high quality"?

src/libprojectM/Audio/WaveformAligner.hpp
```diff
@@ -0,0 +1,55 @@
+/**
+ * @file WaveformAligner.hpp
+ * @brief Mip-based waveform alignment algorithm
```
revmischa, line 3: doc is duplicated

src/libprojectM/Audio/MilkdropFFT.hpp
```diff
@@ -0,0 +1,159 @@
...
+     *   approaches zero; the peaks get tighter and more precise, but
+     *   you also see small oscillations around their bases.
+     * @param power Number of convolutions in the envelope.
+     */
+    void InitEnvelopeTable(float power);
+
+    /**
+     * @brief Calculates equalization factors for each frequency domain.
+     *
+     * @param equalize If false, all factors will be set to 1, otherwise will calculate a frequency-based multiplier.
```
revmischa, line 135: what's the purpose of the equalizer?

revmischa (approved):
sweet

src/libprojectM/Audio/MilkdropFFT.hpp
kblaschke: Don't know, probably not relevant anymore today. Most of the Doxygen text in this class is just a 1:1 copy of the comments in Milkdrop's code, I didn't change it for "historical reasons". :grin:

src/libprojectM/Audio/WaveformAligner.hpp
kblaschke: Yeah... Will just keep the class doc, and only add a "brief" description for the file.

src/libprojectM/Audio/MilkdropFFT.hpp
kblaschke: From what I understand of the code, it adds a log-based scaling curve over the frequencies. I don't know the reasoning behind this, so I just kept the code as-is to match Milkdrop's spectrum analyzer results.

The equalizer curve which is applied to the frequency bands looks like this:
image

kblaschke: The envelope table, which is also used for scaling the input waveform samples (basically using more of the waveform's center than the edges), looks like this:
image

Blaquewithaq (approved):

PR #746, opened 2023-11-24
Fix shading language version detection when using WebGL

kblaschke:
WebGL doesn't follow the OpenGL standard format for the glGetString(GL_SHADING_LANGUAGE_VERSION) return value. Chrome for example will return "OpenGL ES GLSL ES 3.00 (WebGL GLSL ES 3.00 (OpenGL ES GLSL ES 3.0 Chromium))", which does not start with the version number as the standard suggests.

This fix will simply cut off any text preceding the first digit found in the string, if any.


Blaquewithaq (approved):

kblaschke:
Merged.

PR #749, opened 2023-12-25
changed order of includes in projectM-Test-UI so that it could be built

serjykalstryke:
I was having trouble building the entire project from the given code, and I was able to track down the issue with building the test UI project to the order the includes were in. It was calling windows libraries in an order that wasn't working with the newest SDK

Error I was encountering:

Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	C2065	'PKEY_NAME': undeclared identifier	projectM-Test-UI	C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\functiondiscoverykeys_devpkey.h	47		

This error was repeated for every use of PKEY_NAME in that devkey.h file. 

Again, changing the order of the includes fixes this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 build_logs/
 
+.gitmodules
 xcuserdata
 src/libprojectM/config.inp
 src/libprojectM/config.inp.b
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ src/projectM-sdl/build/
 src/libprojectM/build/
 *.pkg
 
+./vcpkg_installed
+
 # CLion
 cmake-build-*
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ to_sync/
 src/projectM-sdl/build/
 src/libprojectM/build/
 *.pkg
+./fontend-qt
 
 ./vcpkg_installed
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "vendor/projectm-eval"]
 	path = vendor/projectm-eval
 	url = https://github.com/projectM-visualizer/projectm-eval.git
+[submodule "frontend-qt"]
+	path = frontend-qt
+	url = https://github.com/serjykalstryke/projectm
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "vendor/projectm-eval"]
-	path = vendor/projectm-eval
-	url = https://github.com/projectM-visualizer/projectm-eval.git
-[submodule "frontend-qt"]
-	path = frontend-qt
-	url = https://github.com/serjykalstryke/projectm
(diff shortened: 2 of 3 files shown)


serjykalstryke:
I think some other things I changed that are unrelated got pushed up with the main change, feel free to ignore those.

kblaschke:
> I think some other things I changed that are unrelated got pushed up with the main change, feel free to ignore those.

Looks like it. Can you update the PR to only include the changes in [pmSDL.hpp](https://github.com/projectM-visualizer/projectm/pull/749/commits/2481c8174a25bfac5b4e8a7dcd4429c7542db599#diff-b75525ef83e2c76eaaf0ad4d7ad192ea56065caa061dbcbb0041784e07378e0a)? You might also keep the .gitignore change if you wish as it seems to make sense for vcpkg users. Then I'll test-build and merge it.

Thanks for the contribution!

serjykalstryke:
Will do that within the next 8 hours!

serjykalstryke:
Alright, I think it is fixed. 

I kept the changes to pmSDL.hpp and .gitignore (and also added some lines so that if someone wanted to have a nested repository for their personal fork of projectm, for better documenting projects built, etc., they could do this without it causing issues for the main project)

.gitignore
```diff
@@ -54,6 +55,9 @@ to_sync/
 src/projectM-sdl/build/
 src/libprojectM/build/
 *.pkg
+./fontend-qt
```
kblaschke, line 58: Please remove this path.

```suggestion
```

For once, this path is misspelled. I'd also recommend not to put any arbitrary paths here which are highly user-specific. Reason is, every user has their own needs, and adding all those possibilities here would blow up the file a bit.

To elaborate further on this: building the Qt frontend (or any other application depending on libprojectM) should not be done this way around (e.g. adding the dependee inside its own dependency) - instead, either build libprojectM separately and use CMAKE_PREFIX_PATH to make the library install dir known to the frontent-qt project. Note that you currently can't just use `add_subdirectory` with libprojectM, as it's not set up to be built as a subproject. We'll probably change this in the future to make integrating libprojectM as a code dependency easier.

See our [Integration Quickstart Guide](https://github.com/projectM-visualizer/projectm/wiki/Integration-Quickstart-Guide) for best practices on using libprojectM in other applications.

.gitignore
```diff
@@ -1,5 +1,6 @@
 build_logs/
 
+.gitmodules
```
kblaschke, line 3: This should also not be ignored because we actually _have_ a .gitmodules file which needs to be there!
```suggestion
```

kblaschke (changes requested):
Please remove the two lines from the .gitignore files as requrested and then squash your commits into one to keep the history a bit cleaner.

Thanks!

.gitignore
serjykalstryke: Good to know, thanks for the feedback and advice! Fixing now.

serjykalstryke:
I closed this request, got everything fixed per your instructions and did a new request

PR #754, opened 2023-12-29
Add new Milkdrop 3 waveforms

kblaschke:
Milkdrop2077 was so kind to provide me with the code his seven newly added default waveform modes and permission to port it into projectM.

Before adding the waveforms, I did some long overdue housekeeping:
- Added consistent namespaces to all library classes.
- Factored out the waveform math for each waveform mode into a separate class to get rid of the huge math function and to be able to easily add more waveforms modes without much boilerplate code required.
- Fixed a few smaller compiler warnings and issues.

There's also a new test preset for each of the seven additional waveform modes.


OfficialIncubo:
I'm very happy for the new waveforms to be ported on projectM!

...but hey! Those 7 new waveforms are still on my BeatDrop improvement, also on BeatDrop2077?

kblaschke:
> EDIT: Do you want to extend custom shapes and waves to 16 or not?

While this would be an easy thing to do, performance-wise it can have a _heavy_ impact, especially the shapes, which put a huge strain on the GPU draw-call-wise (up to two draw calls per shape instance, so 16 shapes with up to 1000 instances would mean doing up to 32.000 individual draw calls, which _will_ give you performance issues even on a modern RTX GPU). Since projectM is also meant to run on slower hardware like the RPi, this can lead to heavy FPS drops. If I can find a way to reduce the draw calls to one per custom shape type, this might be a thing to consider - the per-instance expression code still has to run though, up to 16.000 times, which is heavy on the CPU.

> EDIT 2: [Milkdrop3](https://github.com/milkdrop2077/MilkDrop3)'s X Wave isn't controllable/rotatable via mystery parameter, but [my version of BeatDrop](https://github.com/OfficialIncubo/BeatDrop-Music-Visualizer) does it 😜 Can you make the X Wave (WaveMode 10) controllable via mystery parameter, like this: [OfficialIncubo/BeatDrop-Music-Visualizer@653fda8](https://github.com/OfficialIncubo/BeatDrop-Music-Visualizer/commit/653fda898c48ae22befe2fb692c216725c2a374f) 
Ah, good idea, added!

> Also, you can assign me and Milkdrop2077!
I can assign you, but not Milkdrop2077.

> EDIT 3: There are some presets that I have made using the new waveforms! Look at them in "Not working with the Original Milkdrop"! [My Milkdrop Preset Pack](https://drive.google.com/file/d/1YxOYfQHmFo0N3wHh3mA0MdNSFnSdS60d/view?usp=drivesdk)
Will do, good test cases are always nice!

milkdrop2077:
Great news :)

kblaschke:
Commit 40dddc3 ended up in the wrong branch, will move it back to where it belongs as the document isn't even finished. Just ignore this one.

Blaquewithaq (approved):

PR #758, opened 2024-01-22
Various OpenGL ES Fixes

kblaschke:
A few things were off regarding GLES support, used on the Raspberry Pi, on Android, in Emscripten builds and when building for the Universal Windows Platform (Xbox) as well.

This includes wrong shader headers and some wrongly used defines. Also fixed the CMake package config files installed with the playlist library to properly pass the dependency on boost::filesystem if it was built with C++14.


Blaquewithaq (approved):

kblaschke:
Reverted the merge and merged the actual commits from the PR. In the future, please do not use the Merge button on this base, but rather locally reset master to the PR branch and then push it, otherwise it'll invalidate the code signature and show all commits as "Unverified".

PR #760, opened 2024-01-30
Use STL/Boost Filesystem API in Texture File Scanner

kblaschke:
We already use the filesystem API from Boost or STL/C++17 in the playlist library, so it was a clear thing we should also do this in our file scanner class that searches for textures.

Removed a lot of low-level compatibilty code, as the new API nicely hides away all the OS-specific implementation stuff.


Blaquewithaq (approved):

Blaquewithaq (approved):

kblaschke:
The pkgconfig files don't contain the (required) boost filesystem library yet.  Have to remove other unsused stuff as well, so I'll postpone fixing pkgconfig unitl I do  this cleanup job.

PR #762, opened 2024-01-31
Clean up minor warnings and code style

dpw13:
This PR does modify some of the types in the C API, but in some functions keeps the types the same and performs a cast in the CWrapper file. If we need to maintain the exact function signatures for binary compatibility of libs, I will revert the signature changes. If changing the types is OK, I'll roll out the remaining changes to the signatures and remove the casts.

This removes nearly all of the warnings on Windows builds (outside of the vendor folder which still produces many warnings) and all warnings on Linux (which surprisingly does not emit warnings for the vendor folder files?).


kblaschke:
Keeping binary compatibility is a must in the API, as 4.0 and 4.1 would otherwise not be compatible I either way, so please revert those changes. Everything that's happening in the C/C++ wrapper and beyond can be changed though.

Also, a few unit tests fail, which points to something not working the same as before.

src/libprojectM/TimeKeeper.cpp
```diff
@@ -95,12 +95,12 @@ int TimeKeeper::PresetFrameA()
     return m_presetFrameA;
 }
 
-int TimeKeeper::PresetTimeB()
+double TimeKeeper::PresetTimeB()
```
kblaschke, line 98: These had an int return value? Huh. Never noticed that, makes no sense, right?

src/playlist/Playlist.hpp
```diff
@@ -166,7 +166,7 @@ class Playlist
      * @param predicate The sort predicate.
      * @param order The sort order.
      */
-    virtual void Sort(uint32_t startIndex, uint32_t count, SortPredicate predicate, SortOrder order);
+    virtual void Sort(size_t startIndex, size_t count, SortPredicate predicate, SortOrder order);
```
kblaschke, line 169: Even if the type would be better for the underlying containers, playlists won't ever exceed 4.2 billion presets. Better convert the types in the implementation.

src/playlist/api/projectM-4/playlist_callbacks.h
```diff
@@ -42,7 +43,7 @@ extern "C" {
  * @param user_data A user-defined data pointer that was provided when registering the callback,
  *                  e.g. context information.
  */
-typedef void (*projectm_playlist_preset_switched_event)(bool is_hard_cut, unsigned int index,
+typedef void (*projectm_playlist_preset_switched_event)(bool is_hard_cut, size_t index,
```
kblaschke, line 46: This will change the ABI, please revert.

kblaschke (changes requested):

dpw13:
Sounds good! I'll get on these changes tomorrow. I missed the C++14
compatibility and didn't test OSX builds, so I'll fix the unused parameter
warnings some other way.

src/libprojectM/RandomNumberGenerators.hpp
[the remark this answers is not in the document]
dpw13: Perfect, that'll be less code to maintain in the future.

src/libprojectM/TimeKeeper.cpp
dpw13: Yeah, I was surprised too. One of the few places the type mismatch might matter.

src/playlist/Playlist.hpp
dpw13: Agree, none of the instances of size_t should really matter for this application. size_t made more sense to me since it's an index, but I'm fine with casting especially in this case.

CMakeLists.txt
```diff
@@ -200,10 +200,11 @@ else()
     set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 endif()
 
+add_subdirectory(vendor)
+
 include(features.cmake)
 
 add_subdirectory(presets)
```
kblaschke, line 207: Could also move up the presets dir, as it just contains an install directive for testing, but no build targets.

src/libprojectM/ProjectM.hpp
```diff
@@ -110,7 +110,7 @@ class PROJECTM_EXPORT ProjectM
      */
     void LoadPresetData(std::istream& presetData, bool smoothTransition);
 
-    void ResetOpenGL(size_t width, size_t height);
+    void ResetOpenGL(int width, int height);
```
kblaschke, line 113: Could just change this (and subsequent members etc. storing the window size) to `uint32_t`, as negative window sizes don't make much sense. The API also uses `size_t`, so there's no reason to make it signed.

src/playlist/Filter.cpp
```diff
@@ -124,6 +124,7 @@ auto Filter::ApplyExpression(const std::string& filename, const std::string& fil
 
                 currentFilenameChar++;
                 currentFilterChar++;
+                continue;
```
kblaschke, line 127: 👍🏻 Good catch, this one!

kblaschke (approved):
Looks good and even fixes two bugs!

src/libprojectM/ProjectM.hpp
dpw13: Good catch. The members are already uint32_t, I must have just missed the prototype here.

CMakeLists.txt
dpw13: Erring on the side of more warnings, I think I'll leave this as-is for this PR.

PR #764, opened 2024-02-02
Relocate DLL's to PROJECTM_BIN_DIR on Windows

Blaquewithaq:
I have submitted this pull request with the intention of aligning the output directory structure more closely with the conventions observed on Windows.

DLL's are more commonly found in /bin directories.

I'm also working on the VCPKG packaging and this is their preferred structure.


kblaschke (changes requested):
Instead of doing the same thing twice, better move the code to the main CMakeLists.txt.

Additionally, using `if(WIN32)` is long deprecated, and will cease to work at some point. Please use `if(CMAKE_SYSTEM_NAME STREQUAL Windows)` instead.

kblaschke (approved):
Nice and clean!

PR #766, opened 2024-02-02
Refactor waveform aligner and add test

dpw13:
This PR adds a friend test for WaveformAligner that verifies a few internal implementation details as well as verifying proper alignment for waveforms with supported offsets.

This PR also refactors WaveformAligner for readability and fixes two bugs:
1) The code to calculate errorSum is supposed to calculate a windowed convolution but introduced an indexing bug that resulted in repeatedly calculating the product of the two waveforms instead.
2) The calculated "mip levels" or resampled waveforms were recalculated in the original Milkdrop code but was re-used as an optimization in the projectM code. However the re-used mip levels did not take the applied shift into account, resulting in the calculated offset being relative to the previous *unshifted* waveform. This bug was not present in the original Milkdrop code since the mip levels were recalculated including the applied shift.

### DIFF
--- a/src/libprojectM/Audio/WaveformAligner.cpp
+++ b/src/libprojectM/Audio/WaveformAligner.cpp
@@ -1,5 +1,6 @@
 #include "WaveformAligner.hpp"
 
+#include <algorithm>
 #include <cmath>
 #include <iterator>
 
@@ -9,6 +10,8 @@ namespace Audio {
 WaveformAligner::WaveformAligner()
 {
     static const uint32_t maxOctaves{10};
+    // For AudioBufferSamples = 576 and WaveformSamples = 480:
+    // floor(log2(96)) = 6
     static const uint32_t numOctaves{static_cast<uint32_t>(std::floor(std::log(AudioBufferSamples - WaveformSamples) / std::log(2.0f)))};
     m_octaves = numOctaves > maxOctaves ? maxOctaves : numOctaves;
 
@@ -32,15 +35,19 @@ void WaveformAligner::Align(WaveformBuffer& newWaveform)
 {
     if (m_octaves < 4)
     {
+        // The original code does not align if there isn't enough margin for
+        // alignment but has no explanation for why the limit is 2**4 samples.
         return;
     }
 
     int alignOffset{};
 
     std::vector<WaveformBuffer> newWaveformMips(m_octaves, WaveformBuffer());
+    // Octave 0 is a direct copy of the new waveform
     std::copy(newWaveform.begin(), newWaveform.end(), newWaveformMips[0].begin());
 
     // Calculate mip levels
+    // This downsamples the previous octave's waveform by a factor of 2
     for (uint32_t octave = 1; octave < m_octaves; octave++)
     {
         for (uint32_t sample = 0; sample < m_octaveSamples[octave]; sample++)
@@ -51,6 +58,7 @@ void WaveformAligner::Align(WaveformBuffer& newWaveform)
 
     if (!m_alignWaveReady)
     {
+        // The below is performed only on the first fill.
         m_alignWaveReady = true;
         for (uint32_t octave = 0; octave < m_octaves; octave++)
         {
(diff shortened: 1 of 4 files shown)


src/libprojectM/Audio/WaveformAligner.cpp
```diff
@@ -21,110 +23,109 @@ WaveformAligner::WaveformAligner()
...
+        for (int offset = offsetStart; offset < offsetEnd; offset++) // 0 to 2
         {
             float errorSum{};
 
-            for (size_t i = m_firstNonzeroWeights[octave]; i <= m_lastNonzeroWeights[octave]; i++)
+            // perform the pseudo-autocorrelation
+            for (uint32_t i = m_firstNonzeroWeights[octave]; i <= m_lastNonzeroWeights[octave]; i++)
             {
-                errorSum += std::abs((newWaveformMips[octave][i + sample] - m_oldWaveformMips[octave][i + sample]) * m_aligmentWeights[octave][i]);
+                errorSum += std::abs((newWaveformMips[octave][i + offset] - m_oldWaveformMips[octave][i]) * m_aligmentWeights[octave][i]);
```
dpw13, line 123: Fix 1.

src/libprojectM/Audio/WaveformAligner.cpp
```diff
@@ -155,21 +162,57 @@ void WaveformAligner::Align(WaveformBuffer& newWaveform)
...
+    {
+        std::copy_n(newWaveform.begin() + alignOffset, WaveformSamples, newWaveform.begin());
 
         // Set remaining samples to zero.
         std::fill_n(newWaveform.begin() + WaveformSamples, AudioBufferSamples - WaveformSamples, 0.0f);
     }
+
+    // Store mip levels for the next frame. Note that we need to recalculate the mips for the *shifted*
+    // waveform, so we can't reuse the previous mips.
+    ResampleOctaves(m_oldWaveformMips, newWaveform);
```
dpw13, line 215: Fix 2.

kblaschke (changes requested):
Please split the individual changes into separate commits: one for the reformatting, one for each fix and one for the tests. This will make it easier to review and later much easier to see what was fixed where and why, without the need to dig up this PR.

If possible, add the reasoning for each change as mentioned in the PR into the respective commit messages.

dpw13:
I split the commit into several pieces that should be much easier to review. I've explained what's changing in each commit. The commit that splits `Align` into multiple methods is likely the more difficult to review; I suggest disabling viewing whitespace changes with that commit. I've ordered the methods so hopefully the diff looks cleaner than before (where `ResampleOctaves` was placed too low in the file). The last commit contains only the actual fixes so `git blame` should make it clear what's changed and why.

src/libprojectM/Audio/WaveformAligner.cpp
```diff
@@ -31,104 +31,101 @@ WaveformAligner::WaveformAligner()
...
+        sample = compareSamples - 1;
+        while (m_aligmentWeights[octave][sample] == 0 && compareSamples > 1)
+        {
+            sample--;
+        }
+        m_lastNonzeroWeights[octave] = sample;
     }
+}
 
+int WaveformAligner::CalculateOffset(std::vector<WaveformBuffer> &newWaveformMips)
```
kblaschke, line 122: Better align reference/pointer designators to the type (or ideally use the .clang-format file in the root to format the code automatically).
`int WaveformAligner::CalculateOffset(std::vector<WaveformBuffer>& newWaveformMips)`

src/libprojectM/Audio/WaveformAligner.hpp
```diff
@@ -34,7 +34,11 @@ class WaveformAligner
      */
     void Align(WaveformBuffer& newWaveform);
 
-private:
+protected:
+    void GenerateWeights();
+    int CalculateOffset(std::vector<WaveformBuffer> &newWaveformMips);
```
kblaschke, line 39: Same here, just run clang-format, that should fix it.

kblaschke (approved):
Besides ideally running clang-format on the changed files before each commit, it looks good to me!

kblaschke (approved):
Clean and well tested! If you're ready, I could merge at any time.

Blaquewithaq (approved):

PR #767, opened 2024-02-02
Add VCPKG port

Blaquewithaq:
This pull request introduces VCPKG packaging for the ProjectM library. This enhancement allows users to easily install the library through VCPKG, enabling its integration into CMake projects on the Windows platform.

### Questions

- What do we want the library name to be within VCPKG? I've used `libprojectm` for now, but we could change it to what's preferred.

Let me know what changes you guys want if any!

### DIFF
--- a/.github/workflows/build_windows_vcpkg.yml
+++ b/.github/workflows/build_windows_vcpkg.yml
@@ -16,11 +16,23 @@ jobs:
     name: Shared Library
     runs-on: windows-latest
 
+    env:
+      EVENT_CONTEXT: ${{ toJson(github.event) }}
+
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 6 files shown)


Blaquewithaq:
Alright, I played around with it for a few hours and now got it working with PR's

kblaschke:
I don't think that the vcpkg port files should be part of the library repository. Adding all the packaging system files for all Linux distros and other build systems would be a huge mess.

If there are changes required like the DLL install paths, this of course should be done here, but the actual port should be committed into the [upstream vpkg repository](https://github.com/microsoft/vcpkg/tree/master/ports) instead. Once it's there, we can still run a build here to check that changes in the library didn't break the vcpkg port (or we actually have to update the port, if the breaking change is actually required).

PR #772, opened 2024-02-06
Use renderContext for preset state time, frame, fps, progress

dpw13:
The robotopia preset uses the current time in per-waveform calculations. For custom waveforms, this is pulled directly from the preset state's time field. However, this field is never set and is permanently zero. For custom shapes, the time is pulled from the preset state's render context. Looking through the code, it appears that these values migrated to the render context at some point but weren't updated everywhere. I've removed these fields from the preset state entirely to prevent confusion and updated references to the preset state's time, frame, fps, and progress counters to use the preset state's render context. This fixes (I think) the shifter - robotopia presets.


kblaschke:
shifter - robotopia

This PR _definitely_ fixes the preset - and probably many others, too!
Thanks for looking into it, good to see it was just another part that got lost (or never was fixed) in my refactoring of the renderer. Really helps to have another pair of eyes on the code - _so_ important.

PR #773, opened 2024-02-07
Fix various typos

luzpaz:
Found via `codespell -q 3 -S "./vendor" -L ist`


kblaschke (commented):
Thanks! Would just rephrase this one sentence so it is gramatically sound, the other changes are good to go!

PR #775, opened 2024-02-10
Use PRJM_EVAL_F in ShapePerFrameContext

dpw13:
Noticed that ShapePerFrameContext is still using doubles instead of PRJM_EVAL_F. Should have no noticeable effect apart from perhaps faster execution when PRJM_EVAL_F is a float instead of a double on embedded platforms.


kblaschke (approved):
Nothing to add, that's how it should be done.
Same procedure as with every PR, rebase it on master & I'll FF-merge it ASAP!

Issue #724, opened 2023-09-12
Warp/composite mesh U/V coordinates cause issues with some preset shaders


kblaschke:
After the renderer rewrite, some presets still do not render properly, leaving parts or the whole the image black. A [few presets where this is quite obvious](https://github.com/projectM-visualizer/projectm/files/12584233/Presets.zip) are:
- martin - stormy sea
- Goody + martin - crystal palace - Aqua Lumens
- Goody + martin - crystal palace - Ocular Anima

The reason for this is that some presets use U/V coordinates in the exponent part of the `pow()` function, which is not allowed to be negative. Since texture coordinates - at least in OpenGL - can be negative, this will cause the result to become `NaN`, breaking any subsequent calculations on the value. It might affect other presets as well, not causing clearly visible back areas but other less obvious artifacts as the values may not be in the range expetected by the preset author.

One possible fix might be to wrap the U/V coordinates in the warp mesh vertex shader to always be in the range [0...1[.

kblaschke:
Note this isn't _directly_ related to issue #725. While the other issue fixed some now superfluous negations in the u/v coordinates of the warp mesh, the issue here still remains unsolved.

dpw13:
The following intrinsics have a limited domain where their output is defined and meaningful. Documentation for both HLSL shader model 3 and GLSL say that the output of `sqrt(x)` where `x < 0`, for instance, say the output is NaN. However, in reality this function returns `sqrt(abs(x))` if `x < 0` in DX9's shader model 3 implementation. This has been verified using [BeatDrop 1.3.1](https://github.com/OfficialIncubo/BeatDrop-Music-Visualizer/releases/tag/1.3.1) after installing DX9 (note that BeatDrop requires DX9 specifically; otherwise you'll get errors about shader model 3 being reported as supported but non-functional).

By creating a shader that plots the output of these intrinsics, we can easily verify the behavior of each intrinsic. Here are the presets I used to plot these intrinsics. The R+G channels plot the unmodified intrinsic. The B channel plots the modified function to be used in projectM. If the plot has no yellow or blue regions, these functions overlap perfectly.

[math_presets.zip](https://github.com/projectM-visualizer/projectm/files/14250877/math_presets.zip)

The functions include:
[...]

PR #778, opened 2024-02-12
Match math function behavior to DX9 implementation

dpw13:
See https://github.com/projectM-visualizer/projectm/issues/724

### DIFF
--- a/vendor/hlslparser/src/GLSLGenerator.cpp
+++ b/vendor/hlslparser/src/GLSLGenerator.cpp
@@ -986,6 +986,25 @@ void GLSLGenerator::OutputExpression(HLSLExpression* expression, const HLSLType*
             m_writer.Write("))/log(10.0))");
             handled = true;
         }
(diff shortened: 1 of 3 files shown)


kblaschke:
```c++

        else if (String_Equal(functionName, "pow"))
        {
            HLSLExpression* argument[2];
            if (GetFunctionArguments(functionCall, argument, 2) != 2)
            {
                Error("%s expects 2 arguments", functionName);
                return;
            }
            // See rsqrt above regarding abs().
            m_writer.Write("pow(abs(");
            OutputExpression(argument[0]);
            m_writer.Write("),");
            OutputExpression(argument[1], &functionCall->function->argument->type);
            m_writer.Write(")");
            handled = true;
        }
```
Adding this after handling `log10` in GLSLGenerator.cpp fixes "stormy sea", as `pow()` is also affected by the NaN madness.

kblaschke (approved):

PR #779, opened 2024-02-13
Modify NaN propagation for multiplication

dpw13:
This solves some edge cases in presets that suppress NaN or infinities by multiplying by zero. It's not clear what the performance impact is with this change, and it may not be something we want to always enable. Because of that, I've added an option flag to 
 `GLSLGenerator::Generate()` to enable or disable the behavior. This may be something we don't want to end up merging if the performance impact is too high.

### DIFF
--- a/vendor/hlslparser/src/HLSLParser.cpp
+++ b/vendor/hlslparser/src/HLSLParser.cpp
@@ -620,8 +620,8 @@ const Intrinsic _intrinsic[] =
         INTRINSIC_FLOAT2_FUNCTION( "step" ),
         INTRINSIC_FLOAT2_FUNCTION( "reflect" ),
 
-		INTRINSIC_FLOAT1_FUNCTION("isnan"),
-		INTRINSIC_FLOAT1_FUNCTION("isinf"),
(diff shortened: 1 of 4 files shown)


vendor/hlslparser/src/HLSLParser.cpp
```diff
@@ -619,8 +619,8 @@ const Intrinsic _intrinsic[] =
         INTRINSIC_FLOAT2_FUNCTION( "step" ),
         INTRINSIC_FLOAT2_FUNCTION( "reflect" ),
 
-		INTRINSIC_FLOAT1_FUNCTION("isnan"),
-		INTRINSIC_FLOAT1_FUNCTION("isinf"),
+		Intrinsic("isnan",    HLSLBaseType_Bool, HLSLBaseType_Float),
```
dpw13, line 622: I'm not aware of any presets that use these functions, but in trying to identify NaNs and Infs in shaders I found that the return type was declared incorrectly here. Changing this line results in the resulting shader correctly casting the output where necessary.

kblaschke (approved):

PR #781, opened 2024-02-16
Interpret negative alpha in shapes and waveforms as full intensity

dpw13:
Mostly fixes xtramartin(186).

### DIFF
--- a/src/libprojectM/MilkdropPreset/CustomShape.cpp
+++ b/src/libprojectM/MilkdropPreset/CustomShape.cpp
@@ -3,6 +3,7 @@
 #include "PresetFileParser.hpp"
 
 #include <Renderer/TextureManager.hpp>
+#include <Renderer/RenderItem.hpp>
 
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


dpw13:
image

kblaschke:
It's not actually DirectX doing something different or wrong here, but [the way how Milkdrop converts/passes the color into the vertex data](https://github.com/projectM-visualizer/milkdrop2/blob/master/src/vis_milk2/milkdropfs.cpp#L2387-L2396). Milkdrop converts the float value times 255 to `int` and then masks away all bits except the last eight and assigns each color plus alpha to a 32-bit unsigned int representing an 8-bit RGBA color value:

```c++
                    v[0].Diffuse = 
                        ((((int)(*pState->m_shape[i].var_pf_a * 255 * alpha_mult)) & 0xFF) << 24) |
                        ((((int)(*pState->m_shape[i].var_pf_r * 255)) & 0xFF) << 16) |
                        ((((int)(*pState->m_shape[i].var_pf_g * 255)) & 0xFF) <<  8) |
                        ((((int)(*pState->m_shape[i].var_pf_b * 255)) & 0xFF)      );
```

Since negative int values may have many bits set (including the first one determining that the value is actually negative), this will mostly result in maxed out values, or at least very bright ones.

We could perform a similar calculation to match the original code as closely as possible.

PR #782, opened 2024-02-17
Scale and smooth samples for non-custom waveforms

dpw13:
Partially fixes Benjam and Zylot - Tie-Dye Supernova - prophecies - subconscious feednowhere applets nz+. The issue was that the default waveform is scaled so that it colors the center of the window, which is then zoomed in to produce an outflowing of somewhat random color. Without the amplitude set high enough, the waveform never reaches the center so you only get a black screen.

### DIFF
--- a/src/libprojectM/MilkdropPreset/Waveforms/WaveformMath.cpp
+++ b/src/libprojectM/MilkdropPreset/Waveforms/WaveformMath.cpp
@@ -40,11 +40,23 @@ auto WaveformMath::GetVertices(const PresetState& presetState,
                   begin(m_pcmDataR));
     }
 
-    // Scale waveform data to [-1, 1] range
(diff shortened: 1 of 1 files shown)


kblaschke:
Those default waveform modes still mostly use the original projectM code, including the "maximize color" stuff etc., so there's quite a good chance that the code does many things wrong. This was one of the few parts I didn't refactor much since the 3.x releases.

You could have a look at the [original implementation](https://github.com/projectM-visualizer/milkdrop2/blob/master/src/vis_milk2/milkdropfs.cpp#L2765-L3381) and cross-check/rewrite any code that's much different from the original. Since it's just calculating a few vertex coordinates, we should be able to get this function to behave nearly identical to Milkdrop.

kblaschke (approved):

PR #783, opened 2024-02-20
Search preloaded random textures in TextureManager::GetRandomTexture

dpw13:
We currently only check textures loaded from files when selecting a random texture. Since we purge textures between presets, if a preset uses something like `sampler_rand00`, no presets will be found, the returned sampler will be null, and we end up creating invalid GLSL. This PR doesn't check the results of `GetRandomSampler` to make sure they're valid, but does extend `GetRandomSampler` to include preloaded textures. As this list includes volumetric textures, also filter textures by type. This type defaults to 2D textures and is not otherwise specified elsewhere as I can't find evidence that `GetRandomSampler` is ever expected to return a 3D texture.

### DIFF
--- a/src/libprojectM/Renderer/TextureManager.cpp
+++ b/src/libprojectM/Renderer/TextureManager.cpp
@@ -238,7 +238,7 @@ auto TextureManager::LoadTexture(const std::string& fileName, const std::string&
     return {newTexture, sampler, name, unqualifiedName};
 }
 
-auto TextureManager::GetRandomTexture(const std::string& randomName) -> TextureSamplerDescriptor
+auto TextureManager::GetRandomTexture(const std::string& randomName, GLenum type) -> TextureSamplerDescriptor
 {
     std::string selectedFilename;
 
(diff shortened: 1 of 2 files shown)


kblaschke (changes requested):
This will completely eliminate using the textures from disk, and also removed the prefixed random texture selection. `m_textures` only contains a few textures used by previous presets, the "M" and headphones image and the noise textures - solely using it will reduce the number of visible textures a lot and often use the crappy noise textures, which we don't want.

While the texture purge clears the scanned texture file list, there's a call to `ScanTextures();` in this method which will refresh this list if it was purged.

So rather than changing the whole random texture selection process into something seemingly much worse, we should focus on fixing the reason why the `m_scannedTextureFiles` might be empty in a certain case when it shouldn't be.

Note that if you don't set a texture path, this list will always be empty. Instead of returning a nullptr, we could as well create a 1x1 black texture and return this instead.

dpw13:
This turned out to be an empty search path in frontend-sdl2, which I'll fix separately. It looks like things work as expected when the texture search path is sane.

Issue #682, opened 2023-01-07
Improve line rendering on higher resolutions


kblaschke:
Curently, projectM implements the exact same line rendering technique as Milkdrop. It renders either a single, 1px-wide line or four 1px-wide lines in a 2x2 raster for thick/bold lines. There are a few issues with the current implementation, from which Milkdrop also suffers:

- Line anti-aliasing is not a guaranteed feature, e.g. it's basically only available on desktop platforms _and_ if the line is only 1px wide.
- The line with always stays the same, regardless of the rendering resolution, meaning that the higher the resolution gets, there is less and less brightness/color added to the image overall. This causes many presets to become darker or almost black on 4K+ screens, while very small resoutions will be way too bright.
- Vulkan/Metal don't support line rendering with the optional `VK_EXT_line_rasterization`  extension.

Ideally, line width should scale smoothly with the rendering resolution, e.g. 1px width (or 2px for thick lines) at 1000px. such lines would then render 4px/8px wide on 4K resolutions, covering about the same percentage of image pixels overall.

One feasible approach would be implementing a technique suggested by [this article](https://wwwtyro.net/2019/11/18/instanced-lines.html), which will also work fine for OpenGL and WebGL. This will also take care of line joints and end caps to look nice, e.g. shapes can use miter joints while waveforms should ideally be rendered with round joints and end caps.

PR #785, opened 2024-02-21
Properly cast argument types in GLSL

dpw13:
I was occasionally getting type errors around the `pow` function after https://github.com/projectM-visualizer/projectm/pull/779. This appears to be because the type casting used in `OutputExpressionList` was being omitted. For the other functions with an added `abs` call, I use `OutputExpressionList` to recapture the previous casting behavior and for `pow` I use the function's return type, ensuring that all arguments are expanded into vectors where possible. `pow` in particular is often used in presets where the base or exponent is a scalar and the other argument is a vector.

### DIFF
--- a/vendor/hlslparser/src/GLSLGenerator.cpp
+++ b/vendor/hlslparser/src/GLSLGenerator.cpp
@@ -971,12 +971,6 @@ void GLSLGenerator::OutputExpression(HLSLExpression* expression, const HLSLType*
         }
         else if (String_Equal(functionName, "rsqrt"))
         {
-            HLSLExpression* argument[1];
-            if (GetFunctionArguments(functionCall, argument, 1) != 1)
-            {
-                Error("rsqrt expects 1 argument");
-                return;
-            }
             /* The documentation says that these functions return NaN for negative
              * arguments. However, testing with DX9 shader model 3 shows that they
              * most definitely do take the absolute value of the argument and do
@@ -984,7 +978,7 @@ void GLSLGenerator::OutputExpression(HLSLExpression* expression, const HLSLType*
              * See https://github.com/projectM-visualizer/projectm/issues/724
              */
             m_writer.Write("inversesqrt(abs(");
-            OutputExpression(argument[0]);
+            OutputExpressionList(functionCall->argument, functionCall->function->argument);
             m_writer.Write("))");
             handled = true;
         }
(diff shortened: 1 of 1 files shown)


src/libprojectM/MilkdropPreset/CustomShape.cpp
```diff
@@ -277,54 +277,21 @@ void CustomShape::Draw()
                              static_cast<float>(*m_perFrameContext.border_g),
                              static_cast<float>(*m_perFrameContext.border_b),
                              static_cast<float>(*m_perFrameContext.border_a));
-            glLineWidth(1);
+            if (m_thickOutline)
+            {
+                glLineWidth(2);
```
kblaschke, line 282: Using glLineWidth >1 is not supported on all platforms/drivers (e.g. GLES) and will have really dirty artifacts on those which support it. Plus it mostly disables anti-aliasing. We should keep it that way until we have proper custom shader-based line rendering in place. See issue #682

src/libprojectM/MilkdropPreset/MilkdropShader.cpp
```diff
@@ -47,8 +49,9 @@ MilkdropShader::MilkdropShader(ShaderType type)
     } while (index < sizeof(m_randTranslation) / sizeof(m_randTranslation[0]));
 }
 
-void MilkdropShader::LoadCode(const std::string& presetShaderCode)
+void MilkdropShader::LoadCode(const std::string& presetShaderCode, bool dumpShaderCode)
 {
+    m_dumpShaderCode = dumpShaderCode;
```
kblaschke, line 54: Debug stuff should better be in #ifdefs, so it doesn't affect release builds, potentially writing lots of files on user hard drives. See the other debug outputs in MilkdropPreset. When we implement a debug API for 4.2, this could be handled via the API, returning the shaders etc. via callbacks for example.

Since libprojectM is a library, it should not write anything to the filesystem or even output text to stdout/stderr.

kblaschke (changes requested):
Pleas reconsider the line width changes, as I've specifically kept it this way because of the stated reasons - it will make things worse (e.g. not rendering any lines) on some platforms, which are not working fine.
Plus, debug output should either be done via a proper API (in release 4.2), or wrapped in #ifdefs to only make them available in debug builds.

src/libprojectM/MilkdropPreset/MilkdropShader.cpp
dpw13: Removed, this was a local commit not intended for upstream.

src/libprojectM/MilkdropPreset/CustomShape.cpp
dpw13: Removed, this was a local commit not intended for upstream.

kblaschke (approved):

PR #790, opened 2024-03-06
v4.1.x  - Build system fixes

kblaschke:
A few build fixes for 4.1 which got overlooked, as they only happen in very special circumstances.


dpw13 (approved):
Looks good to me. I'll take your word for it that the GLM build fixes work in the situations where they were previously broken.

kblaschke:
Added another fix to an issue I've found during testing: in the very rare case a preset has no warp shader, but a caomposite shader, the composite shader was not loaded either.

PR #792, opened 2024-03-06
v4.1.x - Add PThreads option for Emscripten

Blaquewithaq:
This PR adds back the USE_PTHREADS option for Emscripten builds only.

This flag is required in Emscripten builds to access certain features in the browser environment.


CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ file(TOUCH "${PROJECTM_DUMMY_SOURCE_FILE}")
 
 if(CMAKE_SYSTEM_NAME STREQUAL Emscripten)
     set(ENABLE_EMSCRIPTEN ON CACHE BOOL "Build for web with emscripten. Will also build the SDL2-based entrypoint." FORCE)
+    option(USE_PTHREADS "Enable pthreads" OFF)
```
dpw13, line 63: You might consider adding a status message below if Emscripten and USE_PTHREADS is enabled to make it clear when configuring. No change necessary though.

dpw13 (approved):
One optional comment.

PR #799, opened 2024-04-18
fix: parser test labels

mertovun:
* swaps misordered label values


kblaschke:
The tests are actually labeled correctly. There's a misleading comment in the first of the two test cases, which might give that impression. That should be deleted, but the tests should keep their respective names.

kblaschke (changes requested):
Keep the labels as-is, but either remove or fix this comment:

https://github.com/projectM-visualizer/projectm/blob/6356543bd61dc1bf742347635bc869fb56b082b6/tests/libprojectM/PresetFileParserTest.cpp#L64

kblaschke (approved):
Thanks!  Now it's both correct and no longer confusing 😄

PR #803, opened 2024-05-03
Add option to use system projectm-eval (used in vcpkg)

Blaquewithaq:
This PR introduces a new option to the CMake build system, enabling the utilization of a system-installed version of the ProjectM-Eval library. This modification is essential to adhere to the established guidelines for VCPKG, which stipulate that submodules should function as their own overlay-port.

Option Name: ENABLE_SYSTEM_PROJECTM_EVAL


cmake/FindProjectMEval.cmake
kblaschke, line 1: There's no need to write a find module, as projectm-eval provides a config package, which find_package() can use. This find script here also doesn't define the `projectM::Eval` target which is linked, so it won't work if the find script is used.

vendor/CMakeLists.txt
```diff
@@ -2,5 +2,7 @@ if(NOT ENABLE_SYSTEM_GLM)
     add_subdirectory(glm)
 endif()
 add_subdirectory(hlslparser)
+if(NOT ENABLE_SYSTEM_PROJECTM_EVAL)
```
kblaschke, line 5: Just check for the target here.

```suggestion
if(NOT TARGET projectM::Eval)
```

CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ cmake_dependent_option(ENABLE_GLES "Enable OpenGL ES support" OFF "NOT ENABLE_EM
 option(ENABLE_BOOST_FILESYSTEM "Force the use of boost::filesystem, even if the compiler supports C++17." OFF)
 cmake_dependent_option(ENABLE_INSTALL "Enable installing projectM libraries and headers." OFF "NOT PROJECT_IS_TOP_LEVEL" ON)
 option(ENABLE_SYSTEM_GLM "Enable use of system-install GLM library" OFF)
+option(ENABLE_SYSTEM_PROJECTM_EVAL "Enable use of system-install projectM-eval library" OFF)
```
kblaschke, line 77: I'd actually prefer this if it's available. Linux distro s like Gentoo also ideally want this as separate package. So the build script should just look for it by default, giving the dev the opportunity to disable it and use the built-in version.

```suggestion
option(ENABLE_SYSTEM_PROJECTM_EVAL "Enable use of system-install projectM-eval library" ON)
```

CMakeLists.txt
```diff
@@ -83,6 +84,10 @@ if(ENABLE_SYSTEM_GLM)
     find_package(GLM REQUIRED)
 endif()
 
+if(ENABLE_SYSTEM_PROJECTM_EVAL)
+    find_package(projectM-eval REQUIRED)
```
kblaschke, line 88: Remove the REQUIRED keyword and use correct casing for the config package dir:

```suggestion
    find_package(projectM-Eval)
```

kblaschke (changes requested):

CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ cmake_dependent_option(ENABLE_GLES "Enable OpenGL ES support" OFF "NOT ENABLE_EM
 option(ENABLE_BOOST_FILESYSTEM "Force the use of boost::filesystem, even if the compiler supports C++17." OFF)
 cmake_dependent_option(ENABLE_INSTALL "Enable installing projectM libraries and headers." OFF "NOT PROJECT_IS_TOP_LEVEL" ON)
 option(ENABLE_SYSTEM_GLM "Enable use of system-install GLM library" OFF)
+option(ENABLE_SYSTEM_PROJECTM_EVAL "Enable use of system-install projectM-eval library" ON)
```
kblaschke, line 77: No one will read this text anyway, but I'd suggest a slight change:
```suggestion
option(ENABLE_SYSTEM_PROJECTM_EVAL "Enable use of a system-installed/external projectM-eval library" ON)
```

kblaschke (approved):
Fix the help text if you want, I'm fine with both versions and the other changes 😉 
Thanks!

Issue #732, opened 2023-09-20
Add an error reporting and general logging API


kblaschke:
Currently, libprojectM has no good way of reporting errors and other information like debugging output to the integrating application. As of now, only failures loading a preset contain a mostly generic error message. Other debug messages can only be enabled in debug builds, printing directly to STDOUT, which isn't an optimal solution for a library.

For application developers and also preset authors, getting more information from the library is a great thing to increase productivity. We need to add a few API functions to configure and retrieve log messages:
- Add API functions to register and unregister a logging callback function.
- Add an API function to set the global log level.
- Add an API function to set log levels per component (audio, presets, renderer etc.).

Since logging will happen in most classes of libprojectM, providing per-instance logging is a bit tricky if we don't want to pass on the logging instance to all classes. I'll put two suggestions up for discussion, but if anyone has a different idea, please post it!

### Suggestion 1: Static global logging

This is the easiest way of adding logging to the library. Logging would not be tied to a specific projectM instance, but is configured globally for all instances created by the application. In almost all cases this will be perfectly fine, as most implementations will only need a single instance at any time. If multiple instances are running in parallel, the application needs to determine which instance has produced the log output, e.g. by using the thread ID or by remembering which projectM instance was called most recently.

**Advantages:**
- Works before even creating an instance, so initialization errors can already be logged.
- The callback doesn't even need to know about the projectM instances.
- Very easy to implement in libprojectM using global variables to store the callback pointer and log levels.

**Disadvantages:**
- No easy way to distinguish log output from different projectM instances.
- Log levels can only be set globally for all instances.
- Integrations have a harder time making the logging thread-safe if multiple projectM instances run in parallel on separate threads.

### Suggestion 2: Thread-local instance binding

This would provide a way to set separate logging callbacks and levels for each projectM instance. Since only one instance can run at a time in each thread, we can use a thread-local global variable to store a pointer to the logging information (callback pointer and log levels). Every time a projectM API function is called, it would set the thread-local variable to the current instance's logging structure.

**Advantages:**
- Callbacks get the projectM instance handle, and different callbacks can be defined per projectM instance, which makes it easy to determine the source instance.
- No multi-threading issues in integrations if callbacks are defined per thread.
- Log levels can be set per projectM instance.

**Disadvantages:**
- Slightly harder to implement in libprojectM, but can use the C++11 `thread_local` specifier.
- Still need a global callback for non-instance errors, e.g. to log errors during `projectm_create()`.
[...]

PR #805, opened 2024-05-12
CHECK_GL_ERROR additions

revmischa:
Maybe helpful for debugging bad GL calls


kblaschke:
Before merging and reworking this code again, I'd suggest leaving it in a branch to fix the GStreamer issues, and then implement a proper debugging API via this issue here:

https://github.com/projectM-visualizer/projectm/issues/732

Thst'd save us from reworking any existing code and use a proper debug class/functionality so an outside app can properly log any errors.

kblaschke:
I guess all the GL_INVALID_ENUM errors seen were due to that uninitialized framebuffer texture being used before the first frame was drawn. This is fixed in PRs #826  and #827.

May I close this PR and postpone proper GL error checking/logging to the implementation of issue #732?

revmischa:
Sure thing

PR #815, opened 2024-06-08
Add custom FBO rendering function

kblaschke:
Added a new API function `projectm_opengl_render_frame_fbo()` to pass an optional custom OpenGL FBO ID for the final rendering output copy operation. This will allow users to render to an offscreen texture or a framebuffer provided by the embedding application's drawing subsystem, e.g. Qt.

### DIFF
--- a/src/api/include/projectM-4/render_opengl.h
+++ b/src/api/include/projectM-4/render_opengl.h
@@ -34,12 +34,17 @@ extern "C" {
 /**
  * @brief Renders a single frame.
  *
- * @note Separate two-pass frame rendering is currently not supported by the C API as it is rarely used
- *       and also depends on the loaded preset.
- *
  * @param instance The projectM instance handle.
  */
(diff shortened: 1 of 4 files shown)


Blaquewithaq (approved):

Jonathhhan:
This is a great feature, thanks! This way it is possible to use it with an fbo in Open Frameworks. I did not had any issues so far. It would be very nice, to have it in the relese versions too.
```
void ofxProjectM::update() {
	ofPushStyle();
	fbo.bind();
	projectm_opengl_render_frame_fbo(projectMHandle, fbo.getId());
	fbo.unbind();
	ofPopStyle();
}
```
I need to use `ofPushStyle()` / `ofPopStyle()` because otherwise the alpha channel of the texture is gone.

kblaschke:
It'll be in the 4.2 release. I hope to get all the backlogged features done ASAP, e.g. the debugging/logging API, user sprites etc. so we can release it.

Issue #823, opened 2024-07-06
GL_INVALID_ENUM error on transition to the first preset


hack-s:
### Please confirm the following points:

- [X] This report is NOT about the Android apps in the Play Store
- [X] I have [searched the project page](https://github.com/search?q=org%3AprojectM-visualizer+type%3Aissue+&type=issues) to check if the issue was already reported

### Affected Project

libprojectM (including the playlist library)

### Affected Version

master branch

### Operating Systems and Architectures

Linux (x86_64)

### Build Tools

Compiler: GNU GCC

### Additional Project, OS and Toolset Details

c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

### Type of Defect

Specific bug in projectM code (please link the code in question)

### Describe the Issue

`GL_INVALID_ENUM` error present in OpenGL context via `glGetError()` on transition to the first/initial preset when the output texture from the current preset (initialized empty at that point) is attempted to being copied.

The error is caused by this line when `m_target = 0`:
https://github.com/projectM-visualizer/projectm/blob/master/src/libprojectM/Renderer/Texture.cpp#L58

The error does not cause any functional issue as far as I can tell, it just trips my error handler.

hack-s:
Here is my workaround for this issue, not sure if that is the preferred way to fix it though.
https://github.com/projectM-visualizer/projectm/pull/824

kblaschke:
This error is almost always happening because the OpenGL context isn't properly set up, as these OpenGL calls are the first ones being executed when initializjng projectM.

To make sure this doesn't happen because of such an issue, please check:
- You have set up an OpenGL context with a _minimum_ GL Core Profile version of 3.3, not a lower compatibility profile.
- You are using GLX, not EGL.
- The created context is active and properly initialized when you call `projectm_create()`.

hack-s:
To be more precise, the issue is happening when transitioning from the idle preset `idle://Geiss & Sperl - Feedback (projectM idle HDR mix).milk` to the first actual file based preset.

I confirmed this issue is present (on my system) when using the latest `frontend-sdl2` from master, tested both, X11 and Wayland on Ubuntu 22.04. OpenGL context setup looks good to me. Detection of the bug is easy to reproduce by adding a minimal GL error handler like the following to the `RenderLoop.cpp` `Run()` method, e.g. L51:
```
    while (true)
    {
        auto error = glGetError();

        switch (error)
        {
            case GL_NO_ERROR:
                break;
[...]

PR #824, opened 2024-07-07
Check if active preset output texture is valid before copying

hack-s:
Possible fix for https://github.com/projectM-visualizer/projectm/issues/823

### DIFF
--- a/src/libprojectM/ProjectM.cpp
+++ b/src/libprojectM/ProjectM.cpp
@@ -243,6 +243,9 @@ void ProjectM::StartPresetTransition(std::unique_ptr<Preset>&& preset, bool hard
     if (m_activePreset)
     {
         auto outputTexture = m_activePreset->OutputTexture();
+
+        // Check if the output texture is valid by testing if the OpenGL texturing target enum has been initialized.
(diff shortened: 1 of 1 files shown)


revmischa (commented):
comments would be cool! what is Type

hack-s:
I think a better fix would be using a `nullptr` for output texture instead of initializing it empty. But that would be a more intrusive change.

kblaschke:
> I think a better fix would be using a `nullptr` for the output texture initially instead of initializing it empty. But that would be a more intrusive change.

I absolutely don't mind intrusive changes if they are cleaner. If there's no texture, we can surely leave the shared pointer empty. Easier to check, and not more complicated to initialize it later.

If you want to make this change, please do so, that'd be highly appreciated!

kblaschke:
Issue comes from a preset not having an output texture yet (e.g. framebuffer color attachment is not yet initialized), so the texture has to be there, but will be created when the framebuffer is resized to something larger than 0x0, which happens on the first frame draw call.

Applied a fix to check for an empty texture ID in CopyTexture, and openend PRs #826 and #827, closing this one here.

PR #826, opened 2024-07-12
[V4.1.x] Fix GL_INVALID_ENUM error in CopyTexture

kblaschke:
Bumped libprojectM to v4.1.2 as well. Will create a bugfix release as this issue can break integrations checking for GL errors, e.g. GStreamer.

### DIFF
--- a/src/libprojectM/Renderer/Texture.cpp
+++ b/src/libprojectM/Renderer/Texture.cpp
@@ -15,7 +15,7 @@ Texture::Texture(std::string name, const int width, const int height, const bool
     , m_format(GL_RGB)
     , m_type(GL_UNSIGNED_BYTE)
 {
-    CreateEmptyTexture();
+    CreateNewTexture();
 }
 
 Texture::Texture(std::string name, int width, int height,
@@ -29,7 +29,7 @@ Texture::Texture(std::string name, int width, int height,
     , m_format(format)
     , m_type(type)
 {
-    CreateEmptyTexture();
+    CreateNewTexture();
 }
 
 Texture::Texture(std::string name, const GLuint texID, const GLenum target,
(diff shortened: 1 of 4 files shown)


Blaquewithaq (approved):

src/libprojectM/Renderer/Texture.cpp
```diff
@@ -99,7 +99,12 @@ auto Texture::IsUserTexture() const -> bool
     return m_isUserTexture;
 }
 
-void Texture::CreateEmptyTexture()
+auto Texture::Empty() const -> bool
+{
+    return m_textureId == 0;
```
hack-s, line 104: I was considering checking the texture id instead of the target. I ended up checking the target, because that was the value that directly triggered the `GL_INVALID_ENUM`. Would it be worth to check both values ?

hack-s: Disregard, just noticed, this is the backport. We should not do anything different here anyways.

hack-s:
Works great!

hack-s (approved):

revmischa (commented):
I don't have anything to add here, don't have much of an idea of what's going on. Adding some comments on the changed lines about why we need these checks or why New instead of Empty is preferable would be grand

src/libprojectM/Renderer/Texture.cpp
kblaschke: `m_target` is, except for the volume noise textures, always set to `GL_TEXTURE_2D`, even if the texture is empty. I think "target" is a misleading name for the texture's dimensionality, but it's how OpenGL calls it.

What triggered the error was actually `m_textureId` being zero because the actual texture wasn't initialized at the point when it was first used (copying the previous frame texture into the new preset - naturally, there's no previous texture right after init).

kblaschke:
> why New instead of Empty is preferable

I took this consideration because the previous name would be misleading now.

"Empty", in the new naming context, means there is no actual texture allocated as all, e.g. the texture object is 0 or "empty", like a `nullptr`. The `CreateNewTexture()` function creates a (_blank_) texture with a vlaid, non-zero GL texture ID, but the actual object is no longer considered empty, as a black texture is a texture, despite not a hugely useful one.

Hope this clears it up a bit.

Issue #825, opened 2024-07-08
[APP BUG] Unexpected crash of projectM visualizer in a custom Android app (Android 13, arm64-v8a)


Kyborg2011:
### Please confirm the following points:

- [X] This report is NOT about the Android apps in the Play Store
- [X] I have [searched the project page](https://github.com/search?q=org%3AprojectM-visualizer+type%3Aissue+&type=issues) to check if the issue was already reported

### Affected Application

Third-Party Integration (Kodi, VLC, etc. - please specify below)

### Affected Version

4.1.1

### Operating System

Other, all or not relevant

### Additional Application Details

I have downloaded projectM from current Github repo. To build it for Android I have used custom script (written by myself):

```shell
#!/bin/bash

ANDROID_SDK="/Users/aroslavivanov/Library/Android/sdk"
ANDROID_NDK="${ANDROID_SDK}/ndk/26.1.10909125"
CMAKE_MAKE_PROGRAM="${ANDROID_SDK}/cmake/3.22.1/bin/ninja"
MIN_SDK_VERSION=24

cd projectm/build

function build_projectm {
    ABI=$1
    mkdir -p ${ABI}

    cmake -S .. -B${ABI} \
        -G "Ninja Multi-Config" \
        -DCMAKE_SYSTEM_NAME=Android \
        -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM} \
        -DBUILD_SHARED_LIBS=ON \
        -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake \
        -DANDROID_ABI=${ABI} \
        -DANDROID_PLATFORM=android-${MIN_SDK_VERSION} \
        -DANDROID_ARM_NEON=TRUE \
        -DANDROID_TOOLCHAIN=clang \
        -DANDROID_STL=c++_shared \
        -DCMAKE_BUILD_TYPE=Release \
        -DBUILD_TESTING=NO \
        -DCMAKE_VERBOSE_MAKEFILE=YES \
        -DENABLE_SDL_UI=OFF \
        -DENABLE_PLAYLIST=OFF

    cmake --build ${ABI} --config "Release" --parallel

    mkdir -p ../../../distribution/libprojectM/lib/${ABI}
    cp ${ABI}/src/libprojectM/libprojectM-4.so ../../../distribution/libprojectM/lib/${ABI}/
}

build_projectm arm64-v8a
[...]

PR #830, opened 2024-08-03
fixed https://github.com/projectM-visualizer/projectm/issues/825

Kyborg2011:
After such a little changes a debug app on both my Android 13 devices started working smoothly...

photo_2024-08-03 21 57 03
photo_2024-08-03 21 56 55

### DIFF
--- a/src/libprojectM/MilkdropPreset/Shaders/PresetWarpVertexShaderGlsl330.vert
+++ b/src/libprojectM/MilkdropPreset/Shaders/PresetWarpVertexShaderGlsl330.vert
@@ -16,8 +16,8 @@ precision mediump float;
 layout(location = 0) in vec2 vertex_position;
 layout(location = 1) in vec2 rad_ang;
 layout(location = 2) in vec4 transforms;
-layout(location = 3) in vec2 center;
-layout(location = 4) in vec2 distance;
+layout(location = 3) in vec2 center2;
+layout(location = 4) in vec2 distance2;
 layout(location = 5) in vec2 stretch;
 
 uniform mat4 vertex_transformation;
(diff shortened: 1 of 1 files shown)


src/libprojectM/MilkdropPreset/Shaders/PresetWarpVertexShaderGlsl330.vert
```diff
@@ -16,8 +16,8 @@ precision mediump float;
 layout(location = 0) in vec2 vertex_position;
 layout(location = 1) in vec2 rad_ang;
 layout(location = 2) in vec4 transforms;
-layout(location = 3) in vec2 center;
-layout(location = 4) in vec2 distance;
+layout(location = 3) in vec2 center2;
```
revmischa, line 19: Can you please explain why changing this fixes it?

kblaschke: Same question. The only explanation I could come up with is that Android's shader Compiler somehow has `center` and `distance` as keywords.

I'd not postfix those variables with a 2 in the case, as there's clearly no 1 here. We could instead rather prefix the variables with `in_` or something, which would both fix the potential keyword issue and not make the names look weird.

kblaschke: I guess renaming the distance variable should suffice, as I cannot find any keyword or function `center` in any GL standard.

Could you try that, keeping `center` as-is and for example rename `distance` to `warpDistance`?

kblaschke:
I've changed the variable names according, keeping the commit info as-is otherwise.

Since we need to merge it into two branches (the fix will become part of the 4.1.2 release), there are now two new PRs open, #832 and #833. Closing this one here.

PR #846, opened 2024-10-07
Fix texture loading if casing mismatches filename

kblaschke:
Always use lower-case unqualified names to find and cache the actual textures.

### DIFF
--- a/src/libprojectM/PresetFactoryManager.cpp
+++ b/src/libprojectM/PresetFactoryManager.cpp
@@ -2,6 +2,8 @@
 
 #include <MilkdropPreset/Factory.hpp>
 
+#include <Utils.hpp>
+
 #include <algorithm>
 #include <cassert>
 #include <iostream>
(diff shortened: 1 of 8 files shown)


dpw13 (approved):
Looks good to me as long as everything builds.

src/libprojectM/PresetFactoryManager.cpp
```diff
@@ -135,9 +137,8 @@ auto PresetFactoryManager::ParseExtension(const std::string& filename) -> std::s
     if (start == std::string::npos || start >= (filename.length() - 1)) {
         return "";
     }
-    std::string ext = filename.substr(start + 1, filename.length());
-    std::transform(ext.begin(), ext.end(), ext.begin(), tolower);
-    return ext;
+
+    return Utils::ToLower(filename.substr(start + 1, filename.length()));
```
revmischa, line 141: why is it start + 1?

kblaschke: Because "start" always includes the `.`, we only want the extension without it.

Blaquewithaq (approved):

Raider1313 (approved):

Raider1313 (approved):

PR #859, opened 2024-12-11
hlslparser: Add missing "ldexp" and "faceforward" HLSL intrinsics.

kblaschke:
Had to simulate it with the "x * exp2(exp)" expression in GLSL, as GLSL only supports integer types as exponents in its native ldexp implementation, while HLSL solely uses floats.


dpw13 (approved):

revmischa (commented):
I don't have anything useful to provide here

PR #875, opened 2025-03-24
Fix playlist api headers and add tests

kblaschke:
Fixes issues mentioned in PR #874 and also adds buildsystem-based compilation tests with proper CMake-based dependency management (e.g. the compile test is re-executed every time an API header changes).


revmischa (approved):
thanks!

PR #880, opened 2025-04-29
Update README.md

revmischa:
Can we think of some other "see also" links to add to related projects/sites?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,12 @@ developers.
   a [development preview](https://github.com/kblaschke/frontend-sdl2/releases/tag/2.0-windows-pre3))
 - [Steam](https://store.steampowered.com/app/1358800/projectM_Music_Visualizer/) (Same as standalone development
   preview)
-- [Windows Store](https://www.microsoft.com/store/apps/9NDCVH0VCWJN) (Old 3.1.12 release )
 
 #### macOS
 
-- Standalone (currently only available as
-  a [development preview](https://github.com/kblaschke/frontend-sdl2/releases/tag/2.0-macos-pre1))
+- Standalone
+  - [C++ app development preview](https://github.com/kblaschke/frontend-sdl2/releases/tag/2.0-macos-pre1)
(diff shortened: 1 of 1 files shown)


kblaschke:
There's already a Wiki page listing projectM applications of all kinds, which we could link in the ReadMe:

https://github.com/projectM-visualizer/projectm/wiki/projectM-Applications

This way, we only have one location to update, and the Wiki is certainly a better place to put links (which come and go, naturally).

README.md
```diff
@@ -41,12 +41,13 @@ developers.
   a [development preview](https://github.com/kblaschke/frontend-sdl2/releases/tag/2.0-windows-pre1))
 - [Steam](https://store.steampowered.com/app/1358800/projectM_Music_Visualizer/) (Same as standalone development
   preview)
-- [Windows Store](https://www.microsoft.com/store/apps/9NDCVH0VCWJN) (Old 3.1.12 release )
+- [Windows Store](https://www.microsoft.com/store/apps/9NDCVH0VCWJN) (Old 3.1.12 release)
```
kblaschke, line 44: Since there's reports saying that this release no longer works on XBox, we may be better off to remove the link. We probably can't easily persuade Microsoft to remove it from the store, but at least no longer claim it to be an official release as it's probably more damaging than doing good.

kblaschke, line 44: ``
```suggestion
```

README.md
```diff
@@ -176,11 +179,11 @@ the [projectM organization's repositories](https://github.com/projectM-visualize
 
 ---
 
-# Todo
+# See Also
 
-- Improve projectM using the recently-released Milkdrop source (WIP).
-- Update the various implementations using libprojectM.
-- Update downstream projects with new versions.
+- [GStreamer plugin for offline rendering](https://github.com/projectM-visualizer/gst-projectm/)
```
mbaetgen-wup: it works for real-time rendering as well now, with the latest PR applied..

mbaetgen-wup (approved):

PR #883, opened 2025-04-30
Fix Emscripten build check workflow

kblaschke:
Updated the emsdk cache action to v14, as NodeJS changes broke the build.


revmischa:
Yep I ran into the same thing and fixed it yesterday

revmischa (approved):

PR #899, opened 2025-08-27
feat: Remove all comments from source code

anmthu008:
This commit removes all comments from the C/C++ source and header files in the `src` directory.

A Perl script was used to reliably remove both single-line and multi-line comments from all files. This was done to reduce the codebase size and remove distracting information.


kid-icarus:
As a fan of literate, self-documenting code I'm in full support of this change !!

kblaschke:
No, we will not remove comments from the codebase, period.

Why?
- Because our API documentation for users is based on Doxy gen comments
- Same for API docs regarding internal classes for projectM developers
- Some header comments contain important licensing information as recommended by the GPL

If you want to clean up your fork, you're free to do so. Just remember to add comments again before opening a PR here.

Closing.

PR #929, opened 2025-11-11
use built-in FindOpenGL module for GLES with cmake 3.27+

mbaetgen-wup:


CMakeLists.txt
```diff
@@ -169,8 +169,10 @@ else()
             message(FATAL_ERROR "OpenGL ES 3 support is currently only available for Linux platforms. You're building for ${CMAKE_SYSTEM_NAME}.")
         endif()
 
-        # We use a local find script for OpenGL::GLES3 until the proposed changes are merged upstream.
-        list(APPEND CMAKE_MODULE_PATH "${PROJECTM_SOURCE_DIR}/cmake/gles")
+        if (CMAKE_VERSION VERSION_LESS_EQUAL "3.22" OR CMAKE_SYSTEM_NAME STREQUAL Android)
```
kblaschke: Checked CMake's sources to be sure and saw there that the `GLES2` and `GLES3` targets were [actually added to CMake's find module in version 3.27](https://github.com/Kitware/CMake/blob/22e4e7448f266e4e655c86bacecea3b0c724753e/Modules/FindOpenGL.cmake#L60-L66) and not in v3.23 as I was expecting when I wrote the ReadMe in the module dir. So I think to be on the safe side, this should rather read:

```suggestion
        if (CMAKE_VERSION VERSION_LESS_EQUAL "3.26" OR CMAKE_SYSTEM_NAME STREQUAL Android)
```

Ditto in the package config file.

kblaschke (changes requested):
Just two small details left.

CMakeLists.txt
kblaschke: (If you want, you can update the ReadMe.md in `cmake/gles` accordingly.)

PR #933, opened 2025-12-01
Fix version in config header when used as subproject

LordMZTE:
When projectM is included in another CMake project using `add_subdirectory`, we will incorrectly insert the version number and commit revision of the top-level project in our `version.h` config header.

This changes the global `CMAKE_PROJECT_VERSION*` values to their local `PROJECT_VERSION*` equivalents for the innermost project. It will also set the working directory of the git process for determining the revision so we don't get the revision of the outer project.


kblaschke (approved):
Yup, that's definitely an issue that needs fixing.

PR #952, opened 2026-01-22
Add API flag to start presets with clean canvas

### LINKED ISSUE #298: [FEATURE] Add flag to C API to start newly loaded presets with a clean canvas if set
Hi All - is there a way to disable Smooth Transitions (ie. the previous preset merging into the new preset)? I found that setting Smooth Transition Duration to 0 has no effect. I am happy to try to fix this if someone can point me in the right direction.

Example:
The left frame is the first idle preset, and the second frame on the right is the first preset that is loaded. I would like to / should be able to have the second preset load fresh without any transition between the two.

image

revmischa:
Adds option to skip the initial texture copy when transitioning to new presets.

When enabled via `projectm_set_preset_start_clean(true)`, the previous
frame's texture is not drawn as the initial state for the new preset.
This provides a clean black canvas instead, which some applications may prefer.

Fixes #298


kblaschke (commented):
If the "Trigger CI" commit is removed/squashed, this is good to merge!

PR #953, opened 2026-01-22
Add preset load callback to playlist library

### LINKED ISSUE #946: [FEATURE] Add callback to playlist library to handle preset loading in the application
### Application or Project

libprojectM (including the playlist library)

### Is Your Enhancement Related to a Problem?

There are different scenarios where applications may need to handle preset loading in a different way than loading a physical file/path, like reading presets from an archive or a network source like HTTP, possibly in an asynchronous way. Currently, the playlist library only supports loading presets from the local filesystem.

### Your Suggested Enhancement

Add a new, optional callback to the playlist library that is called whenever the playlist library needs to switch to a new playlist item. This callback should have three arguments:
- the playlist index to be played 
- the preset filename/url at this index
- a void user data pointer with context information set by the application when registering the callback

Applications can then put any string as preset filename into the playlist and handle it accordingly. In theory, the application could even leave the filename empty and just use the index, as the playlist library doesn't really care about the filename in most of the functions (except for sorting and filtering).

revmischa:
Adds a callback that's invoked before loading each preset from the playlist.

The callback receives the preset index, filename, and whether it's a hard cut.
Returning `false` from the callback prevents the preset from loading, allowing
applications to implement custom filtering or logging.

Fixes #946


kblaschke (approved):
That'll help many devs - including myself 😁 

I'd say it's ready for merging!

PR #973, opened 2026-02-27
remove GL / GLES dependency from projectMConfig for any platform other than WIN32

mbaetgen-wup:


kblaschke (approved):
Let's finally merge this one.

PR #986, opened 2026-03-30
Detach textures from framebuffers before deleting

struktured:
## Summary

GL drivers keep internal references to textures attached to framebuffers. Deleting a texture while still attached can cause use-after-free in driver memory, observed as crashes on NVIDIA during rapid preset switching or window resize.

This commit detaches all textures from their FBOs before deletion in both `~Framebuffer()` and `SetSize()`. The `SetSize()` path uses a three-phase detach → resize → reattach pattern to avoid referencing stale texture IDs after `SetSize()` destroys and recreates the underlying GL texture.

## Changes
- **1 file**: `src/libprojectM/Renderer/Framebuffer.cpp` (+25, -1)

## Test plan
- [x] Clean build on Linux (GCC, RelWithDebInfo)
- [x] Runtime test with rapid preset switching (N-key)
- [x] No NVIDIA driver crashes during sustained preset cycling

### DIFF
--- a/src/libprojectM/Renderer/Framebuffer.cpp
+++ b/src/libprojectM/Renderer/Framebuffer.cpp
@@ -24,11 +24,11 @@ Framebuffer::~Framebuffer()
 {
     if (!m_framebufferIds.empty())
     {
-        // Delete attached textures first
-        m_attachments.clear();
-
+        // Delete FBOs first — this also releases driver references to attached textures.
         glDeleteFramebuffers(static_cast<int>(m_framebufferIds.size()), m_framebufferIds.data());
         m_framebufferIds.clear();
+
+        m_attachments.clear();
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/libprojectM/Renderer/Framebuffer.cpp
```diff
@@ -92,9 +104,21 @@ bool Framebuffer::SetSize(int width, int height)
     for (auto& attachments : m_attachments)
     {
         Bind(attachments.first);
+        // First detach all textures from framebuffer before destroying them.
+        // The GL driver keeps internal references to attached textures, so we must
+        // detach before deleting to avoid use-after-free in driver memory.
+        for (auto& texture : attachments.second)
```
kblaschke: Is there any real technical reason to add two new loops here? Each attachment is independently managed in an FBO, so the detach->reallocate->attach code can be placed in the very same loop without bloating the code.

kblaschke (changes requested):

struktured:
Addressed all review feedback:

1. **Destructor simplified** — moved `glDeleteFramebuffers()` before `m_attachments.clear()` instead of manually detaching. FBO deletion releases driver references to attached textures.
2. **SetSize merged into single loop** — detach → resize → reattach now happens per-attachment in one loop instead of three.
3. **C++17 structured bindings removed** — replaced `auto& [index, attachments]` with standard range-based `auto&` + `.first`/`.second` for C++14 compatibility.

Tested on NVIDIA RTX 3090 (driver 580.126.09), KDE Wayland, Qt 6.9.2.

kblaschke:
> Addressed all review feedback

Good to hear! If you could push these changes, I can re-run the build checks and test it on my boxes after Easter Sunday!

struktured:
Sorry about the delay — the changes were ready but I'd pushed to my fork instead of the PR branch. They're up now in 3303eaa. Ready for re-review whenever you get to it!

kblaschke (approved):
Thanks, looks clean now!

PR #994, opened 2026-04-24
[master/4.2] Fix HLSLParser: Always check preprocessor directive stack

kblaschke:
Was previously dereferencing the top element without checking for the stack to be empty (e.g. if there are mismatched preprocessor directives). The loop now aborts if there is at least one more closing directive (e.g. `#endif`) than opening ones.


mbaetgen-wup (approved):

rootnotez:
Might be worth making explicit that one PR puts these fixes on Master and one puts fixes on the 4.1.x branch.

PR #995, opened 2026-04-24
[v4.1.x] Fix HLSLParser: Always check preprocessor directive stack

kblaschke:
Was previously dereferencing the top element without checking for the stack to be empty (e.g. if there are mismatched preprocessor directives). The loop now aborts if there is at least one more closing directive (e.g. `#endif`) than opening ones.

### DIFF
--- a/vendor/hlslparser/src/HLSLParser.cpp
+++ b/vendor/hlslparser/src/HLSLParser.cpp
@@ -3600,7 +3600,7 @@ bool HLSLParser::ApplyPreprocessor(const char* fileName, const char* buffer, siz
     isCodeActive.push(true);
     m_tokenizer = HLSLTokenizer(fileName, buffer, length);
     sourcePreprocessed.clear();
-    while (m_tokenizer.GetToken() != HLSLToken_EndOfStream)
+    while (m_tokenizer.GetToken() != HLSLToken_EndOfStream && !isCodeActive.empty())
     {
         bool addOriginalSource = true;
 


OfficialIncubo:
There is an error:
`Error: File was unable to be removed Error: EBUSY: resource busy or locked, rmdir 'D:\a\projectm\projectm\vcpkg'`

kblaschke:
> There is an error: `Error: File was unable to be removed Error: EBUSY: resource busy or locked, rmdir 'D:\a\projectm\projectm\vcpkg'`

Again some random build server issue, this time the OS didn't release some lock it seems. Happens every now and then, can't control or fix it.

mbaetgen-wup (approved):

Oddicee (approved):

rootnotez:
Do https://github.com/projectM-visualizer/projectm/pull/995 and https://github.com/projectM-visualizer/projectm/pull/994 need to be applied in a certain order?

kblaschke:
> Do #995 and #994 need to be applied in a certain order?

No. It's the same change, one for the main development branch (#994, to `master`) and this one here is a backport/cherry-pick of the fix to the 4.1.x release branch.

In general, I try to backport crash/bug fixes and visual improvements to the current active release branch if the change can easily be applied or ported (e.g. the fix is not relying on or caused by a large change/refactor for the next minor release). New features and API calls are only added to the main branch through.

AriESQ:
Claude had this to say:


### `#else` branch is not guarded (same bug class)
The `PreprocessorElse` handler calls `.top()` and `.pop()` without checking for an empty stack. A bare `#else` before any `#if` would still crash:

```cpp
// Line 3628 — unguarded:
bool state = isCodeActive.top();
isCodeActive.pop();
isCodeActive.push(!state);
```

The loop-condition fix reduces exposure but does not fully eliminate the class of bug — a `#else` on an empty stack mid-stream is still reachable.

PR #996, opened 2026-04-25
feat(playlist): add M3U parser with UTF-8/Cyrillic filename support

### LINKED ISSUE #962: [DEV BUG] Preset playlist parser stops when it contains random/cyrillic characters
### Please confirm the following points:

- [x] This report is NOT about the Android apps in the Play Store
- [x] I have [searched the project page](https://github.com/search?q=org%3AprojectM-visualizer+type%3Aissue+&type=issues) to check if the issue was already reported

### Affected Project

libprojectM (including the playlist library)

### Affected Version

4.0+ (currently using 4.1.6)

### Operating Systems and Architectures

Windows (x64)

### Build Tools

_No response_

### Additional Project, OS and Toolset Details

_No response_

### Type of Defect

Specific bug in projectM code (please link the code in question)

### Log Output

```shell

```

### Describe the Issue

The playlist parser intentionally stops then it parses a file name with random/cyrillic characters (В, £, � and so on).
That means it only parses the first ones except the malformed file names to the end of the list in every directory.

Affected presets:
- Daft Sex - $o £ong.milk
- PowerMilk + Ech0 - Вown to the trip b.milk
- Xiaolan233 - �.milk
- ⚠ jnob - heat capacity ⏞ melting point⏞
- etc.

A workaround for (maybe) only Windows builds are:
- Create a temporary string variable and set UTF-8 locale with imbue
- Implement a trick that skips BOM (byte order mark) if it detects
- Convert it to wide string
- Parse the preset file with the temporary wide string
I have no idea about the another solution.

I don't know if this affects on other operating systems (Linux, MacOS, Android etc.).

EDIT: Forgot to send the code what it broke, so I think it's likely because of this: https://github.com/projectM-visualizer/projectm/blob/master/src/playlist/Playlist.cpp, else it's because of the - code elsewhere.

EDIT 2: OS I'm using: Windows 11 Pro
Display language or locale: Romania

Dipendra367:
Fixes #962

## What this does
Adds a new `M3UParser` class that reads `.m3u` and `.m3u8` playlist 
files into the projectM playlist library.

## Why it was broken
No M3U parser existed at all — the playlist library could only scan 
directories for `.milk` files.

## How it works
- Opens files in binary mode to preserve UTF-8 byte sequences
- Strips Windows `\r` CR line endings safely
- Skips `#EXTM3U` / `#EXTINF` metadata lines using direct `'#'` char 
  comparison (0x23 is always single-byte in UTF-8)
- Passes all other lines directly to `Playlist::AddItem()` — 
  Cyrillic, CJK, and any UTF-8 filename works transparently

### DIFF
--- a/src/playlist/CMakeLists.txt
+++ b/src/playlist/CMakeLists.txt
@@ -26,6 +26,8 @@ add_library(projectM_playlist_main OBJECT
         Playlist.hpp
         PlaylistCWrapper.cpp
         PlaylistCWrapper.hpp
+        M3UParser.cpp
+        M3UParser.hpp
         )
 
 target_include_directories(projectM_playlist_main
@@ -188,3 +190,4 @@ if(ENABLE_INSTALL)
     endif()
 
 endif()
+
--- a/src/playlist/M3UParser.cpp
+++ b/src/playlist/M3UParser.cpp
@@ -1,0 +1,64 @@
+#include "M3UParser.hpp"
+#include "Playlist.hpp"
+
+#include <fstream>
+#include <string>
+
+namespace libprojectM {
+namespace Playlist {
+
+auto M3UParser::LoadM3U(const std::string& filename,
+                        Playlist& playlist,
+                        bool allowDuplicates) -> uint32_t
+{
+    // Open in binary mode to avoid any platform newline translation
+    // that could corrupt multi-byte UTF-8 sequences
+    std::ifstream file(filename, std::ios::binary);
+    if (!file.is_open())
+    {
+        return 0;
+    }
+
+    uint32_t addedCount{0};
+    std::string line;
+
+    while (std::getline(file, line))
+    {
+        // Strip Windows-style \r (CR) from line endings.
+        // Must be done before any other checks so we don't treat
+        // "\r" as a non-empty path.
+        if (!line.empty() && line.back() == '\r')
+        {
+            line.pop_back();
+        }
+
+        // Skip empty lines
+        if (line.empty())
+        {
+            continue;
+        }
+
+        // Safe ASCII check: '#' is 0x23, always a single byte in UTF-8.
+        // We compare the raw char value directly — no ctype functions —
+        // so high-byte UTF-8 characters (e.g. Cyrillic 0xD0-0xBF) are
+        // never misidentified as comments or skipped.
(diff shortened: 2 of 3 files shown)


kblaschke:
Strange, I thought I had already written a comment on this PR, but probably GH acted up when I sent it. Here I go again:

Firstly, thanks for the contribution! This feature is surely appreciated, yet there's a consideration which wasn't made too obvious in our docs: The core projectM playlist library is, by design, fully agnostic to any playlist format like M3U[8] and others. It was meant as a very simple convenience replacement for the playlist functionality of previous projectM versions up to 3.x to make porting applications from projectM 3.x (and earlier) to 4.x easier. Any additional format or feature supported here will add maintenance cost for us, while probably only helping a few developers.

I'll add a more specific note in the ReadMe on that matter to avoid future confusion, as we don't want contributors to waste their time and effort. Again, sorry for not making this more clear in advance!

That said, I don't want to merge this change into libprojectM itself. This feature might still be useful to other developers/users though, so there are a few other places this can possibly go:

1. Any of our standalone applications ([SDL2/C++](https://github.com/projectM-visualizer/frontend-sdl-cpp), [SDL3/Rust](https://github.com/projectM-visualizer/frontend-sdl-rust) or [Qt/C++](https://github.com/projectM-visualizer/frontend-qt)). Note that the Qt application already uses an XML-based custom playlist format.
2. A new, separate middleware library that works in-between the application and libprojectM, providing support for additional playlist formats or even ZIP archives.
3. As an article in our documentation Wiki providing it as a reference implementation/example, as the code is small enough for that.

The open issue #962 is actually about supporting UTF-16/wchar filenames on Windows, as this was reported to trip up some users due to the library currently only properly supporting single-byte characters and UTF-8. Your PR doesn't fix that underlying issue, as this would require to store filenames as `wchar` on Windows (and possibly require an API change, e.g. adding `wchar` variants of the playlist management functions). If you still want to provide a C++14-compatible fix for the actual issue, which ideally doesn't make use of the now deprecated (C++20 onwards) `std::codecvt` namespace, that'd be highly appreciated! If you have any question regarding encoding (which isn't trivial at all), feel free to ask us on Discord!

I'll close this PR for now, but would really like to see M3U support being added somewhere, even if at a later time.

PR #1007, opened 2026-07-06
Audio: use both channels for beat detection (was left-only)

keeper-of-memes:
Beat detection (bass/mids/treble) updated only from m_spectrumL; the right-channel spectrum was computed every frame and then discarded, so content weighted to the right channel (or with an L/R imbalance) under-reacted. Average L+R into a mono spectrum and drive the three BeatDetect bands from that. Per-channel render waveforms are unchanged.

### DIFF
--- a/src/libprojectM/Audio/PCM.cpp
+++ b/src/libprojectM/Audio/PCM.cpp
@@ -62,14 +62,21 @@ void PCM::UpdateFrameAudioData(double secondsSinceLastFrame, uint32_t frame)
     UpdateSpectrum(m_waveformL, m_spectrumL);
     UpdateSpectrum(m_waveformR, m_spectrumR);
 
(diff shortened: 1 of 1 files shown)


kblaschke (approved):
Looks like I was confused by [this comment](https://github.com/projectM-visualizer/milkdrop2/blob/f05b0d811a87a17c4624170c26c93bac39b05bde/src/vis_milk2/pluginshell.cpp#L2034) which says it only used the left channel for the beat detection bands, while [the code clearly averages _all_ channels](https://github.com/projectM-visualizer/milkdrop2/blob/f05b0d811a87a17c4624170c26c93bac39b05bde/src/vis_milk2/pluginshell.cpp#L2061-L2063).

That should make the visuals look better, thanks!

PR #1009, opened 2026-07-06
CMake: use PROJECT_VERSION for PROJECTM_LIB_VERSION

keeper-of-memes:
`CMAKE_PROJECT_VERSION` refers to the top-level `project()` call, so when libprojectM is built as a subdirectory of another CMake project (`add_subdirectory` or FetchContent), `PROJECTM_LIB_VERSION` picks up the **parent** project's version instead of projectM's own. `PROJECT_VERSION` is scoped to the nearest `project()` call and yields the correct value in both standalone and embedded builds.

One line; no behaviour change for standalone builds.


kblaschke (approved):
Ahh yes, that was on my bucket list for a long time, but always forgot to actually change it.

Thanks!

Issue #1004, opened 2026-05-28
[FEATURE] Make it possible to select between GL Core and GL ES APIs at instance creation


kblaschke:
### Please confirm the following points:

- [x] This feature request is NOT for the Android apps in the Play Store
- [x] I have [searched the project page](https://github.com/search?q=org%3AprojectM-visualizer+type%3Aissue+&type=issues) to check if a similar request was already made

### Application or Project

libprojectM (including the playlist library)

### Is Your Enhancement Related to a Problem?

Currently, developers must decide to use either GL Core profile or the GL Embedded Standard at compile time, so an application that may need to use either one depending on the environment has to ship with two separate libraries.

### Your Suggested Enhancement

Add a new API call that enables the application to choose a specific GL profile for the instance. Alternatively, projectM could try to determine the proper GL variant by querying the current context GL version. This is also necessary for the original `projectm_create()` method, which will be used by applications that have adopted projectM before version 4.2.

### Alternative Solutions

Don't add any additional API functions, and only rely on context detection. Keeps the API slim, but removes some control from application, e.g. to prevent using the wrong GL variant in some edge cases.

keeper-of-memes:
Some device data relevant to the ES-variant selection and minimum-version questions here, from running libprojectM on tvOS through ANGLE's Metal backend:

- On real Apple hardware, ANGLE-Metal only reports ES 3.0 for Apple GPU family 4 and newer (A11+); families 2 and 3 (A8 through A10X) are capped at ES 2.0 (`DisplayMtl::getMaxSupportedESVersion`). Its simulator branch returns ES 3.0 unconditionally, so this only surfaces on device: `eglChooseConfig` succeeds with zero ES3 configs.
- I patched the gate locally to report ES 3.0 on families 2 and 3 as well, and libprojectM (with the requirements floor lowered from ES 3.2 to 3.0) runs the full ~290-preset Cream of the Crop pack on an Apple TV HD (A8): zero GL errors, zero swap failures over extended runs, steady 25 fps at 720p on a 50 Hz panel. Warp and composite shaders, blur and motion vectors all render.
[...]

PR #1010, opened 2026-07-06
GLResolver: fall back to app-managed EGL/GLES when a user resolver is set

keeper-of-memes:
### Problem

On platforms where the current GL context cannot be probed via the platform's native APIs, `GLResolver::Initialize()` fails hard even when the embedding application has explicitly supplied a resolver via `projectm_create_with_opengl_load_proc()`. Concrete case: ANGLE on tvOS/iOS. The app owns an EGL context backed by Metal, and there is no system EGL/GLX/WGL for the probe to query, so projectM refuses to initialise even though every GL entry point is resolvable through the app's proc loader.

### Change

When the context probe fails AND a user resolver is present, assume an app-managed EGL/GLES context and continue (logged at info level). The app has taken responsibility for providing GL entry points, so an unprobeable context is expected there rather than an error.

Behaviour without a user resolver is unchanged (hard error, as before).

### Notes for reviewers

- ~14 lines, single hunk in `GLResolver.cpp`.
- This is the change that makes libprojectM initialise under ANGLE-on-Metal (tvOS/iOS); it should equally help other embedded/sandboxed environments where the app owns the context.
- Builds clean on macOS (Release), unit tests pass. Running in production in a tvOS app (ANGLE, ES 3.0 on Metal).

### DIFF
--- a/src/libprojectM/Renderer/Platform/GLResolver.cpp
+++ b/src/libprojectM/Renderer/Platform/GLResolver.cpp
@@ -404,9 +404,20 @@ auto GLResolver::Initialize(UserResolver resolver, void* userData) -> bool
         std::string reason;
         if (!HasCurrentContext(currentContext, reason))
         {
-            m_loaded = false;
-            LOG_ERROR(std::string("[GLResolver] No current GL context present: ") + reason);
-            return false;
+            if (state.m_userResolver != nullptr)
+            {
+                currentContext.eglCurrent = true;
+                currentContext.eglAvailable = true;
+                LOG_INFO(std::string("[GLResolver] Current context could not be probed: ") +
+                         reason +
+                         "; assuming app-managed EGL/GLES context because a user resolver was supplied");
+            }
+            else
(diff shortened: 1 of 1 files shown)


keeper-of-memes:
Additional field evidence: this fallback is now validated in production on two hardware generations — Apple TV 4K 3rd gen (A15) and Apple TV HD (A8), both running libprojectM through ANGLE-Metal with an app-managed EGL context. Same behaviour on both: without the fallback, initialisation fails at the context probe despite a fully functional user resolver; with it, everything downstream works (the A8 additionally needs the ES-version gate relaxed, tracked separately in #1004 discussion).

kblaschke (approved):
This should work fine, especially since it'll fail later if a required function pointer can't be resolved via the user-provided callback.
If it does, but calling a function fails, IMO this is to blame on the developer as the application needs to provide a properly configured and working GL/EGL/GLES environment and we just have to trust on that to be the case.

PR #1017, opened 2026-07-27
Update GLM to 1.0.3

data-man:
Updated from https://github.com/g-truc/glm/releases/tag/1.0.3 with crlf corrections.


kblaschke (approved):
Changes look good. Just have to verify that the changed GLM target name still works with OS- and vcpkg-provided external dependencies.

kblaschke:
Works fine, apart from vcpkg failing to install libsystemd due to an outdated baseline commit, but that's another, unrelated issue which I'll fix separately. Merging.

PR #1019, opened 2026-08-22
Remove OpenGL requirement from pkg-config file

complexlogic:
The current .pc file generated by CMake includes `Requires: opengl`. Using `pkg-config` to find ProjectM fails on Mac because `opengl.pc` doesn't exist. Since #973, there is no longer any compile time dependency on OpenGL for Linux/Mac, so this line can simply be removed. The `Requires:` list is now empty in the generated .pc file, which is acceptable per spec, and `pkg-config` now works on Mac (tested on 26.5 Tahoe).


kblaschke:
Oh, yeah, this one slipped through as pkg-config is rarely used!
Thanks!

kblaschke (approved):
Should've removed this previously.

PR #1020, opened 2026-09-02
Optimize PNGs via oxipng

C0rn3j:
Shaves a shallow git clone by a couple megabytes. 

Zero loss of quality, this is not compression but pure optimization.


kblaschke (approved):
Definitely an improvement!

We're still working on getting a proper project website (https://projectm-visualizer.org/) up and running, and once it's finished, will remove all screenshots and other images from the source repository, as this isn't the proper place to put these files.

Until then, the smaller images should help reducing clone time and disk usage (even with the repository itself growing a bit).

PR #1021, opened 2026-09-03
Fix build metadata: vcpkg version and undefined summary variable

thedavidweng:
Two small build-metadata fixes, no behavior change:

- `vcpkg.json`: bump `version` from `4.1.2` to `4.2.0` to match `project(libprojectM VERSION 4.2.0)` in `CMakeLists.txt`. It was missed when the version was bumped.
- `CMakeLists.txt`: remove the configure-summary line printing `${ENABLE_SHARED_LINKING}`. That variable is never defined via `option()` anywhere in the tree and is not passed by CI, so the summary always printed an empty value.

Verified with a clean macOS configure (`cmake -S . -B build`): configuration succeeds and the summary no longer contains the empty line. Kept separate from any docs changes.


kblaschke (approved):
Makes total sense, I've totally missed to update the vcpkg manifest, which should help with local port overlays.

PR #1023, opened 2026-09-03
Disable vcpkg applocal for static Windows builds

thedavidweng:
Fixes the recurring Windows CI flake where the Release build fails with MSB3073 (exit code 32, sharing violation) in the vcpkg post-build step, e.g. 'vcpkg.exe z-applocal --target-binary=.../projectM-unittest.exe'.

Root cause: that post-build step exists only to deploy dependent DLLs next to executables (VCPKG_APPLOCAL_DEPS in the vcpkg toolchain). The static matrix legs use -static triplets with a static CRT and fully static linking, so there are no DLLs to deploy, the step is a no-op that still opens the freshly linked exe, exposing the build to a file-lock race (Defender scan / concurrent post-build events of the two unittest projects).

Change: pass -DVCPKG_APPLOCAL_DEPS=OFF whenever matrix.libs == static, in both configure blocks (main build and C++ interface test). Shared legs are untouched and keep DLL deployment for ctest. No build-output change: static executables are self-contained, and install/upload steps are unaffected.

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -65,8 +65,12 @@ jobs:
         run: |
           if("${{ matrix.libs }}" -eq "shared") {
             $shared_libs = "ON"
+            $applocal_deps = "ON"
           } else {
             $shared_libs = "OFF"
+            # Static triplets link everything (incl. CRT) statically, so there are no DLLs to deploy.
+            # Disabling applocal removes a proven file-lock race (MSB3073, exit 32) with zero benefit lost.
+            $applocal_deps = "OFF"
           }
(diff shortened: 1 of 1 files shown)


kblaschke (approved):
Thanks for the fix!
I don't have a  local Windows system to debug this issue, thought this was some kind of build runner issue as it only happened from time to time.

Will merge once the build checks are done.

thedavidweng:
> Thanks for the fix!
> I don't have a  local Windows system to debug this issue, thought this was some kind of build runner issue as it only happened from time to time.
> 
> Will merge once the build checks are done.

Yes, it's CI Flask that happens occasionally, it should be more stable now

